PR #7, opened 2015-08-09
[WIP] DDS import

Squareys:
Hi @mosra !

As promissed, here is the WIP Pullrequest. The tests currently still fail and print two identical arrays (for compressed images). For uncompressed, the size of the imagedata seems larger than it should, did not figure out why that is yet (the size calculated correctly in the importer).

Current features/TODOs:
 - [x] Uncompressed RGB
 - [x] Uncompressed RGBA
 - [x] DXT(1|3|5) Compressed images
 - [x] Mipmaps (untested for compressed images)
 - [x] Cubemaps (untested)
 - [x] 3D Textures

Greetings, 
Squareys :)


mosra:
Okay, first things first: :+1:

The tests are printing the same values but failing because `Containers::Array` does not have `operator==` implemented and so just pointer values are compared. You have to do something like this:
```cpp
#include <Corrade/TestSuite/Compare/Container.h>

...

CORRADE_COMPARE_AS(image->data(), Containers::ArrayView<const char>(pixels),
```

For uncompressed, the failing test is my bug mosra/magnum#104 (yes, you are computing the size correctly).

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -0,0 +1,356 @@
...
+        colorFormat.uncompressed = ColorFormat::RGB;
+        components = 3;
+    } else if (ddsh.ddspf.rgbBitCount == 24 &&
+               ddsh.ddspf.rBitMask == 0x00FF0000 &&
+               ddsh.ddspf.gBitMask == 0x0000FF00 &&
+               ddsh.ddspf.bBitMask == 0x000000FF) {
+        colorFormat.uncompressed = ColorFormat::BGR;
+        components = 3;
+    } else if (ddsh.ddspf.rgbBitCount == 8) {
+//        colorFormat = ColorFormat::Luminance; //TODO equivalence in Magnum?
```
mosra:
```cpp
#ifndef MAGNUM_TARGET_GLES2
ColorFormat::Red;
#else
ColorFormat::Luminance;
#endif
```

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -0,0 +1,356 @@
...
+    DXT1 = 0x31545844, //!< (MAKEFOURCC('D','X','T','1')).
+    DXT2 = 0x32545844, //!< (MAKEFOURCC('D','X','T','2')), not supported.
+    DXT3 = 0x33545844, //!< (MAKEFOURCC('D','X','T','3')).
+    DXT4 = 0x34545844, //!< (MAKEFOURCC('D','X','T','4')), not supported.
+    DXT5 = 0x35545844, //!< (MAKEFOURCC('D','X','T','5')).
+    DXT10 = 0x30315844 //!< (MAKEFOURCC('D','X','1','0')), not supported.
+};
+
+inline UnsignedInt clampSize(UnsignedInt size) {
+    return (size <= 0) ? 1 : size = 1;
```
mosra: Just `Math::max(size, 1)` might be enough ...

Squareys: Yeah, I tried that, but my compiler complained and I didn't have the patience to deal with it yet :) I will try again later.

mosra: Oh, right... `Math::max(size, 1u)` may solve the ambiguous type deduction.

Squareys: @mosra Ah! Perfect, thanks! Will try this later :)

Squareys:
> For uncompressed, the failing test is my bug mosra/magnum#104 (yes, you are computing the size correctly).

Alright, cool :) Tests for compressed images run through now.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -0,0 +1,387 @@
...
+        UnsignedInt w = width;
+        UnsignedInt h = height;
+        UnsignedInt d = depth;
+
+        /* load all mipmaps for current surface */
+        for (UnsignedInt i = 0; i < numMipmaps && (w || h); i++) {
+            /* shrink to next power of 2 */
+            w = Math::max(w >> 1, 1u);
+            h = Math::max(h >> 1, 1u);
+            d = Math::max(d >> 1, 1u);
```
mosra: Just a suggestion, but I think that using `Vector3i`/`Vector3ui` instead of separate width/height/depth would reduce the verbosity and amount of function parameters quite a bit. Math functions will work without change (as expected), including `Math::max()` and operator `>>`.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+void DdsImporter::doOpenFile(const std::string& filename) {
+    /* let doOpenData handle it */
+    AbstractImporter::doOpenFile(filename);
+}
+
+UnsignedInt DdsImporter::doImage2DCount() const { return _imageData2D.size(); }
+
+std::optional<ImageData2D> DdsImporter::doImage2D(UnsignedInt id) {
+    if (id < _imageData2D.size()) {
```
mosra: This check is not needed (the `AbstractImporter` does this already using information from `doImage2DCount()`), similarly for the 3D image below.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -0,0 +1,387 @@
...
+        _imageData3D.emplace_back(format, Vector3i{Int(width), Int(height), Int(depth)}, Containers::Array<char>(data, size));
+    } else {
+        Error() << "Unsupported image dimensions:" << dimensions;
+    }
+}
+
+void DdsImporter::doOpenFile(const std::string& filename) {
+    /* let doOpenData handle it */
+    AbstractImporter::doOpenFile(filename);
+}
```
mosra: Completely removing the `doOpenFile()` override will have the same effect. ;)

(Yes, I had this in the PngImporter too, it was ugly, it's removed now and I'm sorry about that)

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -0,0 +1,387 @@
...
+        colorFormat.uncompressed = ColorFormat::RGB;
+        components = 3;
+    } else if (ddsh.ddspf.rgbBitCount == 24 &&
+               ddsh.ddspf.rBitMask == 0x00FF0000 &&
+               ddsh.ddspf.gBitMask == 0x0000FF00 &&
+               ddsh.ddspf.bBitMask == 0x000000FF) {
+        colorFormat.uncompressed = ColorFormat::BGR;
+        components = 3;
+    } else if (ddsh.ddspf.rgbBitCount == 8) {
+#ifndef MAGNUM_TARGET_GLES2
```
mosra: You can indent the preprocessor stuff too, it's not a portability problem as far as I can tell.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
Squareys: Sure, will do. That is a remnant of the code I started with.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
Squareys: Alright, good to know :D

Squareys: Uhm, really? All tests fail now...

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+    }
+}
+
+UnsignedInt DdsImporter::doImage2DCount() const {  return (_volume) ? 0 : _imageData.size(); }
+
+std::optional<ImageData2D> DdsImporter::doImage2D(UnsignedInt id) {
+    ImageDataOffset& dataOffset = _imageData[id];
+
+    // copy image data
+    Containers::Array<char>* data = new Containers::Array<char>(dataOffset._data.size());
```
mosra: Why the `new` here? (also you are leaking the `data`)

Similarly in `doImage3D()`.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+#ifdef MAGNUM_TARGET_GLES2
+#include <Magnum/Context.h>
+#include <Magnum/Extensions.h>
+#endif
+
+namespace Magnum { namespace Trade {
+
+/**
+ * Flags to indicate which members of a DdsHeader contain valid data.
+ */
```
mosra: These enums are not part of the public API, so no need for Doxygen comments.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+    const Vector3i size{Int(ddsh.width), Int(ddsh.height), Int(Math::max(ddsh.depth, 1u))};
+
+    /* check how many mipmaps to load */
+    const UnsignedInt numMipmaps = (ddsh.flags & DdsDescriptionFlag::MipMapCount)
+                    ? ((ddsh.mipMapCount == 0) ? 0 : ddsh.mipMapCount - 1)
+                    : 0;
+
+    /* load all surfaces for the image (6 surfaces for cubemaps) */
+    const UnsignedInt numImages = (isCubemap) ? 6 : 1;
+    size_t offset = 128;
```
mosra: Can the `128` be represented using some calculation? E.g. `sizeof(DdsHeader) + something`? Ideally in combination with `static_assert()` that it it really equals to 128.

mosra:
Sorry for the late reply -- looks great, really, I'm amazed :)

Commenting inline about some minor things, other than that ready to merge I think?

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
Squareys: Yup, sure, `sizeof(DdsHeader) + 4`.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
Squareys: ImageData2D doesn't data, right? Don't I need to make sure it stays around? For the uncompressed case, I'm moving data, which would, of course make the `new` superfluous. There does not seem to be a constructor to move data into ImageData2D when uncompressed.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
Squareys: Alright, will do.

Squareys:
I think there are some `if (`s around, I will generally check for code style and take care of the issues you pointed out. The data memleak will take some more discussion, though.

> Sorry for the late reply

No problem! I figured you were on vacation ;)

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
mosra: Ahh, right, sorry, I missed the `static_cast<void*>(data->data())` line. I'm now doing a rework of image classes (in order to fix mosra/magnum#104) and then there of course will be a constructor that moves the data into uncompressed `ImageData2D`.

Currently you can do just `data->release().release()` to "disown" the array from image and then disown the data pointer from array. Then the `new` is not needed.

Squareys: > and then there of course will be a constructor that moves the data into uncompressed ImageData2D.

Awesome! 

> data->release().release()

Okay, thanks, will do so.

Squareys:
@mosra Here you go, this is pretty much finished from my side. If there are any bigger issues, I'd be glad to take care of them, for minor issues I will look at your cleanup commit once you merged ;)

mosra:
Great, thanks a lot!

I merged it locally and made some modifications so all the tests pass now, but it depends on not yet published Magnum changes, so I'll push everything at once... another crazily large change :)
> 69 files changed, 3692 insertions(+), 2576 deletions(-)

Squareys:
Awesome, thank you! :D

mosra:
Merged (finally). Thank you.

However, my additions are not yet completely done, so the tests are still failing (in current master). Will be fixed as part of mosra/magnum#104.

PR #8, opened 2015-10-01
Add Vorbis/Ogg Importer based on stb_vorbis

Squareys:
Hi @mosra !

Stb is really something, this went pretty quick! The importer only supports 16 bit mono/stereo, no streaming, but here is ogg files support for magnum.

stb_vorbis could also support float data, which will have to wait until `audio-extension-support` is done, though.

It would be awesome if you could have a quick look at this (it's basically only 10 lines of code or so, everything else is pretty much identical to the wav importer).

Greetings,
Squareys


src/MagnumPlugins/StbVorbisImporter/StbVorbisImporter.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+#include "stb_vorbis.c"
+
+namespace Magnum { namespace Audio {
+
+StbVorbisImporter::StbVorbisImporter() = default;
+
+StbVorbisImporter::StbVorbisImporter(PluginManager::AbstractManager& manager, std::string plugin): AbstractImporter(manager, std::move(plugin)) {}
+
+StbVorbisImporter::~StbVorbisImporter() { close(); }
```
mosra: The destructor is not needed (equivalent code will be generated implicitly).

src/MagnumPlugins/StbVorbisImporter/StbVorbisImporter.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+bool StbVorbisImporter::doIsOpened() const { return _data; }
+
+void StbVorbisImporter::doOpenData(Containers::ArrayView<const char> data) {
+
+    Int numChannels, frequency;
+    Short* decodedData;
+
+    Int samples = stb_vorbis_decode_memory(reinterpret_cast<const UnsignedByte*>(data.data()), data.size(), &numChannels, &frequency, &decodedData);
+    _data = Containers::Array<char>(reinterpret_cast<char*>(decodedData), samples*numChannels*2);
```
mosra: I would save the array into local variable and then move it into `_data` only at the end when everything succeeded (so you don't have to call `_data = nullptr` on each error exit).

Squareys: Sometimes there is an error though and decodedData points to valid heap data. This is the case for unsupported channel counts for example. Should I then just `delete` the data?

src/MagnumPlugins/StbVorbisImporter/Test/StbVorbisImporterTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+    CORRADE_VERIFY(importer.openFile(Utility::Directory::join(STBVORBISIMPORTER_TEST_DIR, "mono16.ogg")));
+
+    CORRADE_COMPARE(importer.format(), Buffer::Format::Mono16);
+    CORRADE_COMPARE(importer.frequency(), 96000);
+    Containers::Array<char> data = importer.data();
+    CORRADE_COMPARE(data.size(), 4);
+    CORRADE_COMPARE(data[0], '\xcd');
+    CORRADE_COMPARE(data[1], '\x0a');
+    CORRADE_COMPARE(data[2], '\x2b');
+    CORRADE_COMPARE(data[3], '\x0a');
```
mosra: Including `<Corrade/TestSuite/Compare/Container.h>` and doing the following should work too, I think:
```cpp
CORRADE_COMPARE_AS(importer.data(), 
    Containers::Array<char>::from('\xcd', ...), 
```

Squareys: Alright, of course. I think I followed the WavImporter code too blindly :)

src/MagnumPlugins/StbVorbisImporter/StbVorbisImporter.cpp
mosra: It will be done implicitly when the local Array instance goes out of scope, no?

If I remember correctly, memory in stb_* is allocated using `malloc` so you can't use `delete`. See e.g. fd9726c52aea146fb2ff93978c57678ed5ffb023 for solution. Consult the documentation for details (might be done differently in stb_vorbis, I don't know).

src/MagnumPlugins/StbVorbisImporter/Test/StbVorbisImporterTest.cpp
mosra: And I should clean it up.

src/MagnumPlugins/StbVorbisImporter/StbVorbisImporter.cpp
Squareys: Ah! Now I get it, I had missunderstood, sorry!

mosra:
One important thing re the naming -- the class and file names are okay, it's `Audio::StbVorbisImporter`, put the plugin (the library and thus also all CMake variables) should be named `StbVorbisAudioImporter` to differentiate it from scene/image importers from `Trade` namespace. Sorry about the chaos :)

Squareys:
@mosra Alright, I fixed all the aforementioned issues and did a fresh rebase onto master :)

src/MagnumPlugins/StbVorbisAudioImporter/pluginRegistration.cpp
```diff
@@ -0,0 +1,29 @@
...
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include "MagnumPlugins/StbVorbisImporter/StbVorbisImporter.h"
+
+CORRADE_PLUGIN_REGISTER(StbVorbisImporter, Magnum::Audio::StbVorbisImporter,
```
mosra: This needs to be `CORRADE_PLUGIN_REGISTER(StbVorbisAudioImporter, Magnum::Audio::StbVorbisImporter, ...` too to make dynamic plugin loading work properly.

doc/building-plugins.dox
```diff
@@ -79,6 +79,8 @@ By default no plugins are built and you need to select them manually:
     plugin.
 -   `WITH_STBPNGIMAGECONVERTER` -- @ref Trade::StbPngImageConverter "StbPngImageConverter"
     plugin.
+-   `WITH_STBVORBISAUDIOIMPORTER` -- @ref Audio::StbVorbisImporter "StbVorbisAudioImporter" plugin.
+    plugin.
```
mosra: Hey, you have `plugin.` twice in here ;)

src/MagnumPlugins/StbVorbisAudioImporter/Test/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/configure.h.cmake
+               ${CMAKE_CURRENT_BINARY_DIR}/configure.h)
+
+include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})
+
+corrade_add_test(StbVorbisImporterTest StbVorbisImporterTest.cpp LIBRARIES MagnumStbVorbisImporterTestLib)
```
mosra: Just for consistency with library/plugin name I would name this `StbVorbisAudioImporterTest` and `MagnumStbVorbisAudioImporterTestLib`, too.

src/MagnumPlugins/StbVorbisAudioImporter/StbVorbisImporter.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+bool StbVorbisImporter::doIsOpened() const { return _data; }
+
+void StbVorbisImporter::doOpenData(Containers::ArrayView<const char> data) {
+
+    Int numChannels, frequency;
+    Short* decodedData;
+
+    Int samples = stb_vorbis_decode_memory(reinterpret_cast<const UnsignedByte*>(data.data()), data.size(), &numChannels, &frequency, &decodedData);
+    Containers::Array<char> tempData = Containers::Array<char>(reinterpret_cast<char*>(decodedData), samples*numChannels*2);
```
mosra: Maybe you missed my earlier comment, the `decodedData` pointer [needs to be freed using `free()` instead of `delete`](https://github.com/nothings/stb/blob/bdcc1535d6b5e2c8386f5126d267ab2bfeb9c8cb/stb_vorbis.c#L233) to avoid undefined behavior. Solution is to use custom deleter instead of the default provided in `Containers::Array`:

```cpp
Containers::Array<char> tempData{reinterpret_cast<char*>(decodedData), samples*numChannels*2,
    [](char* data, std::size_t) { std::free(data); }};
```

Squareys: Yeah, I missed that, sorry.

src/MagnumPlugins/StbVorbisAudioImporter/Test/CMakeLists.txt
Squareys: Does this include renaming the `StbVorbisImporterTest.cpp` file?

src/MagnumPlugins/StbVorbisAudioImporter/pluginRegistration.cpp
Squareys: Oh, and the include is wrong, too, right?

Squareys:
@mosra Thanks for the feedback! I fixed-up the issues you pointed out. If there isn't anything else, I will rebase.

mosra:
Awesome. Will merge as soon as possible. One more thing, could you please enable the new plugin in the CIs (`package/ci/appveyor.yml`, `package/ci/travis.yml`)? Thank you.

Squareys:
Will do as soon as possible (tonight probably).

Squareys:
Alright, everything fixed for CIs, which build properly now. A thing I think might be worth mentioning: I had weird errors on all the CIs, invalid pointers or whatever.

First I thought they were caused by the custom deleter possibly not being moved in the array move constructor, but in the end moving the creation of the decoded data array, after checking for sample counts of `-1` and `-2` (which would have resulted in negative sizes casted to `size_t`), solved the issues.

All in all, this is ready for merge from my side once again :)

mosra:
Merged, thanks a lot.

(I did a whitespace cleanup in `stb_vorbis.c` so the following commits were rebased. Hope that's okay.)

Squareys:
>  so the following commits were rebased. Hope that's okay.

Sure, no problem!

PR #13, opened 2016-07-21
FreeTypeFont: link to Freetype's dependencies if it was built static

LB--:
This allows users to build the FreeTypeFont plugin as a shared library even if they are using a static Freetype installation. CMake's [Freetype find module](https://github.com/Kitware/CMake/blob/v2.8.8/Modules/FindFreetype.cmake) doesn't provide any info about whether Freetype was built as shared or static, nor which dependencies it was built with, so instead I check the library name for `dll`, `so`, or `dylib` to find out - falsely assuming it is static is mostly harmless. Unfortunately there's no easy way to check which dependencies it needs, so I just pull in all of them as optional and leave it to the user to configure them properly if needed.

Part of the motivation for this is that using Visual Studio to build Freetype as a shared library is a pain to do - Freetype's `CMakeLists.txt` actually has a check that ensures only MinGW can be used for creating a shared library on Windows, so instead you have to [work around it by hand](http://stackoverflow.com/a/7387618/1959975). It would be nice if instead we could ignore all that and just build it statically on Windows.

### DIFF
--- a/modules/FindFreetype.cmake
+++ b/modules/FindFreetype.cmake
@@ -1,0 +1,165 @@
+#.rst:
+# FindFreetype
+# ------------
+#
+# Locate FreeType library
+#
+# This module defines
+#
+# ::
+#
+#   FREETYPE_LIBRARIES, the library to link against
+#   FREETYPE_FOUND, if false, do not try to link to FREETYPE
+#   FREETYPE_INCLUDE_DIRS, where to find headers.
+#   FREETYPE_VERSION_STRING, the version of freetype found (since CMake 2.8.8)
+#   This is the concatenation of the paths:
+#   FREETYPE_INCLUDE_DIR_ft2build
+#   FREETYPE_INCLUDE_DIR_freetype2
+#
+#
+#
+# $FREETYPE_DIR is an environment variable that would correspond to the
+# ./configure --prefix=$FREETYPE_DIR used in building FREETYPE.
+
+#=============================================================================
+# Copyright 2007-2009 Kitware, Inc.
+#
+# Distributed under the OSI-approved BSD License (the "License");
+# see accompanying file Copyright.txt for details.
+#
+# This software is distributed WITHOUT ANY WARRANTY; without even the
+# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the License for more information.
+#=============================================================================
+# (To distribute this file outside of CMake, substitute the full
+#  License text for the above reference.)
+
+# Created by Eric Wing.
+# Modifications by Alexander Neundorf.
+# This file has been renamed to "FindFreetype.cmake" instead of the correct
+# "FindFreeType.cmake" in order to be compatible with the one from KDE4, Alex.
+
+# Ugh, FreeType seems to use some #include trickery which
+# makes this harder than it should be. It looks like they
+# put ft2build.h in a common/easier-to-find location which
+# then contains a #include to a more specific header in a
+# more specific location (#include <freetype/config/ftheader.h>).
+# Then from there, they need to set a bunch of #define's
+# so you can do something like:
+# #include FT_FREETYPE_H
+# Unfortunately, using CMake's mechanisms like include_directories()
+# wants explicit full paths and this trickery doesn't work too well.
+# I'm going to attempt to cut out the middleman and hope
+# everything still works.
+
+set(FREETYPE_FIND_ARGS
+  HINTS
+    ENV FREETYPE_DIR
+  PATHS
+    /usr/X11R6
+    /usr/local/X11R6
+    /usr/local/X11
+    /usr/freeware
+    ENV GTKMM_BASEPATH
+    [HKEY_CURRENT_USER\\SOFTWARE\\gtkmm\\2.4;Path]
+    [HKEY_LOCAL_MACHINE\\SOFTWARE\\gtkmm\\2.4;Path]
+)
+
+find_path(
+  FREETYPE_INCLUDE_DIR_ft2build
+  ft2build.h
+  ${FREETYPE_FIND_ARGS}
+  PATH_SUFFIXES
+    include/freetype2
+    include
+    freetype2
+)
+
+find_path(
+  FREETYPE_INCLUDE_DIR_freetype2
+  NAMES
+    freetype/config/ftheader.h
+    config/ftheader.h
+  ${FREETYPE_FIND_ARGS}
+  PATH_SUFFIXES
+    include/freetype2
+    include
+    freetype2
+)
+
+if(NOT FREETYPE_LIBRARY)
+  find_library(FREETYPE_LIBRARY_RELEASE
+    NAMES
+      freetype
+      libfreetype
+      freetype219
+    ${FREETYPE_FIND_ARGS}
+    PATH_SUFFIXES
+      lib
+  )
+  find_library(FREETYPE_LIBRARY_DEBUG
+    NAMES
+      freetyped
+      libfreetyped
+      freetype219d
+    ${FREETYPE_FIND_ARGS}
+    PATH_SUFFIXES
+      lib
+  )
+  include(${CMAKE_CURRENT_LIST_DIR}/SelectLibraryConfigurations.cmake)
+  select_library_configurations(FREETYPE)
+endif()
+
+unset(FREETYPE_FIND_ARGS)
+
+# set the user variables
+if(FREETYPE_INCLUDE_DIR_ft2build AND FREETYPE_INCLUDE_DIR_freetype2)
+  set(FREETYPE_INCLUDE_DIRS "${FREETYPE_INCLUDE_DIR_ft2build};${FREETYPE_INCLUDE_DIR_freetype2}")
+  list(REMOVE_DUPLICATES FREETYPE_INCLUDE_DIRS)
+endif()
+set(FREETYPE_LIBRARIES "${FREETYPE_LIBRARY}")
+
+if(EXISTS "${FREETYPE_INCLUDE_DIR_freetype2}/freetype/freetype.h")
+  set(FREETYPE_H "${FREETYPE_INCLUDE_DIR_freetype2}/freetype/freetype.h")
+elseif(EXISTS "${FREETYPE_INCLUDE_DIR_freetype2}/freetype.h")
+  set(FREETYPE_H "${FREETYPE_INCLUDE_DIR_freetype2}/freetype.h")
+endif()
+
+if(FREETYPE_INCLUDE_DIR_freetype2 AND FREETYPE_H)
+  file(STRINGS "${FREETYPE_H}" freetype_version_str
+       REGEX "^#[\t ]*define[\t ]+FREETYPE_(MAJOR|MINOR|PATCH)[\t ]+[0-9]+$")
+
+  unset(FREETYPE_VERSION_STRING)
+  foreach(VPART MAJOR MINOR PATCH)
+    foreach(VLINE ${freetype_version_str})
+      if(VLINE MATCHES "^#[\t ]*define[\t ]+FREETYPE_${VPART}[\t ]+([0-9]+)$")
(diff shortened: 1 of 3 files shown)


mosra:
Cool!

One thing that I'm not really sure about -- Visual Studio (as opposed to MinGW) doesn't link to DLLs but to `*.lib` files, so `FREETYPE_LIBRARIES` doesn't contain any `*.dll` files, if I remember correctly. And personally I have no idea how one does check for that properly, because from compiler PoV linking to static and shared libs looks completely the same.

LB--:
Actually, we are really lucky here. On my system:
* MinGW shared results in `lib/libfreetype.dll.a` and `bin/libfreetype.dll` - both have `dll` in their name so it doesn't matter which ends up in `FREETYPE_LIBRARIES`
* MinGW static results in just `lib/libfreetype.a`
* Visual Studio shared would be whatever you name it when you follow the workaround
* Visual Studio static results in just `lib/freetyped.lib`

So it all works out in the end. Though I did just realize I am checking the entire path and not just the filename, so let me go fix that real quick...

LB--:
Alright, it should be fixed. But reading your comment again, I realize you're talking about when Visual Studio builds Freetype as shared. Since this is technically an unsupported way to build Freetype on Windows, I wouldn't worry too much about it - the whole point of this PR was to make it so Visual Studio users didn't have to do that.

LB--:
So I looked around, and it turns out `ftoption.h` does report which dependencies were used to build FreeType, but to get that information in CMake the only way I can think of is with `try_compile` for each dependency, which is rather messy, but it would allow us to tell the user which dependencies they need if one can't be found rather than having them run into confusing linker errors later and wondering where they went wrong. What would you suggest I do?

mosra:
@LB-- parse ftoption.h with some regexes? i'm doing that for `Corrade/configure.h` and it isn't as horrible as it seems

LB--:
@mosra Maybe - when an option is not defined, the entire `#define` is still there, just in a block comment, e.g. `/* #define FT_CONFIG_OPTION_USE_HARFBUZZ */` versus `#define FT_CONFIG_OPTION_USE_HARFBUZZ`. Freetype also supports letting the user do whatever they want in the file, so they could even have both the commented out and the not-commented-out version if they wanted. But if we're willing to ignore that scenario...

mosra:
Ideally this should also go in the upstream `FindFreetype.cmake` so other people can make use of it .. but then we need to upstream also `FindHarfBuzz.cmake`...

LB--:
Alright, let me know what you think of the current implementation.

> Ideally this should also go in the upstream `FindFreetype.cmake` so other people can make use of it .. but then we need to upstream also `FindHarfBuzz.cmake`...

Do you want me to do that?

src/MagnumPlugins/FreeTypeFont/CMakeLists.txt
```diff
@@ -31,6 +31,57 @@ set(CMAKE_FIND_FRAMEWORK LAST)
...
+        else()
+            if(NOT "${_use_dep}" STREQUAL "")
+                set(_require_dep ON)
+            endif()
+        endif()
+
+        # Require the dependency only if Freetype was built with it
+        if(_require_dep)
+            find_package(${_dep} QUIET REQUIRED)
+            list(APPEND FREETYPE_LIBRARIES ${${_DEP}_LIBRARY} ${${_DEP}_LIBRARIES})
```
mosra: I would create a new variable (and then link to it also) instead of modifying `FREETYPE_LIBRARIES`. Not sure about the interactions when using FreeType in more than one directory.

mosra:
Just thinking out loud at the moment -- this feels like it should be part of `FindFreeType.cmake` instead of a part of particular buildsystem in particular project. Having it directly in `FindFreeType.cmake` would cover also the case when having static FreeType and static FreeTypeFont plugin.

LB--:
Currently we're using CMake's FindFreetype module, so we would either have to override CMake's module with our own, or I'd have to  contribute this to CMake directly (which would only help people using the newest versions of CMake). I made this PR to work around the fact that CMake's FindFreetype module was inadequate but I guess overriding it with our own would make more sense.

src/MagnumPlugins/FreeTypeFont/CMakeLists.txt
LB--: I changed it to `FREETYPE_DEPENDENCIES`, is that fine?

mosra: Maybe `FREETYPE_DEPENDENCY_LIBRARIES` to follow the naming conventions?

mosra:
What we could definitely do is to "fork" upstream `FindFreetype.cmake` and put it into our `modules` directory so it works at least in the case of building shared `FreeTypeFont` plugin with static FreeType. Then the people get an option to either don't care and use the upstream module or copy our forked module to their own repository in case they want to have static FreeType + static plugin working in their project too.

I wouldn't bother too much with the licensing -- just keeping the original CMake license so there are not many differences from upstream version.

LB--:
Alright, let me know what you think. You'll want to diff against https://github.com/Kitware/CMake/blob/b213a7f6ab0d4aa18e7b704bf1cf4994fae77254/Modules/FindFreetype.cmake
Also, that's the last possible version we can fork from without backporting because after that they switch to using `string(APPEND)` which isn't recognized by older versions of CMake.

mosra:
Great job once again :)

Could you perhaps put the original unchanged upstream module as a separate commit with info about which version it was taken from so it's easily diffable just by viewing file history?

LB--:
@mosra done.

modules/FindFreetype.cmake
```diff
@@ -145,10 +172,9 @@ if(FREETYPE_INCLUDE_DIR_freetype2 AND FREETYPE_H)
   endforeach()
 endif()
 
-
 # handle the QUIETLY and REQUIRED arguments and set FREETYPE_FOUND to TRUE if
 # all listed variables are TRUE
-include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
+include(${CMAKE_ROOT}/Modules/FindPackageHandleStandardArgs.cmake)
```
mosra: I would do just `include(FindPackageHandleStandardArgs)`, similarly for `SelectLibraryConfigurations` above. AFAIK this absolute inclusion is done by CMake developers to prevent accidental module overrides from user-provided module dirs.

modules/FindFreetype.cmake
```diff
@@ -159,6 +185,59 @@ find_package_handle_standard_args(
...
+
+      # Use regex to check if a depdency is needed
+      # Example of when a dependency was used to build Freetype:
+      # #define FT_CONFIG_OPTION_USE_BZIP2
+      # Example of when a dependency was not used to build Freetype:
+      # /* #define FT_CONFIG_OPTION_USE_HARFBUZZ */
+      # We want to match the first and not the second
+      set(_match_before "\n[ \t]*\\#define[ \t]*FT_CONFIG_OPTION")
+      set(_match_after "[ \t]*\n")
+      string(REGEX MATCH "${_match_before}_USE_${_DEP}${_match_after}" _use_dep "${_ftoption_h}")
```
mosra: I would move this line to the `else()` clause so the handling of `_USE_` and `_SYSTEM_` looks symmetric.

modules/FindFreetype.cmake
```diff
@@ -159,6 +185,59 @@ find_package_handle_standard_args(
...
+      else()
+        if(NOT "${_use_dep}" STREQUAL "")
+          set(_require_dep ON)
+        endif()
+      endif()
+
+      # Require the dependency only if Freetype was built with it
+      if(_require_dep)
+        find_package(${_dep} QUIET REQUIRED)
+        list(APPEND FREETYPE_DEPENDENCY_LIBRARIES ${${_DEP}_LIBRARY} ${${_DEP}_LIBRARIES})
```
mosra: Doesn't the `${_DEP}_LIBRARIES` already contain what's in `${_DEP}_LIBRARY`?

modules/FindFreetype.cmake
LB--: Unfortunately we need to know the result of both `_USE_ZLIB` and `_SYSTEM_ZLIB` - the latter may be set while the former is not.

modules/FindFreetype.cmake
```diff
@@ -159,6 +185,59 @@ find_package_handle_standard_args(
...
+        endif()
+      else()
+        if(NOT "${_use_dep}" STREQUAL "")
+          set(_require_dep ON)
+        endif()
+      endif()
+
+      # Require the dependency only if Freetype was built with it
+      if(_require_dep)
+        find_package(${_dep} QUIET REQUIRED)
```
mosra: Why `QUIET`?

modules/FindFreetype.cmake
LB--: Some of the find modules only use one or the other, some use both. I'm just bulletproofing so I don't have to special case them. E.g. There is only `ZLIB_LIBRARIES` and only `HARFBUZZ_LIBRARY`. We could change the latter case, though.

modules/FindFreetype.cmake
LB--: I only did that because FreeType's CMake build does it. I could remove `QUIET` if you want.

modules/FindFreetype.cmake
LB--: Oops, I just made that change without testing and actually it doesn't work for `SelectLibraryConfigurations.cmake` so that one will have to remain absolute.

mosra: It should be just `SelectLibraryConfigurations`, without the extension.

LB--: Er, my mistake, I forgot to remove the `.cmake` at the end...

modules/FindFreetype.cmake
mosra: I'm not sure how the errors will be propagated to the user if both `QUIET` and `REQUIRED` is set.

modules/FindFreetype.cmake
mosra: Yes, please -- only `_LIBRARIES`.

modules/FindFreetype.cmake
LB--: `QUIET` only applies when there are no errors.  See https://github.com/Kitware/CMake/blob/v2.8.12/Modules/FindPackageHandleStandardArgs.cmake#L105-L114

mosra: Ah, great, let's keep it then :)

*(note to self: add `QUIET` when finding dependencies in all my Find modules)*

modules/FindHarfBuzz.cmake
```diff
@@ -61,3 +62,5 @@ if(NOT TARGET HarfBuzz::HarfBuzz)
         IMPORTED_LOCATION ${HARFBUZZ_LIBRARY}
         INTERFACE_INCLUDE_DIRECTORIES ${HARFBUZZ_INCLUDE_DIR})
 endif()
+
+set(HARFBUZZ_LIBRARIES "${HARFBUZZ_LIBRARY}")
```
mosra: Just the following should be enough I think:
```cmake
set(HARFBUZZ_LIBRARIES ${HARFBUZZ_LIBRARY})
```

And that's also my last comment, then I'm happy to merge :)

modules/FindFreetype.cmake
LB--: I just added `HARFBUZZ_LIBRARIES` and made it mirror `HARFBUZZ_LIBRARY`.

modules/FindHarfBuzz.cmake
LB--: Ah, that's my paranoia about spaces in paths kicking in. I find it hard to remember how each command expands arguments and whether they convert spaces to lists etc. so I just surround everything with quotes overzealously. Fixed.

mosra: From my investigation: once you create a string with spaces in it it, it will stay so in all further variable expansions, unless you explicitly replace spaces with `;`.

PR #19, opened 2016-09-30
Enhance DrFlacAudioImporter to allow loading of any format below 24 bits

Alicemargatroid:
DrFlacAudioImporter can now load any format below 24-bits, and correctly loads them to the same specification as WavAudioImporter. 

I'm currently working on converting 24-bit audio to IEEE float, and 32-bit audio to IEEE double, but this is a quick patch to get 4, 8, 12, and 16 bit audio working for DrFlacAudioImporter, and improving test coverage.


Alicemargatroid:
@mosra @Squareys If possible, please review this at your earliest convenience; I can't fork magnum-plugins twice, and I have DrWavAudioImporter working!

mosra:
Thanks!

You can always create another branch and create the other pull request from the other branch.

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -51,21 +87,14 @@ void DrFlacImporter::doOpenData(Containers::ArrayView<const char> data) {
...
-        _format = Buffer::Format::Mono8;
-    else if(numChannels == 1 && bitsPerSample == 16)
-        _format = Buffer::Format::Mono16;
-    else if(numChannels == 2 && bitsPerSample == 8)
-        _format = Buffer::Format::Stereo8;
-    else if(numChannels == 2 && bitsPerSample == 16)
-        _format = Buffer::Format::Stereo16;
-    else {
+    // Normalize bit amounts to multiples of 8, rounding up
+    UnsignedInt normalizedBitsPerSample = (bitsPerSample / 8) + ((bitsPerSample % 8) ? 1 : 0);
```
mosra: maybe `normalizedBytesPerSample` would match the actual value better? Also, shortening the code:

```cpp
UnsignedInt normalizedBytesPerSample = (bitsPerSample + 7)/8;
```

Also, what are the allowed values of `bitsPerSample` in the FLAC format? Can you put that into a comment?

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -35,6 +35,42 @@
...
+        { B::Format::Stereo8, B::Format::Stereo8, B::Format::Stereo16, B::Format::StereoFloat, B::Format::StereoDouble },   // Stereo
+        { B::Format::Quad8, B::Format::Quad8, B::Format::Quad16, B::Format::Quad32, B::Format::Quad32 },                    // Not a thing
+        { B::Format::Quad8, B::Format::Quad8, B::Format::Quad16, B::Format::Quad32, B::Format::Quad32 },                    // Quad
+        { B::Format::Quad8, B::Format::Quad8, B::Format::Quad16, B::Format::Quad32, B::Format::Quad32 },                    // Also not a thing
+        { B::Format::Surround51Channel8, B::Format::Surround51Channel8, B::Format::Surround51Channel16, B::Format::Surround51Channel32, B::Format::Surround51Channel32 }, // 5.1
+        { B::Format::Surround61Channel8, B::Format::Surround61Channel8, B::Format::Surround61Channel16, B::Format::Surround61Channel32, B::Format::Surround61Channel32 }, // 6.1
+        { B::Format::Surround71Channel8, B::Format::Surround71Channel8, B::Format::Surround71Channel16, B::Format::Surround71Channel32, B::Format::Surround71Channel32 }  // 7.1
```
mosra: Why is the Float and Double format here if it's not supported by FLAC?

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -51,21 +87,14 @@ void DrFlacImporter::doOpenData(Containers::ArrayView<const char> data) {
...
-    else if(numChannels == 2 && bitsPerSample == 8)
-        _format = Buffer::Format::Stereo8;
-    else if(numChannels == 2 && bitsPerSample == 16)
-        _format = Buffer::Format::Stereo16;
-    else {
+    // Normalize bit amounts to multiples of 8, rounding up
+    UnsignedInt normalizedBitsPerSample = (bitsPerSample / 8) + ((bitsPerSample % 8) ? 1 : 0);
+
+    if(numChannels == 0 || numChannels == 3 || numChannels == 5 || numChannels > 8 ||
+       normalizedBitsPerSample == 0 || normalizedBitsPerSample > 2) {
```
mosra: I'd do the inverse, listing all combinations that are supported:

```cpp
if(!(numChannels == 1 || numChannels == 2 || numChannels == 4 || numChannels = 6 || numChannels = 8) ||
   !(normalizedBytesPerSample == 1 || normalizedBytesPerSample == 2)) {
```

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -74,12 +103,30 @@ void DrFlacImporter::doOpenData(Containers::ArrayView<const char> data) {
         return;
     }
 
-    Containers::Array<char> tempData(samples*sizeof(int32_t));
+    _frequency = handle->sampleRate;
+    _format = flacFormatTable[numChannels][normalizedBitsPerSample];
```
mosra: If there are just 1 or 2 bytes per sample allowed, why the `flacFormatTable` above contains also 4 and 8 byte formats (Float and Double)? These will never get used.

mosra: Also, since `numChannels` and `normalizedBytesPerSample`  is always >= 1, I would do it like this:

```cpp
_format = flacFormatTable[numChannels - 1][normalizedBytesPerSample - 1];
```

and then remove the sentinel first column/row from the table.

Then I would also make unreachable entries (3 and 5 channels) of the table default-constructed (`Buffer::Format{}`) and then assert on them just to be sure:
```cpp
CORRADE_INTERNAL_ASSERT(_format != Buffer::Format{});
```

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -74,12 +103,30 @@ void DrFlacImporter::doOpenData(Containers::ArrayView<const char> data) {
...
+            for(char& item : _data) {
+                item = item - 128;
+            }
+            break;
+        }
+        case 2: {
+            _data = convert32PCM(tempData, samples, sizeof(Short));
+            break;
+        }
+    }
```
mosra: I'd prefer just a simple if/else as it is way shorter, then also asserting that there's no other value ever taken:

```cpp
if(normalizedBytesPerSample == 1) {
    _data = convert32PCM(tempData, samples, sizeof(UnsignedByte);
    for(char& item: _data) item -= 128;
} else if(normalizedBytesPerSample == 2) {
    _data = convert32PCM(tempData, samples, sizeof(Short));
} else CORRADE_ASSERT_UNREACHABLE();
```

Isn't the `convert32PCM()` equivalent to `_data = std::move(tempData);` in the first case anyway? Or am I missing something?

src/MagnumPlugins/DrFlacAudioImporter/Test/DrFlacImporterTest.cpp
```diff
@@ -42,30 +42,83 @@ class DrFlacImporterTest: public TestSuite::Tester {
...
+
+              &DrFlacImporterTest::stereo8,
+              &DrFlacImporterTest::stereo16,
+//              &DrFlacImporterTest::stereo24,
+
+              &DrFlacImporterTest::quad16,
+//              &DrFlacImporterTest::quad24,
+
+              &DrFlacImporterTest::surround51Channel16
+//              &DrFlacImporterTest::surround51Channel24
```
mosra: Why are these commented out?

In case the tests should pass but they don't (due to a limitation in the dr_flac implementation or due to a limitation in the plugin), then there should be [CORRADE_EXPECT_FAIL()](http://mosra.cz/blog/corrade-doc/Tester_8h.html#afe4fc83d4e99d4e56d989a6eb005f7c5) used. In case the test won't *ever* pass (because there's no 24bit format in OpenAL, e.g.), then they shouldn't be there at all.

Since there are no 24bit test files present, I assume the latter :)

src/MagnumPlugins/DrFlacAudioImporter/Test/DrFlacImporterTest.cpp
```diff
@@ -42,30 +42,83 @@ class DrFlacImporterTest: public TestSuite::Tester {
...
+            });
 }
 
 void DrFlacImporterTest::mono8() {
     DrFlacImporter importer;
     CORRADE_VERIFY(importer.openFile(Utility::Directory::join(DRFLACAUDIOIMPORTER_TEST_DIR, "mono8.flac")));
 
     CORRADE_COMPARE(importer.format(), Buffer::Format::Mono8);
-    CORRADE_COMPARE(importer.frequency(), 22254);
+    CORRADE_COMPARE(importer.frequency(), 22050);
```
mosra: Huh, I never thought there could be such small differences in frequencies...

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
Alicemargatroid: @mosra OpenAL does not support 24, 32, or 64 bit integer audio.

24 bit integer audio corresponds to a Float, as the mantissa of an IEEE 754 float is 24 bits.
32-bit integer audio corresponds to a Double, as the mantissa of an IEEE 754 double is 32 bits.

Therefore, if the FLAC contains 24-bit or 32-bit audio, we need to use those extensions to fully preserve the audio information.

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
Alicemargatroid: @mosra 4-64; I'll add some documentation

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
Alicemargatroid: @mosra The reason is because I am working on 3, 4, and 8 byte samples at the moment!

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
Alicemargatroid: @mosra It won't be shorter once I have 3 (24 bit audio), 4 (32 bit audio), and 8 (64 bit audio) working.

convert32PCM() drops a configurable amount of data; DrFlac loads EVERYTHING into 32 bit PCM by zero padding, so we drop the padding to recover the original data.

src/MagnumPlugins/DrFlacAudioImporter/Test/DrFlacImporterTest.cpp
Alicemargatroid: @mosra These are commented out because I intend to enable them as soon as I have 24 bit audio working.

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
mosra: Ah, right. Now I understand :)

(Sorry if my comments sound angry, mean and/or grumpy today, I'm completely overloaded with work and it's killing me.)

src/MagnumPlugins/DrFlacAudioImporter/Test/DrFlacImporterTest.cpp
mosra: Ok. I understand. Sorry :)

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
mosra: Okay (apologies) :)

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
Alicemargatroid: @mosra Absolutely no problem, and I didn't notice any offending tone!

src/MagnumPlugins/DrFlacAudioImporter/Test/DrFlacImporterTest.cpp
Alicemargatroid: No problem! Maybe I should have mentioned it in my PR head

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
Alicemargatroid: @mosra Rather than use _v (as anything starting with _ is reserved as a global identifier), I used bvf.

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -35,6 +35,42 @@
...
+        { B::Format::Quad8, B::Format::Quad8, B::Format::Quad16, B::Format::Quad32, B::Format::Quad32 },                    // Not a thing
+        { B::Format::Quad8, B::Format::Quad8, B::Format::Quad16, B::Format::Quad32, B::Format::Quad32 },                    // Quad
+        { B::Format::Quad8, B::Format::Quad8, B::Format::Quad16, B::Format::Quad32, B::Format::Quad32 },                    // Also not a thing
+        { B::Format::Surround51Channel8, B::Format::Surround51Channel8, B::Format::Surround51Channel16, B::Format::Surround51Channel32, B::Format::Surround51Channel32 }, // 5.1
+        { B::Format::Surround61Channel8, B::Format::Surround61Channel8, B::Format::Surround61Channel16, B::Format::Surround61Channel32, B::Format::Surround61Channel32 }, // 6.1
+        { B::Format::Surround71Channel8, B::Format::Surround71Channel8, B::Format::Surround71Channel16, B::Format::Surround71Channel32, B::Format::Surround71Channel32 }  // 7.1
+    };
+
+    Containers::Array<char> convert32PCM(const Containers::Array<char>& container, UnsignedInt samples, UnsignedInt size) {
```
Alicemargatroid: Done

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
mosra: Argh! Really? I thought that only `__something` and `_Something` are reserved (for both names and macros), `_something` is fine as long as it is either a macro or *not* in global scope.

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
mosra: (But still my suggestion about removing the first column/row and asserting on unreachable table entries holds ;))

Alicemargatroid: Done!

Alicemargatroid:
@mosra This should be good to go too!

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -44,42 +84,107 @@ auto DrFlacImporter::doFeatures() const -> Features { return Feature::OpenData;
...
+    // 8-bit needs to become unsigned
+    if(normalizedBytesPerSample == 1) {
+        // Convert to unsigned
+        for(char& item : _data) {
+            item = item - 128;
+        }
+    }
+
+    // 24-bit needs to become float
+    else if(normalizedBytesPerSample == 3) {
```
mosra: Coding style and braces:

```cpp
...
} else if(normalizedByttesPerSample = 3) {
...
```

Similarly elsewhere ;)

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -44,42 +84,107 @@ auto DrFlacImporter::doFeatures() const -> Features { return Feature::OpenData;
...
+        const char* floatEnd = reinterpret_cast<const char*>(floatData.end());
+
+        _data = Containers::Array<char>(samples*sizeof(Float));
+        std::copy(floatBegin, floatEnd, _data.begin());
+    }
+
+    // 32-bit integers need to be normalized to Double (with a 32 bit mantissa)
+    else if(normalizedBytesPerSample == 4) {
+        Containers::Array<Int> tempData(samples);
+        drflac_read_s32(handle.get(), samples, reinterpret_cast<Int*>(tempData.begin()));
```
mosra: The `drflac_read_s32()` function is called twice in case of 32-bit samples. Can you redesign the flow so it gets called only once in all cases?

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -44,42 +84,107 @@ auto DrFlacImporter::doFeatures() const -> Features { return Feature::OpenData;
...
+    else if(normalizedBytesPerSample == 3) {
+        Containers::Array<Float> floatData(samples);
+
+        for (UnsignedInt i = 0; i < samples; ++i) {
+
+            UnsignedInt s0 = _data[i*3 + 0];
+            UnsignedInt s1 = _data[i*3 + 1];
+            UnsignedInt s2 = _data[i*3 + 2];
+
+            Int sample32 = (Int)(s0 << 8) | (s1 << 16) | (s2 << 24);
```
mosra: Please don't use old-style casts. In this case use a "constructor cast":

```cpp
Int sample32 = Int((s0 << 8) | (s1 << 16) | (s2 << 24));
```

GCC/Clang should also emit warnings on this by default -- there are four such occurences in the code.

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -44,42 +84,107 @@ auto DrFlacImporter::doFeatures() const -> Features { return Feature::OpenData;
...
+            floatData[i] = (Float)(sample32 / 2147483648.0);
```
mosra: There's a [Math::normalize()](http://mosra.cz/blog/magnum-doc/namespaceMagnum_1_1Math.html#a767c5cede6c4e244234055ed897fec84) function for this:

```cpp
floatData[i] = Math::normalize<Float>(sample32);
```

Similarly for the other two cases below.

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -44,42 +84,107 @@ auto DrFlacImporter::doFeatures() const -> Features { return Feature::OpenData;
...
-        _format = Buffer::Format::Mono8;
-    else if(numChannels == 1 && bitsPerSample == 16)
-        _format = Buffer::Format::Mono16;
-    else if(numChannels == 2 && bitsPerSample == 8)
-        _format = Buffer::Format::Stereo8;
-    else if(numChannels == 2 && bitsPerSample == 16)
-        _format = Buffer::Format::Stereo16;
-    else {
+    // FLAC supports any bitspersample from 4 to 64, but DrFlac always gives us 32-bit samples
+    // So we normalize bit amounts to multiples of 8, rounding up
```
mosra: Coding style: please use `/* */`-style comments everywhere. The `//` ones are semantically considered as temporary TODO or FIXME-style comments, should not be used for documenting the code.

Similarly in other cases.

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
```diff
@@ -35,6 +35,46 @@
...
+    // number of bits = 5
+    const Buffer::Format  flacFormatTable[8][5] = {
+        { bfv(Mono8),   bfv(Mono16),   bfv(MonoFloat),   bfv(MonoDouble)   },       // Mono
+        { bfv(Stereo8), bfv(Stereo16), bfv(StereoFloat), bfv(StereoDouble) },       // Stereo
+        { Buffer::Format{}, Buffer::Format{}, Buffer::Format{}, Buffer::Format{} }, // Not a thing
+        { bfv(Quad8), bfv(Quad16), bfv(Quad32), bfv(Quad32) },                      // Quad
+        { Buffer::Format{}, Buffer::Format{}, Buffer::Format{}, Buffer::Format{} }, // Also not a thing
+        { bfv(Surround51Channel8), bfv(Surround51Channel16), bfv(Surround51Channel32), bfv(Surround51Channel32) }, // 5.1
+        { bfv(Surround61Channel8), bfv(Surround61Channel16), bfv(Surround61Channel32), bfv(Surround61Channel32) }, // 6.1
+        { bfv(Surround71Channel8), bfv(Surround71Channel16), bfv(Surround71Channel32), bfv(Surround71Channel32) }  // 7.1
```
mosra: Can you also update the documentation in the header file to mention all newly supported formats?

src/MagnumPlugins/DrFlacAudioImporter/DrFlacImporter.cpp
Alicemargatroid: @mosra I'll do this, but I would consider this a bad coding style choice: /* */ comments can't nest, which makes it very difficult to comment out large blocks of code while testing. I've encountered this a few times so far.

Alicemargatroid:
@mosra Lets do it this way

mosra:
You could just rebase and force-push and keep the PR opened. GitHub can handle that.

I'm fine either way, just a suggestion to save you unnecessary work :)

PR #20, opened 2016-10-05
DrFlacAudioImporter: Add 24/32 bit support

Alicemargatroid:
DrFlacAudioImporter can now load any format supported by the reference FLAC encoder, and correctly loads them to the same specification as WavAudioImporter.


Alicemargatroid:
@mosra It looks like they are caused by my FLAC files being too big? Any idea on that?

mosra:
It fails on the quad16 test for which the test file is 60 kB, I doubt that the decompression would inflate the file to over 16 MB... I also doubt this would be due to some endianness issue (a value interpreted as some gigabytes instead of kilobytes) -- that would blow up also on desktop.

I suspect this might be caused by unaligned reads/writes (yes, of all places, that's actually a thing here) -- you are allocating a char array that theoretically might get allocated on a odd address and then reinterpreting it as an int so then there might be some suspicious behavior. It might be also possible that the DrFlac is doing some unaligned reads -- majority of platforms does not care about unaligned reads anymore so such bug might get unnoticed until now.

Can you try e.g. verifying that the allocated array is aligned on 4 byte boundary or that the sample count is sane?

Also, maybe for better sleep it might be wise to actually check the returned array sizes in the tests -- now I see just checking the first N bytes, but not caring how the returned array is actually big. (Yes, I'm also guilty in this, my poor initial WavAudoImporter tests also didn't check for that..)

Alicemargatroid:
@mosra I'll give it a look, unaligned looks like a good place to start.

As well, I think a lot of Magnum code is not compatible with strict aliasing; you aren't really supposed to reinterpret_cast pointers to anything but char*. Worth looking into as a possible problem throughout the source code: http://blog.regehr.org/archives/1307

mosra:
Strict aliasing -- yes, I know about it. I tried to fight it with [Corrade::Utility::bitCast()](http://mosra.cz/blog/corrade-doc/namespaceCorrade_1_1Utility.html#a3831cece822dfde455336664676f6867) at some point (inspired by some utility function that Chromium uses to avoid strict aliasing issues), but that's such a *huge* workaround (it involves memcpy) that I gave up on it. Another "well-known" workaround is to have an union, save to one field and then read from another. But The Standard says it's undefined behavior, so ...

Counter-argument from Linus: https://lkml.org/lkml/2003/2/26/158 ;)

Also, a lot of third-party code that is used here (e.g. the `stb_*` libraries) explicitly claims that it *doesn't* work with strict aliasing and they have no plans in supporting that.

So maybe I should just document whether the code supports it or not. It's just a problem of GCC if I remember correctly, no other compiler (even Clang) is doing such crazy optimizations.

Alicemargatroid:
@mosra You actually have it backwards; every compiler _except Clang_ is doing crazy optimizations with it! They wrote a little blog post about it and other UB that might be informative: http://blog.llvm.org/2011/05/what-every-c-programmer-should-know.html

Ironically, the union trick _while non-standard_ is _supported everywhere_, while the memcpy fix _while standard_ is pretty awful, I agree. I think because the union trick is so well tolerated _because_ of Linus (it is used all the time in Linux). I'd use that for bitcast.

There is a flag in most compilers to disable strict aliasing; might be worth looking into. Looks like this problem with Travis is due to too large of file sizes though; DrFlac increases the size of the binaries, because it loads everything as 32-bit PCM.

mosra:
Ah, okay, so I did not remembered correctly :) In my limited world view, I see just GCC doing it, Clang not and MSVC also not (?). GCC has `-fno-strict-aliasing` that I desperately wanted to enable when fighting some annoying bugs, but didn't do it yet.

Re the too large files: so if you pass the suggested flag to Emscripten, do the test pass for you at least locally? If yes, then I might just enable that this for the particular test and be done with it.

Alicemargatroid:
@mosra Yeah, MSVC only recently started taking advantage of some of those rules, so you probably haven't hit them too much yet.

The tests pass for me on Emscripten; as well, I have worked to make the files smaller, and can shrink the size of DrWavAudioImporter's files once this is accepted. We could also just disable some of the tests for DrFlac to avoid the issue.

mosra:
Cherry-picked in 9db292233580d54d239b5f4ceffa27a2707070ec. Sorry it took so long.

I stripped the `quad16` and `surround51Channel24` files even more, limiting them to 50k samples. The `quad16` file had originally around 5M samples per channel, which after expansion to 32bit was almost 100 MB of memory and I didn't want to increase the Emscripten memory heap size that high just for that one test case.

PR #23, opened 2016-10-23
DevilImageImporter: Fast and full featured image loader and manipulator

Alicemargatroid:
Faster than libjpg or libpng, more image formats than StbImageLoader, and just as optimized as DdsLoader, it's DevilImageLoader! Using the highly optimized and easy to find DevIL library, DevilImageLoader provides all sorts of goodies at minimum price.

Optimized binaries for Windows are available, but I have not integrated DevilImageLoader into the CI builds as I don't know how to test if they have the DevIL binary or not.


src/MagnumPlugins/DevilImageImporter/DevilImageImporter.h
```diff
@@ -0,0 +1,98 @@
...
+#endif
+
+namespace Magnum { namespace Trade {
+
+/**
+@brief JPEG importer plugin
+
+Supports RGB or grayscale images with 8 bits per channel.
+
+This plugin depends on **libJPEG** library and is built if `WITH_DEVILIMAGEIMPORTER`
```
Squareys: The doc probably needs to be updated here ;)

Alicemargatroid: You are right, always the nitpicky details haha. Let me patch that up.

mosra:
Merged in 25648ff9ff0a36eda2b9643e892095042b9682d3. I made some minor modifications and kept some TODOs (such as 16-bit, floating-point and compressed image import) for myself, will try to address them later.

Thanks a lot for the outstanding work!

PR #30, opened 2017-04-02
Modernize OpenDdl and StanfordImporter

alicemargatroid:
Real small PR, replaces push_back with emplace_back and adds some qualifiers to improve performance

### DIFF
--- a/src/MagnumPlugins/OpenGexImporter/OpenDdl/Document.h
+++ b/src/MagnumPlugins/OpenGexImporter/OpenDdl/Document.h
@@ -428,7 +428,7 @@ class MAGNUM_TRADE_OPENGEXIMPORTER_EXPORT Document {
 
         MAGNUM_TRADE_OPENGEXIMPORTER_LOCAL std::size_t dereference(std::size_t originatingStructure, Containers::ArrayView<const char> reference) const;
 
-        MAGNUM_TRADE_OPENGEXIMPORTER_LOCAL bool validateLevel(std::optional<Structure> first, Containers::ArrayView<const std::pair<Int, std::pair<Int, Int>>> allowedStructures, Containers::ArrayView<const Validation::Structure> structures, std::vector<Int>& counts) const;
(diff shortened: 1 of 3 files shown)


src/MagnumPlugins/OpenGexImporter/OpenDdl/Document.h
```diff
@@ -428,7 +428,7 @@ class MAGNUM_TRADE_OPENGEXIMPORTER_EXPORT Document {
 
         MAGNUM_TRADE_OPENGEXIMPORTER_LOCAL std::size_t dereference(std::size_t originatingStructure, Containers::ArrayView<const char> reference) const;
 
-        MAGNUM_TRADE_OPENGEXIMPORTER_LOCAL bool validateLevel(std::optional<Structure> first, Containers::ArrayView<const std::pair<Int, std::pair<Int, Int>>> allowedStructures, Containers::ArrayView<const Validation::Structure> structures, std::vector<Int>& counts) const;
+        MAGNUM_TRADE_OPENGEXIMPORTER_LOCAL bool validateLevel(const std::optional<Structure>& first, Containers::ArrayView<const std::pair<Int, std::pair<Int, Int>>> allowedStructures, Containers::ArrayView<const Validation::Structure> structures, std::vector<Int>& counts) const;
```
mosra: Hmm. The `Structure` is just two pointers in a POD, so I thought it is completely okay to pass it via value. But the `std::optional` probably adds some nontrivial boilerplate around so it's not as free to copy anymore.

mosra:
Merged, thank you!

src/MagnumPlugins/OpenGexImporter/OpenDdl/Document.h
Squareys: *POD*? Plain old ...? 😅

LB--: @Squareys Plain Old Data, unless you're suggesting something else?

mosra: Data. :D

alicemargatroid: The current term would be "Standard Layout"; we're working in C++11 I think.
But std::optional is not standard layout, I believe.

PR #31, opened 2017-05-11
Assimp importer

Squareys:
**Hello everybody!**

This is some code I started working on during the magnum hackathon in January: A scene importer plugin which makes use of assimp!
I didn't have much time until recently to start cleaning this up, but now that it finally compiles, here's a pullrequest with the *early work in progress*.

Since this is based off of the OpenGexImporter code, there may be copy&paste errors, if you find any, please do leave a comment.

Cheers, Jonathan.

### Current State

 - [x] Compiles
 - After coming across [a bunch of issues](https://github.com/assimp/assimp/issues/created_by/squareys), I decided to manually create files that assimp loads and test the mapping between assimp structures to magnum `Trade::*` structures rather than testing assimp functionality.

### Todo List

 - [x] Tests
 - [x] Fix the 3 TODO comments
 - [x] Implement texture property import
 - [x] More documentation required in "Behaviour and Limitations" section
 - [ ] Enable on CI (requires getting assimp on there somehow)
    - [x] Travis
    - [ ] AppVeyor


mosra:
Regarding CI: I see that Travis has `libassimp-dev` whitelisted: https://github.com/travis-ci/apt-package-whitelist/blob/be9250651ea376a6afcbdba87c3b63f25a491e95/ubuntu-precise#L4334

modules/FindAssimp.cmake
```diff
@@ -0,0 +1,92 @@
...
+        set(ASSIMP_MSVC_VERSION "vc140")
+    else()
+        message(ERROR "Unsupported MSVC version.")
+    endif()
+
+    if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+        set(ASSIMP_LIBRARY_DIR "lib64")
+    elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
+        set(ASSIMP_LIBRARY_DIR "lib32")
+    endif(CMAKE_SIZEOF_VOID_P EQUAL 8)
```
mosra: `endif()` without the inside, please (I think new CMake versions even warn about that) ;)

modules/FindAssimp.cmake
```diff
@@ -0,0 +1,92 @@
...
+
+    find_library(ASSIMP_LIBRARY_RELEASE assimp-${ASSIMP_MSVC_VERSION}-mt.lib PATHS ${ASSIMP_LIBRARY_DIR})
+    find_library(ASSIMP_LIBRARY_DEBUG assimp-${ASSIMP_MSVC_VERSION}-mtd.lib PATHS ${ASSIMP_LIBRARY_DIR})
+    set(ASSIMP_LIBRARY optimized ${ASSIMP_LIBRARY_RELEASE} debug ${ASSIMP_LIBRARY_DEBUG})
```
mosra: I would just not define `ASSIMP_LIBRARY` at all -- it's not meant to be used anyway. The same for `ASSIMP_LIBRARIES`.

modules/FindAssimp.cmake
```diff
@@ -0,0 +1,92 @@
...
+#  ASSIMP_LIBRARY         - Assimp library
+#  ASSIMP_LIBRARIES       - Same as ASSIMP_LIBRARY
+#  ASSIMP_INCLUDE_DIR     - Include dir
+#
+
+#
+#   This file is part of Magnum.
+#
+#   Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016
+#             Vladimír Vondruš <mosra@centrum.cz>
```
mosra: Could you put 2017 in all the copyright headers so it's in sync with all the other code? Thank you :)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,375 @@
...
+        return ImageData2D(PixelFormat::RGBA, PixelType::UnsignedByte, dimensions, std::move(data), texture);
+    }
+}
+
+const void* AssimpImporter::doImporterState() const {
+    return _scene;
+}
+
+void AssimpImporter::createNodeIndices() {
+    if(_nodes.empty()) {
```
mosra: :thinking: 

`if(!_nodes.empty())`? :)

Squareys: Indeed :D

- [ ] Done

modules/FindAssimp.cmake
Squareys: Sure 👍 

- [ ] Done

modules/FindAssimp.cmake
Squareys: - [ ] Done

modules/FindAssimp.cmake
Squareys: So... use a temp variable instead, for adding that to the imported target later?

mosra: In the imported target you are using `ASSIMP_LIBRARY` only if `ASSIMP_LIBRARY_DEBUG` / `_RELEASE` is not defined, so I think not defining it in this case shouldn't be a problem.

Squareys: 🤔 
I use `ASSIMP_LIBRARY` if `ASSIMP_LIBRARY_DEBUG / _RELEASE` are not *bth* defined. Also I use it with `find_package_handle_standard_args`. I don't want to require both debug and release libraries to be present. I will try specifying both, maybe it knows how to handle that intelligently.

Squareys: I think I got it now. But didn't test it with missing Release xor Debug yet.

mosra:
Just a note to not forget later: Travis and most of the systems out there would probably still have the crashy version of Assimp loaded, so in the tests it's needed to skip such a case if `OpenDDLParser::getVersion()` is less than 0.4.0 (I hope this API is public, though).

modules/FindAssimp.cmake
```diff
@@ -0,0 +1,92 @@
...
+        set(ASSIMP_LIBRARY_DIR "lib64")
+    elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
+        set(ASSIMP_LIBRARY_DIR "lib32")
+    endif()
+
+    find_library(ASSIMP_LIBRARY_RELEASE assimp-${ASSIMP_MSVC_VERSION}-mt.lib PATHS ${ASSIMP_LIBRARY_DIR})
+    find_library(ASSIMP_LIBRARY_DEBUG assimp-${ASSIMP_MSVC_VERSION}-mtd.lib PATHS ${ASSIMP_LIBRARY_DIR})
+else()
+    find_library(ASSIMP_LIBRARY_RELEASE libassimp PATHS lib)
+    find_library(ASSIMP_LIBRARY_DEBUG libassimpd PATHS lib)
```
mosra: This should be just `assimp` and `assimpd` and PATHS doesn't need to be there at all (this is the default).

modules/FindAssimp.cmake
```diff
@@ -0,0 +1,92 @@
...
+endif()
+
+include(SelectLibraryConfigurations)
+select_library_configurations(Assimp)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Assimp DEFAULT_MSG
+    ASSIMP_LIBRARY_DEBUG
+    ASSIMP_LIBRARY_RELEASE
```
mosra: Use just `ASSIMP_LIBRARY` here (it's coming from `SelectLibraryConfigurations`) -- this will make it work properly on systems that don't have both versions installed.

modules/FindAssimp.cmake
```diff
@@ -0,0 +1,92 @@
...
+if(NOT TARGET Assimp::Assimp)
+    add_library(Assimp::Assimp UNKNOWN IMPORTED)
+
+    if(ASSIMP_LIBRARY_DEBUG AND ASSIMP_LIBRARY_RELEASE)
+        set_target_properties(Assimp::Assimp PROPERTIES
+            IMPORTED_LOCATION_DEBUG ${ASSIMP_LIBRARY_DEBUG}
+            IMPORTED_LOCATION_RELEASE ${ASSIMP_LIBRARY_RELEASE})
+    else()
+        set_target_properties(Assimp::Assimp PROPERTIES
+            IMPORTED_LOCATION ${ASSIMP_LIBRARY_DEBUG} ${ASSIMP_LIBRARY_RELEASE})
```
mosra: Should be just `${ASSIMP_LIBRARY}`.

modules/FindAssimp.cmake
```diff
@@ -0,0 +1,92 @@
...
+    endif()
+
+    set_target_properties(Assimp::Assimp PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES ${ASSIMP_INCLUDE_DIR})
+endif()
+
+set(ASSIMP_LIBRARIES ${ASSIMP_LIBRARY})
```
mosra: I would *not* set this variable at all. The user should use the imported targets instead.

(Also remove it from the docs above.)

package/ci/travis.yml
```diff
@@ -14,6 +14,7 @@ addons:
     - libpng12-dev
     - libqt4-dev
     - libdevil-dev
+    - libassimp-dev
 
 matrix:
```
mosra: This is hopefully enough to make it working on macOS as well:

```patch
diff --git a/package/ci/travis.yml b/package/ci/travis.yml
index 8906dd8..0e43084 100644
--- a/package/ci/travis.yml
+++ b/package/ci/travis.yml
@@ -162,8 +162,8 @@ install:
 # works. Replace with just `brew install emscripten` when sane again
 - if [ "$TRAVIS_OS_NAME" == "osx" ] && [ "$TARGET" == "emscripten" ]; then brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/53f53f6498ba0f507c443dd4f0c6217937f1ddf2/Formula/emscripten.rb && export LLVM=/usr/local/opt/emscripten/libexec/llvm/bin && emcc; fi
 
-# HarfBuzz
-- if [ "$TRAVIS_OS_NAME" == "osx" ] && [ "$TARGET" == "desktop" ]; then brew install harfbuzz; fi
+# HarfBuzz, Assimp
+- if [ "$TRAVIS_OS_NAME" == "osx" ] && [ "$TARGET" == "desktop" ]; then brew install harfbuzz assimp; fi
 
 script:
 - if [ "$TRAVIS_OS_NAME" == "linux" ] && ( [ "$TARGET" == "desktop" ] || [ "$TARGET" == "desktop-sanitizers" ] ); then ./package/ci/travis-desktop.sh; fi


```

src/MagnumPlugins/AssimpImporter/AssimpImporter.conf
```diff
@@ -0,0 +1 @@
+depends=AnyImageImporter
```
mosra: More of a reminder for myself (I can do that post-merge):

There should be a bunch of `provides=` lines here, based on what all formats this supports (at least it should alias *all* the existing OBJ, COLLADA, Stanford, .. importers), mentioning this also in the docs and then finally updating `AnyImporter` to detect and handle the new formats.

modules/FindAssimp.cmake
Squareys: ✔️

modules/FindAssimp.cmake
Squareys: ✔️

modules/FindAssimp.cmake
Squareys: ✔️

modules/FindAssimp.cmake
Squareys: ✔️

package/ci/travis.yml
Squareys: ✔️

src/MagnumPlugins/AssimpImporter/CMakeLists.txt
```diff
@@ -0,0 +1,104 @@
...
+    "${MAGNUM_PLUGINS_IMPORTER_DEBUG_BINARY_INSTALL_DIR};${MAGNUM_PLUGINS_IMPORTER_DEBUG_LIBRARY_INSTALL_DIR}"
+    "${MAGNUM_PLUGINS_IMPORTER_RELEASE_BINARY_INSTALL_DIR};${MAGNUM_PLUGINS_IMPORTER_RELEASE_LIBRARY_INSTALL_DIR}"
+    AssimpImporter.conf
+    $<TARGET_OBJECTS:AssimpImporterObjects>
+    pluginRegistration.cpp)
+if(BUILD_STATIC_PIC)
+    set_target_properties(AssimpImporter PROPERTIES POSITION_INDEPENDENT_CODE ON)
+endif()
+target_include_directories(AssimpImporter PUBLIC ${PROJECT_SOURCE_DIR}/src)
+target_link_libraries(AssimpImporter Magnum::Magnum Assimp::Assimp AnyImageImporter)
```
mosra: This will fix the Travis failure:

```cmake
target_link_libraries(AssimpImporter Magnum::Magnum Assimp::Assimp)
if(CORRADE_TARGET_WINDOWS)
    target_link_libraries(AssimpImporter AnyImageImporter)
endif()
```

To explain: the plugins are built as `MODULE` libraries on almost all platforms except Windows (where I had some issues *way back* and didn't have time to look into them yet). That means they *can* have unresolved symbols, because the plugin manager will handle that on runtime. So it's not needed (not even possible) to link AnyImageImporter to this.

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
```diff
@@ -0,0 +1,257 @@
+/*
```
mosra: What about the `src/MagnumPlugins/AssimpImporter/Test/.Test.cpp.swp` file here? ;)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include "AssimpImporter.h"
+
+#include <algorithm>
+#include <cstring> /* std::memcpy */
+#include <limits>
```
mosra: Is this needed?

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+#include <unordered_map>
+#include <Corrade/Containers/ArrayView.h>
+#include <Corrade/Utility/Directory.h>
+#include <Magnum/Mesh.h>
+#include <Magnum/Math/Quaternion.h>
```
mosra: Is this needed? :)

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
```diff
@@ -0,0 +1,257 @@
...
+*/
+
+#include <sstream>
+#include <Corrade/Containers/ArrayView.h>
+#include <Corrade/TestSuite/Tester.h>
+#include <Corrade/TestSuite/Compare/Numeric.h>
+#include <Corrade/TestSuite/Compare/Container.h>
+#include <Corrade/Utility/Directory.h>
+#include <Magnum/Mesh.h>
+#include <Magnum/Math/Quaternion.h>
```
mosra: Is this needed? :)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    static aiColor3D to(const Vector<3, Float>& other) {
+        return {other[0], other[1], other[2]};
+    }
+};
+
+}}}
+
+namespace Magnum { namespace Trade {
+
+using namespace Magnum::Math::Literals;
```
mosra: I don't see this being used anywhere.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    _textureIndices.clear();
+}
+
+Int AssimpImporter::doDefaultScene() { return 0; }
+
+UnsignedInt AssimpImporter::doSceneCount() const { return _scene->mRootNode ? 1 : 0; }
+
+std::optional<SceneData> AssimpImporter::doScene(UnsignedInt) {
+    const aiNode* root = _scene->mRootNode;
+    return SceneData{{}, {_nodeIndices[root]}, root};
```
mosra: The root node should be equivalent to the scene node, so SceneData should contain directly the children of the root node.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    }
+
+    /* Import indices */
+    std::vector<UnsignedInt> indices;
+    indices.reserve(mesh->mNumFaces*3);
+
+    for(size_t faceIndex = 0; faceIndex < mesh->mNumFaces; ++faceIndex) {
+        const aiFace& face = mesh->mFaces[faceIndex];
+
+        CORRADE_ASSERT(face.mNumIndices == 3, "Expected number of indices per face to always be 3, since using triangulation post processing of assimp", {});
```
mosra: Please prefix the assert message with the function name the same way as the error messages.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    if(!_nodes.empty()) {
+        /* Already done previously */
+        return;
+    }
+
+    /* reserve memory for the amount of nodes we know will at least be present */
+    aiNode* root = _scene->mRootNode;
+    if(!root) {
+        return; /* Nothing to do */
+    }
```
mosra: Coding style: this could be all one-liners with `return` on the same line, I don't enforce braces everywhere.

src/MagnumPlugins/AssimpImporter/AssimpImporter.h
```diff
@@ -0,0 +1,147 @@
...
+and @ref plugins for more information.
+
+### Behaviour and limitations
+
+#### Material import
+
+- Only materials with shading mode `aiShadingMode_Phong` are supported
+- Only the first diffuse/specular/ambient texture is loaded
+
+
```
mosra: Mentioning the types of `importerState()` return values here would be nice (look in the OGEX importer for an example).

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+using namespace Magnum::Math::Literals;
+
+/* Stream implementation for outputting assimp log messages to Debug() */
+class MagnumDebugStream: public Assimp::LogStream {
+public:
+    void write(const char* message) override {
+        Debug(Debug::Flag::NoNewlineAtTheEnd) << Debug::color(Debug::Color::Yellow) << "assimp:" << message;
+    }
+};
```
mosra: What about wrapping this all in some `#define MAGNUM_ASSIMPIMPORTER_DEBUG`, along with the include etc.?

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+
+        std::unique_ptr<Trade::ObjectData3D> lightObject = importer.object3D(i + 1);
+        CORRADE_COMPARE(lightObject->instanceType(), ObjectInstanceType3D::Light);
+        CORRADE_COMPARE(lightObject->instance(), i);
+    }
+
+    if(aiGetVersionMajor() <= 3 && aiGetVersionMinor() < 1) {
+        /* Cannot load lights with undefined light type yet */
+        return;
```
mosra: I would use `CORRADE_SKIP()` here so it's visible in the output. Also, like I mentioned on Gitter, use some combined version integer for comparison (this would fail for theoretical version 2.9 for example).

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
```diff
@@ -0,0 +1,257 @@
...
+    CORRADE_COMPARE(importer.defaultScene(), 0);
+    std::optional<Trade::SceneData> data = importer.scene(0);
+    CORRADE_VERIFY(data);
+
+    CORRADE_COMPARE(data->children2D(), {});
+    CORRADE_COMPARE(data->children3D(), {0});
+
+    std::unique_ptr<Trade::ObjectData3D> rootObject = importer.object3D(0);
+    CORRADE_COMPARE(rootObject->children(), {1});
+    //CORRADE_COMPARE(rootObject->instanceType(), ObjectInstanceType3D::Mesh); // Should be Empty, but Assimp creates mesh for root node for some reason
```
mosra: `CORRADE_EXPECT_FAIL()` then?

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
```diff
@@ -0,0 +1,257 @@
...
+    CORRADE_COMPARE(explicitRootObject->children(), {2});
+    CORRADE_COMPARE(explicitRootObject->instanceType(), ObjectInstanceType3D::Empty);
+    CORRADE_COMPARE(explicitRootObject->transformation(), Matrix4());
+
+    std::unique_ptr<Trade::ObjectData3D> childObject = importer.object3D(2);
+    CORRADE_COMPARE(childObject->transformation(),
+                    Matrix4({0.813798f, 0.469846f, -0.34202f, 0.0f},
+                            {-0.44097f, 0.882564f, 0.163176f, 0.0f},
+                            {0.378522f, 0.0180283f, 0.925417f, 0.0f},
+                            {1.0f, 2.0f, 3.0f, 1.0f}).transposed());
```
mosra: Huh. Why transposed? This is how it should look (bottom row being zeros).

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
```diff
@@ -0,0 +1,257 @@
...
+}
+
+void AssimpImporterTest::texture() {
+    AssimpImporter importer;
+    CORRADE_VERIFY(importer.openFile(Utility::Directory::join(ASSIMPIMPORTER_TEST_DIR, "embedded_texture.blend")));
```
mosra: Naming scheme nitpick (sorry) -- `embedded-texture.blend`? I'm not using underscores anywhere.

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+    CORRADE_COMPARE(importer.textureCount(), 1);
+    CORRADE_COMPARE(importer.image2DCount(), 1);
+
+    CORRADE_SKIP("This test requires using AssimpImporter via a plugin manager which has access to PngImporter for blender embedded image data.");
```
mosra: Shouldn't be a problem to manage -- look in the OGEX importer tests how it's done.

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
Squareys: Uh oh, seeems like I committed the vim swap file ~~accidentally~~ (on purpose, of course, so that everyone thinkgs I'm a vim-magician, hah :P )

Squareys: ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: Nope ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: Nope ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+namespace Magnum { namespace Math { namespace Implementation {
+
+template<> struct VectorConverter<3, Float, aiColor3D> {
+    static Vector<3, Float> from(const aiColor3D& other) {
+        return {other.r, other.g, other.b};
+    }
+
+    static aiColor3D to(const Vector<3, Float>& other) {
+        return {other[0], other[1], other[2]};
+    }
```
Squareys: ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+void AssimpImporter::doOpenFile(const std::string& filename) {
+    _scene = _importer.ReadFile(filename, aiProcess_Triangulate | aiProcess_SortByPType);
+
+    /* If succeeded, save file path for later */
+    if (_scene) {
+        _filePath = Utility::Directory::path(filename);
+        createNodeIndices();
+        createMaterialIndices();
+    }
+}
```
Squareys: Because I load the file differently in doOpenFile() (As assimp guesses file format based on file ending). Or is this only for the create*Indices() part?

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    _filePath = std::nullopt;
+
+    _nodes.clear();
+    _nodeIndices.clear();
+    _nodeInstances.clear();
+    _materialIndicesForName.clear();
+    _textureIndices.clear();
+}
+
+Int AssimpImporter::doDefaultScene() { return 0; }
```
Squareys: I would guess there always is a root node an therefore always a default scene. As I am not entirely sure, though, I will make this `_scene->m_rootNode ? 0 : -1;`

Squareys: Did that for doSceneCount, too, so this also makes it consistent

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: I think the root node can have transformation...

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+        children.push_back(_nodeIndices[child]);
+    }
+
+    const Matrix4 transformation = Matrix4::from(reinterpret_cast<const float*>(&node->mTransformation));
+
+    auto instance = _nodeInstances.find(node);
+    if(instance != _nodeInstances.end()) {
+        return std::unique_ptr<ObjectData3D>{new ObjectData3D(children, transformation, (*instance).second.first, (*instance).second.second, node)};
+    }
+    return std::unique_ptr<ObjectData3D>{new ObjectData3D(children, transformation, node)};
```
Squareys: Done ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+    LightData::Type lightType;
+    if(l->mType == aiLightSource_DIRECTIONAL) {
+        lightType = LightData::Type::Infinite;
+    } else if(l->mType == aiLightSource_POINT) {
+        lightType = LightData::Type::Point;
+    } else if(l->mType == aiLightSource_SPOT) {
+        lightType = LightData::Type::Spot;
+    } else {
+        Error() << "Trade::AssimpImporter::light(): Undefined light type is not supported.";
```
Squareys: aiLightSource_UNDEFINED?

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
mosra: But then the same transformation would be applied to the camera as well, so no visible difference..?

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+        return {};
+    }
+
+    Color3 ambientColor = Color3::from(reinterpret_cast<const float*>(&l->mColorDiffuse));
```
Squareys: I like the converters, they hide the ugly casting, which is also long to type

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
mosra: Yes, or some garbage if assimp goes haywire :)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+    normals.emplace_back(); // FIXME(C++17) assign to pos directly
+    auto& norms = normals[0];
+    norms.reserve(mesh->mNumVertices);
+    norms.insert(norms.end(), reinterpret_cast<Vector3*>(mesh->mNormals),
+            reinterpret_cast<Vector3*>(mesh->mNormals) + mesh->mNumVertices*sizeof(Vector3));
+
+    /** @todo only first uv layer (or "channel") supported) */
+    textureCoordinates.reserve(mesh->GetNumUVChannels());
+    for (size_t layer = 0; layer < mesh->GetNumUVChannels(); ++layer) {
```
Squareys: Done ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+        textureCoordinates.emplace_back();
+        auto& texCoords = textureCoordinates[layer];
+        texCoords.reserve(mesh->mNumVertices);
+        for(size_t i = 0; i < mesh->mNumVertices; ++i) {
+            /** GCC 4.7 has a problem with .x/.y here */
+            texCoords.emplace_back(mesh->mTextureCoords[layer][i][0], mesh->mTextureCoords[layer][i][1]);
```
Squareys: MeshData doesn't support 3D texture coords, so I though, this at least gives as much information from the file as possible... but, yes there is numUVComponents in the aiMesh structure

mosra: for now i would just put only the 2D texture coords in

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: Done ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    std::vector<std::vector<Vector2>> textureCoordinates;
+    std::vector<std::vector<Vector3>> normals;
+    std::vector<std::vector<Color4>> colors;
+
+    /* Import positions */
+    positions.emplace_back(); // FIXME(C++17) assign to pos directly
+    auto& pos = positions[0];
+    pos.reserve(mesh->mNumVertices);
+    pos.insert(pos.end(), reinterpret_cast<Vector3*>(mesh->mVertices),
+            reinterpret_cast<Vector3*>(mesh->mVertices) + mesh->mNumVertices);
```
Squareys: ✔️  Done

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    std::vector<UnsignedInt> indices;
+    indices.reserve(mesh->mNumFaces*3);
+
+    for(size_t faceIndex = 0; faceIndex < mesh->mNumFaces; ++faceIndex) {
+        const aiFace& face = mesh->mFaces[faceIndex];
+
+        CORRADE_ASSERT(face.mNumIndices == 3, "Expected number of indices per face to always be 3, since using triangulation post processing of assimp", {});
+        for(const int i: {0, 1, 2}) {
+            indices.emplace_back(face.mIndices[i]);
+        }
```
Squareys: Done ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+std::unique_ptr<AbstractMaterialData> AssimpImporter::doMaterial(const UnsignedInt id) {
+    /* Put things together */
+    const aiMaterial* mat = _scene->mMaterials[id];
+
+    aiShadingMode shadingMode;
+
+    if(mat->Get(AI_MATKEY_SHADING_MODEL, shadingMode) == AI_SUCCESS) {
+        if(shadingMode != aiShadingMode_Phong) {
+            Error() << "Unsupported shading mode";
```
Squareys: Done ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    Sampler::Wrapping wrappingU = Sampler::Wrapping::ClampToEdge;
+    Sampler::Wrapping wrappingV = Sampler::Wrapping::ClampToEdge;
+    constexpr Sampler::Wrapping wrappingW = Sampler::Wrapping::ClampToEdge;
+
+    auto toWrapping = [](aiTextureMapMode mapMode) {
+        switch (mapMode) {
+        case aiTextureMapMode_Wrap:
+            return Sampler::Wrapping::Repeat;
+        case aiTextureMapMode_Decal:
+            Warning() << "Trade::AssimpImporter::texture(): no wrapping enum to match aiTextureMapMode_Decal, using Sampler::Wrapping::ClampToEdge.";
```
Squareys: Ugh ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+    const aiMaterial* mat = _textures[id].first;
+    const aiTextureType type = _textures[id].second;
+    if(mat->Get(AI_MATKEY_MAPPINGMODE_U(type, 0), mapMode) == AI_SUCCESS) {
+        wrappingU = toWrapping(mapMode);
+    }
+    if(mat->Get(AI_MATKEY_MAPPINGMODE_V(type, 0), mapMode) == AI_SUCCESS) {
+        wrappingV = toWrapping(mapMode);
+    }
+
+    return TextureData{TextureData::Type::Texture2D, filterU, filterV, {Sampler::Mipmap::Linear}, {wrappingU, wrappingV, wrappingW}, id, &_textures[id]};
```
Squareys: I am expecting this may be loadable some day, or I possibly missed something. Then it's easier to add...
Since the constants are rather long, it makes that return statement rather long and less readible. As the compiler will optimize it away, I didn't see any harm :D

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+        Error() << "Trade::AssimpImporter::image2D(): Error getting path for texture with id" << id;
+        return std::nullopt;
+    }
+
+    std::string path = texturePath.C_Str();
+    if(path[0] == '*') {
+        /* Load embedded texture */
+        char* err;
+        const Int index = Int(strtol(path.substr(0).c_str(), &err, 10));
+        if(err == nullptr) {
```
Squareys: Yes, it's the pointer to the first char after the number. If no number, no position after, hence nullptr.
Maybe the naming of my var is confusing there

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+            } else if(texture->CheckFormat("jpg")) {
+                importerName = "JpegImporter";
+            } else if(texture->CheckFormat("pcx")) {
+                importerName = "PxcImporter";
+            } else if(texture->CheckFormat("png") || std::strncmp(textureData.suffix(1).data(), "PNG", 3) == 0) {
+                importerName = "PngImporter";
+            } else {
+                Error() << "Trade::AssimpImporter::image2D(): Could not detect filetype of embedded data.";
+                return std::nullopt;
+            }
```
Squareys: `AbstractImporter::openData(data, "optional-extension-hint")` would also be useful! ;)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+            importer->openData(textureData);
+            return importer->image2D(0);
+        } else {
+            /* Uncompressed image data */
+            const Vector2i dimensions{Int(texture->mHeight), Int(texture->mWidth)};
+            size_t size = dimensions.product()*4;
+            Containers::Array<char> data{Containers::NoInit, size};
+            std::memcpy(texture->pcData, data.data(), size);
```
Squareys: Hah, it's really confusing: `unsigned char b,g,r,a;` is the definition in the code.
Not even sure if it is okay to assume any order here. There is a function to convert to `aiColor4D`, which I will use in this case.

mosra: `Math::swizzle()`?

Squareys: Alright, swizzle brong bgra to rgba as discussed via gitter

Squareys: Done ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
mosra: [But](http://en.cppreference.com/w/cpp/string/byte/strtol):

> If the str is empty or does not have the expected form, no conversion is performed, and (if str_end is not NULL) the value of str is stored in the object pointed to by str_end.

I'm confused (and yes, the naming is confusing also).

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: Done ✔️

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
Squareys: Done ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
mosra: I have a different idea: [per-plugin configuration](https://github.com/mosra/corrade/projects/6#card-577831)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+    aiString texturePath;
+    if(mat->Get(AI_MATKEY_TEXTURE(type, 0), texturePath) != AI_SUCCESS) {
+        Error() << "Trade::AssimpImporter::image2D(): Error getting path for texture with id" << id;
+        return std::nullopt;
+    }
+
+    std::string path = texturePath.C_Str();
+    if(path[0] == '*') {
+        /* Load embedded texture */
```
Squareys: But doesn't that change the meaning?
~~~
if (path[0] == '*') => Load the embedded texture
else => load external texture...
~~~

I could do something like
~~~
/* Load embedded or external texture? */
if(path[0] == ...
else ...
~~~

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: Done ✔️

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: Oh, well, then `(err == nullptr || err == thatString)` is the error condition

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: (Discussed via gitter) ✔️ Done

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
Squareys: I believe the root can potentially have a mesh, though 🤔

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
Squareys: I think the rootObject has a mesh, but I don't know why (aiScene->mRoot has one already)

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
Squareys: Including an entire CMAKE option?

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
Squareys: ✔️  done

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
Squareys: ✔️  Done

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
Squareys: ✔️ Done

src/MagnumPlugins/AssimpImporter/AssimpImporter.h
Squareys: ✔️ Done

src/MagnumPlugins/AssimpImporter/Test/Test.cpp
mosra: no, just a file-local define

mosra: no, just a file-local define

Squareys: ok, done ✔️

mosra:
Merged with a few changes in 63732bc25158458422d6fa00a9fcc867eaa7412a..245db0848f102170ab308e6d86643b833ef14efe.

I reduced the test COLLADA files a bit, integrated the new formats into AnySceneImporter and ignored the memory leak that seems to be present only in the Assimp that's on Travis.

Thank you a lot for this!

Squareys:
Thanks for merging! 🎉

PR #34, opened 2017-06-23
DdsImporter: (most) uncompressed Dxt10 formats support

Squareys:
Hi @mosra, 

here is the work in progress pullrequest for the DXT10 uncompressed formats support in the DdsImporter, which I require for an upcoming magnum example.

Cheers, Jonathan.

TODO:
- [x] (Instanced) Tests for the new supported file formats
- [x] Test for too short files (which cannot contain DXT10 extended header) and fixes in Importer
- [x] Test for unsupported dds compression type (DXT4)
- [x] Test for unsupported DXT10 format
- [x] Update documentation to advertise new formats


src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -143,6 +285,160 @@ void swizzlePixels(const PixelFormat format, Containers::Array<char>& data) {
...
+            return {PixelFormat::RGInteger, PixelType::Int};
+
+        /* RGBA8 Formats */
+        case DxgiFormat::R8G8B8A8Typeless:
+        case DxgiFormat::R8G8B8A8UInt:
+            return {PixelFormat::RGBAInteger, PixelType::UnsignedByte};
+        case DxgiFormat::R8G8B8A8UNorm:
+            return {PixelFormat::RGBA, PixelType::UnsignedByte};
+        //case DxgiFormat::R8G8B8A8UNormSRGB:
+            //return {PixelFormat::SRGBA, PixelType::Byte}; // TODO: No such thing as PixelFormat::SRGBA
```
mosra: There is sRGB + Alpha  *texture format*, but somehow there's no corresponding pixel format for this. Weird, I'll investigate.

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -143,6 +285,160 @@ void swizzlePixels(const PixelFormat format, Containers::Array<char>& data) {
...
+                , PixelType::Short};
+        case DxgiFormat::R16SInt:
+            return {PixelFormat::RedInteger, PixelType::Short};
+
+        /* R8 and A8 Formats */
+        case DxgiFormat::R8Typeless:
+        case DxgiFormat::R8UInt:
+            return {PixelFormat::RedInteger, PixelType::UnsignedByte};
+        case DxgiFormat::A8UNorm:
+            //return {PixelFormat::Alpha, PixelType::UnsignedByte}; //TODO:Alpha? Using same as R8Unorm
```
mosra: GL treats both as single-channel (`PixelFormat::Red`) without any semantic distinction. There was Luminance, Alpha and LuminanceAlpha once, but were deprecated in favor of those that don't have any implicitly associated meaning. The app should know what given texture contains.

I would make it the same as `R8Unorm` but put a detailed chapter into the docs about such mapping "exceptions".

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
```diff
@@ -89,15 +94,6 @@ void DdsImporterTest::unknownFormat() {
...
 }
 
-void DdsImporterTest::unknownCompression() {
-    std::ostringstream out;
-    Error redirectError{&out};
-
-    DdsImporter importer;
-    CORRADE_VERIFY(!importer.openFile(Utility::Directory::join(DDSIMPORTER_TEST_DIR, "unknown_compression.dds")));
-    CORRADE_COMPARE(out.str(), "Trade::DdsImporter::openData(): unknown compression DX10\n");
-}
```
mosra: This would make the "unknown compression" error branch uncovered. I think we should still check this to make sure we handle corrupted files well. Is it possible to craft such a file that fits the criteria?

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
```diff
@@ -241,6 +237,23 @@ void DdsImporterTest::dxt5() {
             TestSuite::Compare::Container);
 }
 
+void DdsImporterTest::dxt10rgba() {
+    DdsImporter importer;
+
+    CORRADE_VERIFY(importer.openFile(Utility::Directory::join(DDSIMPORTER_TEST_DIR "/Dxt10TestFiles", "2D_R32G32_FLOAT.DDS")));
+    CORRADE_VERIFY(importer.openFile(Utility::Directory::join(DDSIMPORTER_TEST_DIR "/Dxt10TestFiles", "2D_R8G8B8A8_UINT.DDS")));
```
mosra: btw., the uppercase extension would make this broken on Linux -- the files have lowercase as far as I see ;)

Squareys: Yes, this is fixed locally. The windows tools output upper case, but I didn't like that ;)

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
Squareys: Yes, of course :)

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
mosra: Hmm. Apparently there's no way to explicitly mark the pixel data as sRGB and in this case it would be just `RGBA`. Numerous sources (e.g. [Apple](https://developer.apple.com/library/content/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/AdoptingOpenGLES3/AdoptingOpenGLES3.html) or e.g. the OpenGL ES 3.2 spec, table 8.2 on page 147) say that the matching pixel format for sRGB(Alpha) is just RGB(A). So I guess that settles it.

This is a bit bad, because only input textures that actually *are* in sRGB should be uploaded to sRGB textures. OTOH, almost all 8-bit color data already are in sRGB, unless someone went an extra step and kept them linear.

More info (huh, I apparently went through all of this already in the past): https://github.com/mosra/magnum/issues/125

Squareys: Would adding a simple flag to `PixelStorage` make sense maybe?

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
Squareys: I will use DXT4 fourcc instead

Squareys:
@mosra I started adding a instanced test. I am unsure how much sense it would make to compare data for every test file (after all, it is loaded with the same code for every of the formats, so I think checking the Pixel Type and Format should be sufficient?)
I would still do the data comparison for at least one of the file formats, of course.

The dxt10unsupportedFormat is still missing a file (coming soon).

mosra:
Right, if it's loaded the same every time, then I would verify the loaded binary data just once, along with all the other properties like width, height, pixel storage, preferably with some more complex case (non-square dimensions, some alignment padding, if that's supported etc.).

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
mosra: PixelStorage is meant to just wrap the data layout parameters, in the extent that OpenGL supports. sRGB doesn't really fit there. I will think about this.

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
```diff
@@ -68,34 +74,41 @@ DdsImporterTest::DdsImporterTest() {
...
               &DdsImporterTest::dxt5,
 
+              &DdsImporterTest::dxt10data,
+              &DdsImporterTest::dxt10tooShort,
+              &DdsImporterTest::dxt10unsupportedFormat,
+
               &DdsImporterTest::useTwice});
+
+    addInstancedTests({&DdsImporterTest::dxt10formats2d}, 46);
+    addInstancedTests({&DdsImporterTest::dxt10formats3d}, 12);
```
mosra: Could you use an enum constant for the size and then also use the same constant below there in the `files` struct? Talking from experience, it will prevent annoying problems when extending/shortening the list later. Details [in the docs](http://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#details).

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
```diff
@@ -241,6 +254,153 @@ void DdsImporterTest::dxt5() {
             TestSuite::Compare::Container);
 }
 
+void DdsImporterTest::dxt10formats2d() {
+    static struct TestFile {
```
mosra: I fear that the `static` keyword implies some unnecessary code generation (mutexes for thread safety). I'm completely okay with this being a `constexpr` anonymous struct in an anonymous namespace *outside* of a function. Details how that might look like [again in the docs](http://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#details) (sorry) ;)

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
```diff
@@ -241,6 +254,153 @@ void DdsImporterTest::dxt5() {
...
+        {"3D_R16G16B16A16_UNORM.dds", PixelFormat::RGBA, PixelType::UnsignedShort},
+        {"3D_R32G32B32A32_FLOAT.dds", PixelFormat::RGBA, PixelType::Float},
+        {"3D_R32G32B32A32_SINT.dds", PixelFormat::RGBAInteger, PixelType::Int},
+        {"3D_R32G32B32A32_UINT.dds", PixelFormat::RGBAInteger, PixelType::UnsignedInt},
+        {"3D_R32G32B32_FLOAT.dds", PixelFormat::RGB, PixelType::Float},
+        {"3D_R32G32B32_SINT.dds", PixelFormat::RGBInteger, PixelType::Int},
+        {"3D_R32G32B32_UINT.dds", PixelFormat::RGBInteger, PixelType::UnsignedInt},
+        {"3D_R32G32_FLOAT.dds", PixelFormat::RG, PixelType::Float},
+    };
+    const TestFile& file = files[testCaseInstanceId()];
```
mosra: could be just `const auto&` and then you can drop the struct name

src/MagnumPlugins/DdsImporter/DdsImporter.h
```diff
@@ -56,6 +56,24 @@ Supports DirectDraw Surface images (`*.dds`) in the following formats:
...
+    - `R16G16B16A16_(TYPELESS|UINT|SINT|FLOAT|UNORM|SNORM)`
+    - `R16G16B16_(TYPELESS|UINT|SINT|FLOAT|UNORM|SNORM)`
+    - `R16G16_(TYPELESS|UINT|SINT|FLOAT|UNORM|SNORM)`
+    - `R16_(TYPELESS|UINT|SINT|FLOAT|UNORM|SNORM)`
+    - `D16_UNORM`
+    - `R8G8B8A8_(TYPELESS|UINT|SINT|UNORM|SNORM)`
+    - `R8G8B8_(TYPELESS|UINT|SINT|UNORM|SNORM)`
+    - `R8G8_(TYPELESS|UINT|SINT|UNORM|SNORM)`
+    - `R8_(TYPELESS|UINT|SINT|UNORM|SNORM)`
+    - `A8_UNORM` (Loaded as @ref PixelFormat::Red)
```
mosra: :+1: 

Can you mention the sRGB thing here as well?

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
```diff
@@ -143,6 +285,158 @@ void swizzlePixels(const PixelFormat format, Containers::Array<char>& data) {
...
+        case DxgiFormat::R32G32SInt:
+            return {PixelFormat::RGInteger, PixelType::Int};
+
+        /* RGBA8 Formats */
+        case DxgiFormat::R8G8B8A8Typeless:
+        case DxgiFormat::R8G8B8A8UInt:
+            return {PixelFormat::RGBAInteger, PixelType::UnsignedByte};
+        case DxgiFormat::R8G8B8A8UNorm:
+        case DxgiFormat::R8G8B8A8UNormSRGB:
+            return {PixelFormat::RGBA, PixelType::UnsignedByte}; // TODO: Propagate sRGB property
```
mosra: Can this be a `/** @todo */` to have Doxygen pick it up?

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
```diff
@@ -241,6 +254,153 @@ void DdsImporterTest::dxt5() {
...
```
mosra: Same here, please :)

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
```diff
@@ -241,6 +254,153 @@ void DdsImporterTest::dxt5() {
...
+        {"3D_R16G16B16A16_SNORM.dds", PixelFormat::RGBA, PixelType::Short},
+        {"3D_R16G16B16A16_UINT.dds", PixelFormat::RGBAInteger, PixelType::UnsignedShort},
+        {"3D_R16G16B16A16_UNORM.dds", PixelFormat::RGBA, PixelType::UnsignedShort},
+        {"3D_R32G32B32A32_FLOAT.dds", PixelFormat::RGBA, PixelType::Float},
+        {"3D_R32G32B32A32_SINT.dds", PixelFormat::RGBAInteger, PixelType::Int},
+        {"3D_R32G32B32A32_UINT.dds", PixelFormat::RGBAInteger, PixelType::UnsignedInt},
+        {"3D_R32G32B32_FLOAT.dds", PixelFormat::RGB, PixelType::Float},
+        {"3D_R32G32B32_SINT.dds", PixelFormat::RGBInteger, PixelType::Int},
+        {"3D_R32G32B32_UINT.dds", PixelFormat::RGBInteger, PixelType::UnsignedInt},
+        {"3D_R32G32_FLOAT.dds", PixelFormat::RG, PixelType::Float},
```
mosra: Just a heads-up: 

- you either need to list all those files explicitly in CMakeLists.txt so they are properly picked up by Android and Emscripten (there is already a list, so just extend it), but I fear this might be too long to fit on the compiler command-line for Emscripten
- or compile them in as a resource

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
Squareys: Damn, vim doesn't highlight it anymore now 😛

mosra: you are holding it wrong ;)

I have a list of keywords to highlight in my editor, I bet vim has that too.

src/MagnumPlugins/DdsImporter/DdsImporter.h
Squareys: Ah, I did indeed forget to list that :D Good catch 👍 
Is listing enough or should I mention it is loaded without a hint at the sRGB-ness?

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
Squareys: Oh, magic! learned something new :D

src/MagnumPlugins/DdsImporter/DdsImporter.h
mosra: > should I mention it is loaded without a hint at the sRGB-ness?

yes please

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
Squareys: Even better... I am using `sizeof(that constexpr array in the anonymous namespace)` now ;)

mosra: that doesn't work like that :P

the only way is `std::size()` and that is since C++17

Squareys: Which uncovered that I still had the 3D file names in the 2d files array 😅

Squareys: (Okay, I understand how that was a bad idea, works with the constant now)

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
Squareys: Done ✔️

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
Squareys: Done ✔️

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
Squareys: Done ✔️

src/MagnumPlugins/DdsImporter/Test/DdsImporterTest.cpp
Squareys: Done ✔️

src/MagnumPlugins/DdsImporter/DdsImporter.h
Squareys: Done ✔️

src/MagnumPlugins/DdsImporter/DdsImporter.cpp
Squareys: Done ✔️

src/MagnumPlugins/DdsImporter/DdsImporter.h
Squareys: Done ✔️

mosra:
Merged, thank you!

PR #35, opened 2017-09-03
Add include directory based on CMAKE variables to assimp

alicemargatroid:
Looks like someone missed the ASSIMP_INCLUDE_DIR variable :-1:

### DIFF
--- a/src/MagnumPlugins/AssimpImporter/CMakeLists.txt
+++ b/src/MagnumPlugins/AssimpImporter/CMakeLists.txt
@@ -49,6 +49,7 @@ target_include_directories(AssimpImporterObjects PUBLIC
     ${PROJECT_SOURCE_DIR}/src
     ${PROJECT_BINARY_DIR}/src)
 target_compile_definitions(AssimpImporterObjects PRIVATE "AssimpImporterObjects_EXPORTS")
+target_include_directories(AssimpImporterObjects SYSTEM PRIVATE ${ASSIMP_INCLUDE_DIR})
 if(NOT BUILD_STATIC OR BUILD_STATIC_PIC)
     set_target_properties(AssimpImporterObjects PROPERTIES POSITION_INDEPENDENT_CODE ON)
 endif()


Squareys:
Assimp related code is probably me! 😅

The problem is somewhere else, though:
~~~
 target_link_libraries(AssimpImporter Magnum::Magnum Assimp::Assimp)
~~~
uses an imported target `Assimp::Assimp` which is defined in `FindAssimp.cmake` and usually automatically sets the include directory! It is probably not found there.

@alicemargatroid Which system are you on? I remember this being enabled on at least one CI, so this must be a platform specific problem.

mosra:
@alicemargatroid oh! terribly sorry about this. Thanks for the patch!

@Squareys the line you mean affects `AssimpImporter`, but not `AssimpImporterObjects` .. in our case (and also in case of the CI, unfortunately) AssImp was installed in the default locations, making the includes "just work". Putting its headers into some non-implicit location exposes this problem.

alicemargatroid:
@mosra No problem, very easy oversight to make.

@Squareys This is not platform specific; I compile everything myself (I don't use the package manager or whatever to install my dependencies), so I need Magnum to know where I've assigned `ASSIMP_INCLUDE_DIR`, otherwise it won't pick up my includes.

Spent a day and a half scratching my head over this before just directly diff'ing the `CMakeLists.txt` of `JpegImporter` vs `AssimpImporter`, and it turned out to be this tiny innocuous line :)

Can confirm this fixes builds for me on Ubuntu, MingW, and OSX.

mosra:
Merged as 530c75c5dd90b1d9ea0f68aeb7c462b7737af545 -- I replaced `${ASSIMP_INCLUDE_DIR}` with `$<TARGET_PROPERTY:Assimp::Assimp,INTERFACE_INCLUDE_DIRECTORIES>` to avoid using internal variables of the Find module.

Hope I didn't break it for you again with that change :D

alicemargatroid:
@mosra Why is such a change not done here then: 
https://github.com/mosra/magnum-plugins/blob/master/src/MagnumPlugins/JpegImporter/CMakeLists.txt#L50

mosra:
Because `FindJPEG.cmake` [doesn't have imported targets](https://cmake.org/cmake/help/v3.9/module/FindJPEG.html) yet ...

PR #37, opened 2018-01-18
Add openState feature to AssimpImporter

costashatz:
This PR adds the `openState` feature to `AssimpImporter` in order to be able to import an `aiScene`.

### DIFF
--- a/src/MagnumPlugins/AssimpImporter/AssimpImporter.h
+++ b/src/MagnumPlugins/AssimpImporter/AssimpImporter.h
@@ -163,6 +163,8 @@ importer state methods:
     -   @ref LightData::importerState() returns `aiLight`
     -   @ref ImageData2D::importerState() may return `aiTexture`, if texture was embedded
         into the loaded file.
+-   @ref openState() expects a pointer to an Assimp scene (i.e., `const aiScene*`)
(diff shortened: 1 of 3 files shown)


src/MagnumPlugins/AssimpImporter/AssimpImporter.h
```diff
@@ -196,6 +196,7 @@ class MAGNUM_TRADE_ASSIMPIMPORTER_EXPORT AssimpImporter: public AbstractImporter
 
         MAGNUM_TRADE_ASSIMPIMPORTER_LOCAL bool doIsOpened() const override;
         MAGNUM_TRADE_ASSIMPIMPORTER_LOCAL void doOpenData(Containers::ArrayView<const char> data) override;
+        MAGNUM_TRADE_ASSIMPIMPORTER_LOCAL void doOpenState(const void* state, const std::string& filePath = {}) override;
```
mosra: No need to have the `= {}` here, it's a private function that's never called from outside.

mosra: One more thing: can you document what type should be passed to the `openState()` function in the "Access to internal importer state" section of the docs?

Besides that, I'm happy :) :+1:

mosra:
Merged, thank you!

PR #40, opened 2018-03-11
TinyGltfImporter

NussknackerXXL:
As discussed with @Squareys here is the first version of the tinyGltfImporter plugin.
If there is anything to improve, i will do what i can :)


Squareys:
@mosra This is finally ready for review. Have fun and thanks in advance! 😉

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
```diff
@@ -0,0 +1,484 @@
...
+}
+
+std::unique_ptr<ObjectData3D> TinyGltfImporter::doObject3D(UnsignedInt id) {
+    const tinygltf::Node& node = _d->model.nodes[id];
+
+    std::vector<UnsignedInt> children(node.children.begin(), node.children.end());
+
+    Matrix4 transformation;
+
+    if(node.rotation.size() == 4) {
```
mosra: What if `node.rotation.size() == 1`, for example? Can that happen? Or is it always either zero or the proper size?
I would put some sanity assertions here, for example: 

```cpp
CORRADE_INTERNAL_ASSERT(node.rotation.size() == 0 || node.rotation.size() == 4);
CORRADE_INTERNAL_ASSERT(node.translation.size() == 0 || node.translation.size() == 3);
CORRADE_INTERNAL_ASSERT(node.scale.size() == 0 || node.scale.size() == 3);
```

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    const tinygltf::Primitive& primitive = mesh.primitives[0];
+
+    int mode = primitive.mode;
+    if (mode == TINYGLTF_MODE_POINTS) {
+        meshPrimitive = MeshPrimitive::Points;
+    } else if (mode == TINYGLTF_MODE_LINE) {
+        meshPrimitive = MeshPrimitive::Lines;
+    }
```
mosra: What happens if `mode` is neither points, lines nor triangles? There's also triangle strip, triangle fan etc. I would personally do something like this:

```cpp
    MeshPrimitive meshPrimitive = MeshPrimitive{}; // (default-init to zero)
    if(primitive.mode == TINYGLTF_MODE_POINTS) {
        meshPrimitive = MeshPrimitive::Points;
    } else if(primitive.mode == TINYGLTF_MODE_LINE) {
        meshPrimitive = MeshPrimitive::Lines;
    } else if(primitive.mode == TINYGLTF_MODE_TRIANGLE) {
        meshPrimitive = MeshPrimitive::Triangles;
    ...
    } else {
        Error{} << "Trade::TinyGltfImporter::mesh3D(): unrecognized primitive" << primitive.mode;
        return Containers::NullOpt;
    }

File: src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp [hunk 3]
```diff
@@ -0,0 +1,484 @@
...
+
+    if (normalAccessor.type != TINYGLTF_TYPE_VEC3) {
+        Error() << "Trade::TinyGltfImporter::mesh3D(): expected type of normal is VEC3";
+        return Containers::NullOpt;
+    }
+
+    std::vector<Vector3> normalElement;
+    size_t numNormals = normalBufferview.byteLength / sizeof(Vector3);
+    normalElement.resize(numNormals);
+    std::memcpy(normalElement.data(), normalBuffer.data.data() + normalBufferview.byteOffset, normalBufferview.byteLength);
```
mosra: Why sometimes `std::memcpy()` and sometimes `std::copy()`? I would prefer `std::copy()` in all cases. Also, there's `std::copy_n()` if you'd want to avoid calling `reinterpret_cast` on both arguments.

mosra:
@NussknackerXXL wow, all done already? :O great job!

@Squareys said he'll do some image-related updates, then I'm ready to merge this :)

Squareys:
@mosra Done, finally! We'll see what coveralls says to all this, I may do another test for the image loading.

I tried to condense my cleanup commits a bit. Currently github shows the commits in a different order than I have them locally (maybe sorted chronologically here?).

mosra:
Merged, thank you a lot!

Squareys:
Thank you very much! 🎉 
And thanks for all the cleanup and feedback ;)

PR #41, opened 2018-04-29
Tinygltf importer fixes

Squareys:
Hi @mosra !

As discussed, here are some fixes to the TinyGltfImporter.

Cheers, Jonathan.

**TODO**

 - [x] UVs import
 - [x] Colors import
 - [x] Cleanup test


src/MagnumPlugins/TinyGltfImporter/Test/Test.cpp
```diff
@@ -432,6 +464,55 @@ void TinyGltfImporterTest::material() {
...
+    auto texture = importer->texture(0);
+    CORRADE_VERIFY(texture);
+    CORRADE_COMPARE(texture->image(), 0);
+    CORRADE_COMPARE(texture->type(), TextureData::Type::Texture2D);
+
+    CORRADE_COMPARE(texture->magnificationFilter(), Sampler::Filter::Linear);
+    CORRADE_COMPARE(texture->minificationFilter(), Sampler::Filter::Nearest);
+    CORRADE_COMPARE(texture->mipmapFilter(), Sampler::Mipmap::Linear);
+
+    //CORRADE_COMPARE(texture->wrapping(), Array3D<Sampler::Wrapping>(Sampler::Wrapping::Repeat, Sampler::Wrapping::Repeat, Sampler::Wrapping::Repeat));
```
mosra: Would be nice to have this tested, ideally with different values for X and Y :)

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
```diff
@@ -300,33 +300,77 @@ Containers::Optional<MeshData3D> TinyGltfImporter::doMesh3D(const UnsignedInt id
...
-    const tinygltf::Buffer& normalBuffer = _d->model.buffers[normalBufferview.buffer];
+    for (auto& attribute : primitive.attributes) {
+        const tinygltf::Accessor& accessor = _d->model.accessors[attribute.second];
+        const tinygltf::BufferView& bufferView = _d->model.bufferViews[accessor.bufferView];
+        const tinygltf::Buffer& buffer = _d->model.buffers[bufferView.buffer];
 
-    if(normalAccessor.type != TINYGLTF_TYPE_VEC3) {
-        Error() << "Trade::TinyGltfImporter::mesh3D(): expected type of normal is VEC3";
-        return Containers::NullOpt;
+        if (attribute.first.compare("POSITION") == 0) {
```
mosra: why not just `== "POSITION"`? Similarly for NORMAL below.

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
```diff
@@ -300,33 +300,77 @@ Containers::Optional<MeshData3D> TinyGltfImporter::doMesh3D(const UnsignedInt id
...
+                Error() << "Trade::TinyGltfImporter::mesh3D(): expected type of" << attribute.first << "is VEC3";
+                return Containers::NullOpt;
+            }
+
+            const size_t numNormals = bufferView.byteLength/sizeof(Vector3);
+            normals.reserve(numNormals);
+            std::copy_n(reinterpret_cast<const Vector3*>(buffer.data.data() + bufferView.byteOffset), numNormals, std::back_inserter(normals));
+
+        /* Texture coordinate attribute ends with _0, _1 ... */
+        } else if (attribute.first.compare(0, 8, "TEXCOORD") == 0) {
```
mosra: You could use `Utility::String::beginsWith()` here ;) Similarly for colors below.

mosra:
Re CI failures: 

- it seems that the `mesh-colors.bin` file (containing the actual mesh data?) is missing. 
- the new files are also not embedded for Emscripten/Android, so it fails there completely
- but it doesn't fail on openFile (as it should), rather quite later at unexpected data count. It looks like opening a nonexistent file is not handled correctly and openFile returns `true`, can you add a new test case for that (and fix it, if that's really the problem?)?
- the failures with `<?>`: it seems like tinygltf is reading past the end of the file, encountering non-ascii data. That's kinda bad :/ Maybe a patch to tinygltf is needed?

Squareys:
> (and fix it, if that's really the problem?)?

That is probably a problem in TinyGltf... I will check it, though.

> That's kinda bad :/ Maybe a patch to tinygltf is needed?

I think that's because I hand edited the file... And that are the actual contents of the file btw. (binary data is appended in .glb)

Thanks for investigating, will have time to finish this up later today.

mosra:
> That is probably a problem in TinyGltf

I don't think so, `isOpened()` checks for `!!_d` and in the current code that's `true` even if the actual file opening inside `AbstractImporter::doOpenFile()` fails. Not sure how to fix that, tho :)

> I think that's because I hand edited the file

Ah I see. Probably some size information mismatch. Use some gltf-to-glb converter then, maybe?

Squareys:
> I don't think so

You may be mixing up `.bin` with `.glb` here. The latter is the binary representation of GLTF, the former are files included in non-binary `.gltf` files which contain binary data and are opened by TinyGTLF.
TinyGltfImporter doesn't know the `.bin` extension -- and that wouldn't make sense as it is unstructured raw binary data (the metadata is contained in the `.gltf` file).

Apart from that, everything should be done, let's see what the CIs have to say.

mosra:
I tried *very hard* to find more things to complain about, but nope. I got nothing. :laughing: 

Merged. Thanks a lot! :+1:

Squareys:
Awesome, thanks! 🎉 👍

PR #45, opened 2018-07-23
Update tiny_gltf.h to upstream

Squareys:
Hi @mosra !

As discussed per gitter.

(Summary for others: the patches I applied downstream back then seem to be obsolete.)

Cheers, Jonathan.


mosra:
Looking at https://codecov.io/gh/mosra/magnum-plugins/src/b16dac21cbc2cd9c7b3cb3a94742dfc7b3e83bb1/src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp#L655, the URI image import doesn't seem to be tested at all, so can't be sure if the update breaks this code path or not.

mosra:
Thank you! Merged into `master` now :) For the record, with those modifications:

- tinygltf update as 29661ad4dd3ed8d6bc0613fbc9803040781949c8, but had to merge a few more things upstream, so the commit is different
- embedded images test as 61afb916229e6a93b73bd607ad0a453f9e7ea545, but the code was crashing on images with embedded data URIs, fixed in 34f3adbe26e01f6a1f5625d1e67133940fcf72dd
- test file cleanup in 5ee0a5fcc49281dfafe3210e70c6c836e2ef394e, but the original commit forgot to update the corresponding glb files and missed a few cases, i did an overall cleanup in the following commits

PR #46, opened 2018-07-25
[WIP] Basic animation import in glTF

mosra:
Depends on mosra/magnum#191. Things to do:

- [x] Importing separate TRS as such and not as a single transformation matrix
- [x] Importing node animation data
- [x] ~~Cubic Hermite Spline import~~ postponed after merge
- [x] Verifying this on a bunch of real-life glTF models (from experience, having just handcrafted test data was not a good idea at all)
- [x] remove the `animation` branch from CI setups after the Magnum part is merged into master

At the moment, this won't cover skinning or morph targets.


mosra:
Merged! :tada:

PR #49, opened 2018-10-23
FreeType: reset library variable during finalize()

ffreling:
By not resetting the `library` variable, we can't safely initialize twice FreeType in the lifetime of the application.


mosra:
Ugh, sorry, how did I not test this? :see_no_evil: 

This looks simple enough to still get into 2018.10, I'll write a test case to ensure this never happens again.

mosra:
Merged in a71ad5e69cf73305c0dc0a437a61235a5e4ce4fc, the corresponding test is one commit earlier. Thanks a lot! :+1:

PR #50, opened 2019-01-26
package/msys: Add PKGBUILDs.

williamjcm:
And associated `.gitattributes` file.

Once again, comments and feedback are welcome! While mosra/corrade#56 and mosra/magnum#307 allowed me to do less mistakes this time, there might still be a few issues remaining.

### DIFF
--- a/package/msys/.gitattributes
+++ b/package/msys/.gitattributes
@@ -1,0 +1,2 @@
+magnum/PKGBUILD text eof=lf
+PKGBUILD text eof=lf
--- a/package/msys/.gitattributes
+++ b/package/msys/.gitattributes
@@ -1,2 +1,2 @@
-magnum/PKGBUILD text eof=lf
+magnum-plugins/PKGBUILD text eof=lf
 PKGBUILD text eof=lf
--- a/package/msys/magnum-plugins/PKGBUILD
+++ b/package/msys/magnum-plugins/PKGBUILD
@@ -15,6 +15,7 @@ depends=("${MINGW_PACKAGE_PREFIX}-magnum>=$pkgver"
          "${MINGW_PACKAGE_PREFIX}-assimp"
          "${MINGW_PACKAGE_PREFIX}-devil"
          "${MINGW_PACKAGE_PREFIX}-freetype"
+         "${MINGW_PACKAGE_PREFIX}-harfbuzz"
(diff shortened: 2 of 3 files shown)


package/msys/.gitattributes
```diff
@@ -0,0 +1,2 @@
+magnum/PKGBUILD text eof=lf
```
mosra: `magnum-plugins/PKGBUILD` :grin:

package/msys/magnum-plugins/PKGBUILD
```diff
@@ -0,0 +1,71 @@
...
+    MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+        ${MINGW_PREFIX}/bin/cmake \
+            -G'Ninja' \
+            -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+            -DWITH_ASSIMPIMPORTER=OFF \
+            -DWITH_DDSIMPORTER=ON \
+            -DWITH_DEVILIMAGEIMPORTER=ON \
+            -DWITH_DRFLACAUDIOIMPORTER=ON \
+            -DWITH_DRWAVAUDIOIMPORTER=ON \
+            -DWITH_FREETYPEFONT=ON \
```
mosra: Oh, interesting, HarfBuzz has a MSYS package too? Then I would do `WITH_HARFBUZZFONT=ON` here as well (and the corresponding dependency above).

The reason I did this in the Arch PKGBUILD was originally to reduce the amount of dependencies, but harfbuzz is pretty ubiquitous nowadays, so I think I'll enable it on Arch as well.

package/msys/.gitattributes
williamjcm: 🤦‍♂️ 

I knew I forgot something.

package/msys/magnum-plugins/PKGBUILD
williamjcm: > HarfBuzz has a MSYS package too?

Yeah, it does. It's possible I mentioned it on Gitter months ago, back when I started with Magnum, but I could be wrong.

williamjcm:
Should be good now. 😅

mosra:
Merged in 294ff831316289339a4f6f3dcc14eb402ae0c3ed, thank you!

PR #53, opened 2019-02-23
JpegImporter: redirect libjpeg errors to Corrade::Utility::Error

xqms:
I'm crawling a large dataset with a few faulty JPEGs - which I would like to ignore silently. However, `libjpeg` errors are currently directly printed to `stderr`. This PR redirects them to `Corrage::Utility::Error`, which can then be redirected by the user to capture any errors that might occur.

### DIFF
--- a/src/MagnumPlugins/JpegImporter/JpegImporter.cpp
+++ b/src/MagnumPlugins/JpegImporter/JpegImporter.cpp
@@ -73,7 +73,9 @@ Containers::Optional<ImageData2D> JpegImporter::doImage2D(UnsignedInt) {
     } errorManager;
     file.err = jpeg_std_error(&errorManager.jpegErrorManager);
     errorManager.jpegErrorManager.error_exit = [](j_common_ptr info) {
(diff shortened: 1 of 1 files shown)


mosra:
Hi, thanks a lot! :+1: 

For some reason this was already done for the `JpegImageConverter`. Not sure why I didn't do the exact same thing here too. 

Looks like this code path is not tested either. Will merge this later today / tomorrow together with a test case.

xqms:
> For some reason this was already done for the `JpegImageConverter`. Not sure why I didn't do the exact same thing here too.

Ah ok, didn't see that. That implementation even looks slightly better with an integrated one-line error message, so maybe just take that one ;-)

mosra:
*Finally* merged in ee69e5cbee84712a04303bbe9d5a89a2c7ac0aa0. While adding the test cases for proper code coverage I discovered a handful of random bugs in all image importers (the PNG importer, especially) and, 20 commits later, these should be fixed as well.

PR #59, opened 2019-05-20
[TiledImporter] plugin and test project.

roig:
Discuss here all the things that the importer should have.


mosra:
Thank you for all the work! I think I will need to dive a bit into Tiled first to have at least a rough idea about the structure and what's all there to support. Can't promise when that will be, tho.

PR #62, opened 2019-07-31
BasisImporter

Squareys:
\> BasisImporter<| [BasisImageConverter](https://github.com/mosra/magnum-plugins/pull/65) | [AnyImageImporter/Converter](https://github.com/mosra/magnum/pull/370)
------------ | ------------- | -------------

Hi @mosra !

This is the pullrequest for the BasisImporter plugin. The BasisConverter will be in a separate pullrequest.

I will ping you once you can take the first look at it. Big warning: I currently added all of the https://github.com/binomialLLC/basis_universal code with my patch. I am still hoping https://github.com/BinomialLLC/basis_universal/pull/46 or https://github.com/BinomialLLC/basis_universal/issues/11 or maybe https://github.com/BinomialLLC/basis_universal/pull/13 will be fixed/merged/dealt with by the time this pullrequest is ready or the next release that Rich is working on solves linking to it as a library.

If not, we can rediscuss options for how to integrate it best.

Sincerely,
Jonathan

**TODOs**

 - [x] Transcode a basis file
 - [x] Target format configuration
 - [x] Magnum plugins find module
 - [x] Basis find module
    - [x] Implement
    - [x] Await tested on all platforms
 - [x] Add to all the packages
 - [x] Remove basis source code
 - [x] Basis dependency on CI
     - [x] AppVeyor
     - [x] travis
 - [x] ... and packages?
 - [x] Add to changelog and other parts of docs
 - [x] Compiled snippets
 - [x] Enable on CI
 - [x] Test
   - [x] Basic, open twice, import twice
   - [x] All aliases (instanced)
   - [x] Some more basis input file configurations
       - [x] Multiple images in a file
 - ~~[ ] AnyImageImporter delegation~~ => https://github.com/mosra/magnum/pull/370


Squareys:
@mosra So, there has been zero visible progress on build system / library support in basis_universal since June. With this PR, I just copied the files of my forked repository with the debug iterator level fix into external, but that was just a temporary solution.

How are we going to include basis? I currently don't see a way that doesn't involve our own patched version, but maybe a submodule is a better idea than just copying the files?
Otoh, we can stick to only the essential files this way and provide our own CMake setup.

mosra:
Oh well.

I would say we do it similarly as we did with ImGui (finding sources and building through and external CMake script), plus relying on your fork for Windows builds until https://github.com/BinomialLLC/basis_universal/pull/46 gets finally merged. Or is there anything else needed beyond that?

Submodules are problematic since `git submodule update --init` implicitly pulls in everything (what if the 3rd party repo suddenly starts including hundreds of megabytes of data, like assimp did? then everyone downloading plugins or having plugins as a submodule would have to transitively suffer through that too) and this would add a dangerous precedent when other dependencies get considered in the future. (Plus iterating on that 3rd party lib or trying out various forks of it is harder when you have a submodule pinned -- a problem I'm running into very often recently.)

Squareys:
That submodule concern makes a lot of sense, yes.

Finding the sources via CMake is a good idea, I remember you suggested that before 👍 But good to have it documented here on GitHub now.

> Or is there anything else needed beyond that?

For windows support, no further code changes. But to build as a library, we will have to add some cmake project. (Theirs just builds the basisu tool.)
But that can be done through the find module.

src/MagnumPlugins/BasisImporter/BasisImporter.h
```diff
@@ -0,0 +1,198 @@
...
+
+@cpp
+    /\* Choose Ect1 target format *\/
+    Containers::Pointer<AbstractImporter> importer =
+        manager.instantiate("BasisImporterEct1");
+
+    /\* Choose Bc5 target format *\/
+    Containers::Pointer<AbstractImporter> importer =
+        manager.instantiate("BasisImporterBc5");
+@ce
```
mosra: Compiled code snippets using `@snippet`, please :) That's the first time for the plugins repo, look into `doc/snippets/` of other repos (and occurences of `@snippet` elsewhere) for how that's done. Note that it's important to do this in the root CMakeLists to have them *actually* compiled:

```cmake
# Build snippets as part of testing
if(BUILD_TESTS)
    add_subdirectory(doc/snippets)
endif()
```

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,217 @@
...
+        /* Has type prefix */
+        if(plugin.length() > 13) {
+            /* We can assume the substring results in a valid value
+               as the plugin conf limits it to known suffixes */
+            configuration().setValue("format", plugin.substr(13));
+
+        /* Default configuration to Etc2 */
+        } else {
+            configuration().setValue("format",
+                BasisTranscodingType::Etc2);
```
mosra: This default ~~could~~ should be done directly inside the `BasisImporter.conf` so people can modify it by hand, when needed:

```ini
provides=BasisImporterEtc1
provides=BasisImporterEtc2
provides=BasisImporterBc1
provides=BasisImporterBc3
provides=BasisImporterBc4
provides=BasisImporterBc5
provides=BasisImporterBc7M6OpaqueOnly
provides=BasisImporterPvrtc1_4OpaqueOnly

format=Etc2
```

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -0,0 +1,188 @@
...
+    /* Explicitly checking non-null but empty view */
+    CORRADE_VERIFY(!importer->openData({&a, 0}));
+    CORRADE_COMPARE(out.str(), "Trade::BasisImporter::openData(): the file is empty\n");
+}
+
+void BasisImporterTest::invalid() {
+    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("BasisImporter");
+    std::ostringstream out;
+    Error redirectError{&out};
+    CORRADE_VERIFY(!importer->openFile(Utility::Directory::join(PNGIMPORTER_TEST_DIR, "rgb.png")));
```
mosra: Could be just `openData("hello?")`, you wouldn't need to bother with PNGIMPORTER_TEST_DIR or anything else.

src/MagnumPlugins/BasisImporter/Test/README.md
```diff
@@ -0,0 +1,65 @@
+Creating input PNG files
```
mosra: This is still a copypaste leftover from the PngImporter ... and I think the `toolchains` update as well? :)

src/MagnumPlugins/BasisImporter/BasisImporter.h
```diff
@@ -0,0 +1,188 @@
...
+
+        Magnum::Int i = 0;
+        for(const char* name : FormatNames)
+            if(name == value) return Magnum::Trade::BasisTranscodingType(i);
+
+        return Magnum::Trade::BasisTranscodingType(-1);
+    }
+};
+
+}}
```
mosra: Since this is only used from the test or plugin implementation (and never by plugin users), instead of having the `ConfigurationValue` along with everything else that it needs always present in the public header, what about putting it into a separate internal header instead?

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -0,0 +1,188 @@
...
+    CORRADE_COMPARE("Etc1", ConfigValue::toString(BasisTranscodingType::Etc1, {}));
+
+    CORRADE_COMPARE("<invalid>", ConfigValue::toString(BasisTranscodingType(-1), {}));
+    CORRADE_COMPARE("<invalid>", ConfigValue::toString(BasisTranscodingType(42), {}));
+}
+
+void BasisImporterTest::rgb() {
+    const size_t instanceId = testCaseInstanceId();
+    const std::string suffix = (instanceId == 0) ? "" :
+        Corrade::Utility::ConfigurationValue<BasisTranscodingType>::toString(BasisTranscodingType(instanceId-1), {});
```
mosra: Maybe use the canonical instance data definition instead? It'll make the code cleaner, free of the `?:` things and that way you can use the names directly, annotate the test cases etc.

```cpp
const struct {
    const char* format;
} FormatData[] {
    {"Etc2"},
    {"PvrtcBla"},
    ...
};

...

BasisImporterTest::BasisImporterTest() {
    addInstancedTests({...}, 
        Containers::arraySize(FormatData));
}

...

void BasisImporterTest::rgb() {
    auto& data = FormatData[testCaseInstanceId()];
    setTestCaseDescription(data.format);

    ...
```

src/MagnumPlugins/BasisImporter/Test/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+else()
+    configure_file(${CMAKE_CURRENT_SOURCE_DIR}/configure.h.cmake
+                   ${CMAKE_CURRENT_BINARY_DIR}/configure.h)
+endif()
+
+corrade_add_test(BasisImporterTest BasisImporterTest.cpp
+    LIBRARIES Magnum::Trade
+    FILES
+        rgb.basis
+        rgba.basis)
```
mosra: (Emscripten/Android builds will complain on their own, but the `*_2_images.basis` aren't listed here :)

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+    State(): _codebook(basist::g_global_selector_cb_size,
+        basist::g_global_selector_cb) {}
+};
+
+BasisImporter::BasisImporter() = default;
+
+BasisImporter::BasisImporter(PluginManager::AbstractManager& manager, const std::string& plugin): AbstractImporter{manager, plugin} {
+    /* Does nothing if already initialized */
+    basist::basisu_transcoder_init();
```
mosra: ~~Maybe move this line to the State constructor? Since that one does similar initialization stuff.~~

Actually, since this can be done just once, what about putting it into a static `initialize()` function instead? (See `FreeTypeFont` for an example.) Is there a corresponding `deinit()` function?

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include <Corrade/Utility/Debug.h>
+#include <Magnum/PixelFormat.h>
+#include <Magnum/Trade/ImageData.h>
+
+namespace Magnum { namespace Trade {
+
+namespace {
+
+/* Map BasisTranscodingType to Magnum::CompressedPixelFormat */
+CompressedPixelFormat getTextureFormat(BasisTranscodingType type, bool hasAlpha) {
```
mosra: No `get` prefix, please :) Similarly below.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,217 @@
...
+    Containers::Array<char> dest{Containers::NoInit, requiredSize};
+    const UnsignedInt status = _state->_transcoder.transcode_image_level(
+      _in.data(), _in.size(), index, level, dest.data(), dest.size()/bytes_per_block,
+      static_cast<basist::transcoder_texture_format>(format));
+    if(!status) {
+        Error{} << "Trade::BasisImporter::doImage2D(): transcoding failed";
+        return Containers::NullOpt;
+    }
+
+    /* Always using the default 4-byte alignment */
```
mosra: This is universally true for all compressed formats I think (no format has one- or three-byte blocks), so the comment is redundant.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+    const BasisTranscodingType targetFormat =
+        configuration().value<BasisTranscodingType>("format");
+    const basist::transcoder_texture_format format =
+        getTranscoderTextureFormat(targetFormat);
+    const UnsignedInt bytes_per_block = basis_get_bytes_per_block(format);
+
+    basist::basisu_image_info info;
+    if (!_state->_transcoder.get_image_info(_in.data(), _in.size(), info, index)) {
+        Error{} << "Trade::BasisImporter::doImage2D(): unabled to get image info";
```
mosra: unable, not unabled :) Same below.

Besides that, the prefix should have the public-facing API name to avoid confusion, so `image2D()` instead of `doImage2D()`. Similarly in other cases.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,217 @@
...
+    UnsignedInt origWidth, origHeight, totalBlocks;
+    if (!_state->_transcoder.get_image_level_desc(_in.data(), _in.size(), index,
+            level, origWidth, origHeight, totalBlocks))
+    {
+        Error{} << "Trade::BasisImporter::doImage2D(): unabled to retrieve mip level description";
+        return Containers::NullOpt;
+    }
+
+    const UnsignedInt requiredSize = totalBlocks*bytes_per_block;
+    Containers::Array<char> dest{Containers::NoInit, requiredSize};
```
mosra: Can't use `NoInit` in plugins since it causes a custom deleter to be used (which then gets dangling if the plugin is unloaded sooner than the data is deleted).

This could be theoretically fixed on Magnum side for plain types (since there `NoInit` is equivalent to `DefaultInit`) but not there yet...

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,217 @@
...
+        basist::transcoder_texture_format::cTFBC1,
+        basist::transcoder_texture_format::cTFBC3,
+        basist::transcoder_texture_format::cTFBC4,
+        basist::transcoder_texture_format::cTFBC5,
+        basist::transcoder_texture_format::cTFBC7_M6_OPAQUE_ONLY,
+        basist::transcoder_texture_format::cTFPVRTC1_4_OPAQUE_ONLY
+    };
+
+    return formats[Int(type)];
+}
```
mosra: What about doing just `TranscoderTextureFormat[Int(type)]` (and having the array file-local) instead of calling the function?

src/MagnumPlugins/BasisImporter/BasisImporter.h
```diff
@@ -0,0 +1,188 @@
...
+in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNETCION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+*/
```
mosra: Indentation is off here (it would make it harder for me to do batch copyright updates, then). The comment below seems to be as well.

Actually, the class too -- even though it's excessive, for consistency also the `public:` markers should be indented (so the contents twice).

doc/building-plugins.dox
```diff
@@ -1,6 +1,5 @@
 /*
     This file is part of Magnum.
-
```
mosra: Accident? :)

src/MagnumPlugins/BasisImporter/BasisImporter.h
Squareys: Oh! I was so certain that public would not be indented, that's kind of counter-intuitive to the [overarching goal](https://doc.magnum.graphics/corrade/corrade-coding-style.html#corrade-coding-style-cpp-style)

> The goal is to preserve readability with minimal amount of wasted vertical and horizontal space.

No idea how the license indentation happened 😅

src/MagnumPlugins/BasisImporter/BasisImporter.h
Squareys: `Configuration.h`? That would need to include BasisImporter.h for the enum, I suppose. Or is there a way around that?

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
Squareys: Yes, definitely.

src/MagnumPlugins/BasisImporter/BasisImporter.h
mosra: Yes, it would, but that's okay -- whoever needs the `ConfigurationValue` will need the importer as well to use the values, but not the other way around.

src/MagnumPlugins/BasisImporter/BasisImporter.h
mosra: I know, it's excessive, but it's done like this everywhere else and I think consistency is what matters. Maybe one day I'll go over all classes and change this to something less excessive, but not yet.

src/MagnumPlugins/BasisImporter/BasisImporter.h
Squareys: It's no longer required in the tests, so I will move it into BasisImporter.cpp

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
Squareys: There is no deinit function and successive calls to init() [don't do anything](https://github.com/BinomialLLC/basis_universal/blob/master/transcoder/basisu_transcoder.cpp#L1660)

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
Squareys: Ah, of course 😅

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
mosra: Okay, thank you -- can you then move it to the State constructor to be together with similar setup calls?

*(Oh god, a `static` function-local variable :see_no_evil: -- I need to double check the supposed thread safety of this thing.)*

mosra: Checking https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables, there doesn't seem to be any special case for plain data types, so this should work in a multi-threaded setting. However ... if one thread is in the middle of the `transcoder_init()` (so `s_initialized` not `true` yet) and another thread calls it as well, it will result in a data race. Which leads me to think this should really be in `initialize()` instead, as that's called just once during plugin load and not in every plugin constructor (which might be called simultaenously from multiple threads).

Correct me if I'm wrong of course :)

Squareys:
@mosra How should I handle the basis dependency on debian/arch/etc packages? Clone the repo or download from specific revision like on travis and appveyor?

mosra:
Do it the same way as ImGui -- `WITH_BASISIMPORTER=OFF` for all those. I'll make Arch package myself post-merge for testing. For other platforms ... bad luck I guess, and I doubt any packaging effort will happen before they sort out their buildsystem issues.

Arch has a package ([basis-universal-git](https://aur.archlinux.org/packages/basis-universal-git/)), but guess what, it's just the executable again.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
Squareys: Sounds good to me, moved it to `initialize()` 👍

modules/FindBasisUniversal.cmake
```diff
@@ -0,0 +1,176 @@
...
+
+        find_path(BasisUniversalTranscoder_DIR NAMES basisu_transcoder.cpp
+            HINTS "${BASISU_DIR}/transcoder" "${BASISU_DIR}"
+            NO_CMAKE_FIND_ROOT_PATH)
+        add_library(BasisUniversalTranscoder STATIC
+            ${BasisUniversalTranscoder_DIR}/basisu_transcoder.cpp)
+        target_include_directories(BasisUniversalTranscoder PUBLIC
+            ${BasisUniversalTranscoder_INCLUDE_DIR})
+
+        if(CMAKE_CXX_COMPILER_ID MATCHES "(Apple)?(Clang)|GCC") # FIXME: What's the right pattern here?
```
Squareys: This still remains

mosra: Same what UseCorrade.cmake does I think: https://github.com/mosra/corrade/blob/35a7b2bc29838eead10ed2d0382374402ed1a327/modules/UseCorrade.cmake#L72

Squareys: Wow, what a line, would have never figured that out correctly 😂

src/MagnumPlugins/BasisImporter/BasisImporter.h
```diff
@@ -0,0 +1,160 @@
...
+The list of valid suffixes is equivalent to enum value names in @ref BasisTranscodingType.
+
+If you want to be able to change the target format dynamically, you may want to
+set the "format" configuration of the plugin:
+
+@snippet MagnumPluginsBasisImporter.cpp basis-target-format-config
+
+If you link to the plugin you may also use @ref BasisImporter::setTargetFormat().
+
+*/
```
mosra: The documentation could be improved:

- Explicitly saying what this is for -- parsing a basis file and giving one of the GPU-specific *compressed* formats on the output. And because there isn't any "one size fits all", one first *has to* explicitly check for support first, and only then then choose a particular format to transcode.
- Referencing (`@ref`) the concrete `CompressedPixelFormat` values for each `BasisTranscodingType`, because those then link to concrete values for GL and Vulkan together with extension information. Without that, the users would have to guess what format will the result be in and what extensions to check for.
- Mentioning the license info like all other plugins have, to be listed at https://doc.magnum.graphics/magnum/credits-third-party.html

src/MagnumPlugins/BasisImporter/Test/README.md
```diff
@@ -0,0 +1,16 @@
+Creating input Basis files
+==========================
+
+The basis test files were converted from the PngImporter's test files
+`rgb.png`, `rgba.png` (and 90° clockwise rotated version for the second images
+for the `rgb(a)_2_images.basis` files) using the official basis universal
+[conversion tool](https://github.com/BinomialLLC/basis_universal/#command-line-compression-tool).
```
mosra: Could you craft larger files (64x32 at least, for example)? Those two are 3x2 and probably encode just into random noise, as GPU compression blocks are at least 4x4 pixels. Not sure if block-aligning is a good idea, so maybe 63x27 or some such :)

There's this now: https://github.com/mosra/magnum/tree/master/src/Magnum/Shaders/Test/TestFiles feel free to use/resize/crop any of those as you see fit (some are quite dark tho, so check first). The `magnum-imageconverter` can convert them to PNGs, or create a new one from the SVG.

src/MagnumPlugins/BasisImporter/BasisImporter.conf
```diff
@@ -0,0 +1,13 @@
...
+provides=BasisImporterBc1
+provides=BasisImporterBc3
+provides=BasisImporterBc4
+provides=BasisImporterBc5
+provides=BasisImporterBc7M6OpaqueOnly
+provides=BasisImporterPvrtc1_4OpaqueOnly
+
+# [configuration]
+[configuration]
+format=Etc2
```
Squareys: I assume this should be `Format` analog to Assimp configuration values? :thinking:

mosra: no, those are uppercase only because the Assimp internal enums are named that way too (and so I have the mapping easier)

src/MagnumPlugins/BasisImporter/Test/README.md
Squareys: Replace all of the existing files with larger files?

src/MagnumPlugins/BasisImporter/BasisImporter.h
Squareys: I am unsure whether I fulfilled point 1 sufficiently, but the improvement on `BasisTranscodingType` is super valuable! And point 3 is done also ✔️

mosra: Just saw that, wonderful. And it something turns out to be missing, we can always improve later :)

mosra:
For the no-default thing I mentioned in mosra/magnum#370, I think it would be enough to have just an empty format line in the config, like:

```ini
[configuration]

# No format is specified by default and you have to choose one either
# by changing this value or by loading the plugin under an alias. See 
# class documentation for more information.
format=
```

And then this should have an explicit test so we are sure this fails gracefully with a good UX :)

Squareys:
Actually, that fails with a `Numerical` exception

Squareys:
Alright, done! 🎉

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -0,0 +1,246 @@
...
+    /* Explicitly forbid system-wide plugin dependencies */
+    PluginManager::Manager<AbstractImporter> _manager{"nonexistent"};
+};
+
+constexpr struct {
+    const char* suffix;
+    const char* expectedFormatStr;
+    const CompressedPixelFormat expectedFormat;
+    const CompressedPixelFormat expectedFormatAlpha;
+    const bool needsPow2;
```
mosra: This again leads to `?:` in the test, allowing just two hardcoded variants. Would be better as 

```cpp
const char* file;
Vector2i expectedSize;
```

as it allows more parametrization in the future ;)

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -0,0 +1,246 @@
...
+    void rgb();
+    void rgba();
+
+    /* Explicitly forbid system-wide plugin dependencies */
+    PluginManager::Manager<AbstractImporter> _manager{"nonexistent"};
+};
+
+constexpr struct {
+    const char* suffix;
+    const char* expectedFormatStr;
```
mosra: Since those two are always the same, why not have just one of them?

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -0,0 +1,246 @@
...
+void BasisImporterTest::unconfigured() {
+    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("BasisImporter");
+    std::ostringstream out;
+    Error redirectError{&out};
+
+    CORRADE_VERIFY(importer->openFile(
+        Utility::Directory::join(BASISIMPORTER_TEST_DIR, "rgb.basis")));
+    CORRADE_VERIFY(!importer->image2D(0));
+
+    CORRADE_COMPARE(out.str(), "Trade::BasisImporter::image2D(): invalid configuration: format is expected to be one of: Etc1, Etc2, Bc1, Bc3, Bc4, Bc5, Bc7M6OpaqueOnly, Pvrtc1_4OpaqueOnly\n");
```
mosra: This is exactly what I meant by "have an explicit test so we are sure this fails gracefully with a good UX" ;) This error would be valid if you set a format to some bogus value, but then it should say *what* the wrong format is -- so if you set `format=Bananas`, it complains with

> Trade::BasisImporter::image2D(): invalid configured transcoding format Bananas, expected to be one of [...]

However, when the format is empty (which is the default case), from the user POV the above doesn't really say *what to do*. In that case, something like this would be better:

> Trade::BasisImporter::image2D(): no format to transcode to was specified. Either load the plugin via one of its BasisImporterEtc1, ... aliases, or set the format explicitly via plugin configuration.

This also means there should be two tests for this, one `unconfigured()` and one `wrongConfiguredFormat()`.

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -0,0 +1,246 @@
...
+    /* Load the plugin directly from the build tree. Otherwise it's static and
+       already loaded. */
+    #ifdef BASISIMPORTER_PLUGIN_FILENAME
+    CORRADE_INTERNAL_ASSERT(_manager.load(BASISIMPORTER_PLUGIN_FILENAME) & PluginManager::LoadState::Loaded);
+    #endif
+}
+
+void BasisImporterTest::empty() {
+    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("BasisImporter");
+    importer->configuration().setValue("format", "Etc2");
```
mosra: Is this needed here? I think not.

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -0,0 +1,246 @@
...
+        Containers::Optional<Trade::ImageData2D> image = importer->image2D(0);
+        CORRADE_VERIFY(image);
+        CORRADE_COMPARE(image->size(), (Vector2i{63, 27}));
+    }
+}
+
+void BasisImporterTest::rgb() {
+    auto& formatData = FormatData[testCaseInstanceId()];
+    const std::string pluginName = "BasisImporter" + std::string(formatData.suffix);
+    setTestCaseDescription(pluginName);
```
mosra: the description could be just `formatData.suffix`

src/MagnumPlugins/BasisImporter/Test/README.md
```diff
@@ -0,0 +1,29 @@
+Creating input Basis files
+==========================
+
+The basis test files were converted from the PngImporter's test files
```
mosra: This is not true anymore. A leftover?

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -0,0 +1,246 @@
...
+        CORRADE_VERIFY(image);
+        CORRADE_COMPARE(image->size(), (Vector2i{63, 27}));
+    } {
+        importer->configuration().setValue("format", "Bc1");
+
+        Containers::Optional<Trade::ImageData2D> image = importer->image2D(0);
+        CORRADE_VERIFY(image);
+        CORRADE_COMPARE(image->size(), (Vector2i{63, 27}));
+    }
+}
```
mosra: I would expand this one to test also the imported format like the others do, call it `importMultipleFormats()` and move it below the rgb/rgba ones. Because how it is now it feels like a variant of the trivial `importTwice()` tests from elsewhere that's testing the target format switch only by accident.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    if(!_state->_transcoder.validate_header(data.data(), data.size())) {
+        Error() << "Trade::BasisImporter::openData(): invalid header";
+        return;
+    }
+
+    if(!_state->_transcoder.start_transcoding(data.data(), data.size())) {
+        Error() << "Trade::BasisImporter::openData(): bad basis file";
+        return;
+    }
```
mosra: I think you could craft a `fileTooShort()` test for this one by passing just a prefix of the data to `openData()`.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,251 @@
...
+};
+
+}
+
+struct BasisImporter::State {
+    /* There is only this type of codebook */
+    basist::etc1_global_selector_codebook _codebook;
+    basist::basisu_transcoder _transcoder{&_codebook};
+
+    State(): _codebook(basist::g_global_selector_cb_size,
```
mosra: `explicit`

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,251 @@
...
+        case BasisTranscodingType::Bc4:
+            return CompressedPixelFormat::Bc4RUnorm;
+        case BasisTranscodingType::Bc5:
+            return CompressedPixelFormat::Bc5RGUnorm;
+        case BasisTranscodingType::Bc7M6OpaqueOnly:
+            return CompressedPixelFormat::Bc7RGBAUnorm;
+        case BasisTranscodingType::Pvrtc1_4OpaqueOnly:
+            return CompressedPixelFormat::PvrtcRGB4bppUnorm;
+        default:
+            CORRADE_ASSERT_UNREACHABLE();
```
mosra: `CORRADE_ASSERT_UNREACHABLE(); /* LCOV_EXCL_LINE */` will make it squeeze the last bits of code coverage :)

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,251 @@
...
+        }
+        #undef _c
+
+        return "<invalid>";
+    }
+
+    static Magnum::Trade::BasisTranscodingType fromString(
+        const std::string& value, ConfigurationValueFlags)
+    {
+        const Containers::StaticArray<8, const char*> FormatNames{
```
mosra: I don't think this needs a StaticArray. Range-based for will work with `const char* FormatNames[]` as well (remove the include, too). 

Actually I think it would be good to have it outside of the function and next to the `TranscoderTextureFormat`, because those two need to have the same order.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    const UnsignedInt requiredSize = totalBlocks*bytes_per_block;
+    Containers::Array<char> dest{Containers::DefaultInit, requiredSize};
+    const UnsignedInt status = _state->_transcoder.transcode_image_level(
+      _in.data(), _in.size(), index, level, dest.data(), dest.size()/bytes_per_block,
+      static_cast<basist::transcoder_texture_format>(format));
+    if(!status) {
+        Error{} << "Trade::BasisImporter::image2D(): transcoding failed";
+        return Containers::NullOpt;
+    }
```
mosra: Is there an easy way to trigger either of those with a hand-patched file? For JPEG/PNG tests setting image size to zero in the file header usually worked, so maybe a similar thing here? Or what else does the transcoder blow up on?

package/ci/appveyor-desktop.bat
```diff
@@ -69,6 +69,7 @@ cmake .. ^
     -DCMAKE_INSTALL_PREFIX=%APPVEYOR_BUILD_FOLDER%/deps ^
     -DCMAKE_PREFIX_PATH=%APPVEYOR_BUILD_FOLDER%/openal;%APPVEYOR_BUILD_FOLDER%/devil;C:/Tools/vcpkg/installed/x64-windows ^
     -DWITH_ASSIMPIMPORTER=OFF ^
+    -DWITH_BASISIMPORTER=OFF -DBASISU_DIR=%APPVEYOR_BUILD_FOLDER%/basis_universal ^
```
mosra: Adding a note here so we don't forget to flip this on.

package/ci/appveyor-desktop-mingw.bat
```diff
@@ -67,6 +67,7 @@ cmake .. ^
     -DCMAKE_INSTALL_PREFIX=%APPVEYOR_BUILD_FOLDER%/deps ^
     -DCMAKE_PREFIX_PATH=%APPVEYOR_BUILD_FOLDER%/openal;%APPVEYOR_BUILD_FOLDER%/devil ^
     -DWITH_ASSIMPIMPORTER=OFF ^
+    -DWITH_BASISIMPORTER=OFF -DBASISU_DIR=%APPVEYOR_BUILD_FOLDER%/basis_universal ^
```
mosra: Would be good to mention the `OFF` above in the comment (due to mingw not compiling), otherwise we forget why is it disabled.

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
Squareys: Ah, remnant from `""` prefix. 👍

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
Squareys: This will create a couple more `namespace ... {` sections. I guess that's alright?

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
Squareys: E.g. trying to transcode non pow2 image to Pvrtc

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -190,55 +223,51 @@ void BasisImporterTest::importTwice() {
...
-    else CORRADE_COMPARE(image->size(), (Vector2i{63, 27}));
+    CORRADE_COMPARE(image->size(), formatData.expectedSize);
 
     /* Verify that the 90° rotated second image can be loaded also */
     image = importer->image2D(1);
     CORRADE_VERIFY(image);
     CORRADE_COMPARE(image->compressedFormat(), formatData.expectedFormatAlpha);
-    if(formatData.needsPow2) CORRADE_COMPARE(image->size(), (Vector2i{32, 64}));
-    else CORRADE_COMPARE(image->size(), (Vector2i{27, 63}));
+    CORRADE_COMPARE(image->size(), (Vector2i{formatData.expectedSize.y(), formatData.expectedSize.x()}));
```
mosra: `.flipped()` ;)

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
mosra: Yeah that's fine. The way you just did that is fine as well.

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
Squareys: Even checked the docs because I guessed there would be something like this :D

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
mosra: Right! :) Does it provide a human-readable error message in that case that could be passed to Error? Or does it print it on its own?

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
mosra: There *always* is :P

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
Squareys: This will fail validate header first independent of how long i leave the prefix. But I verifiy the `invalid header` error now.

package/ci/appveyor-desktop.bat
Squareys: Oh no, I thought I did 😅

Squareys: Ah, yes, but the commit was later

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
mosra: Ah because there's the CRC check. In what cases does `start_transcoding()`` actually return false?

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
Squareys: Ergo, rename and move `importTwice` and add format check, correct?

mosra: Yes.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
Squareys: Well, the header check [checks the file size ](https://github.com/BinomialLLC/basis_universal/blob/master/transcoder/basisu_transcoder.cpp#L4226)

I assume one could pass random data after the signature 🤔

Squareys:
@mosra Ready! 👍

doc/snippets/MagnumPluginsBasisImporter.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include <Magnum/Trade/AbstractImporter.h>
+
+#ifdef __has_include
+#if __has_include(<MagnumPlugins/BasisImporter/importStaticPlugin.cpp>)
+/* [static-import] */
+/* No need to do this if you use CMake */
+#include <MagnumPlugins/BasisImporter/importStaticPlugin.cpp>
+/* [static-import] */
+#endif
+#endif
```
mosra: No matter how hard I try, I can't find this snippet being used anywhere. Since it's the same for all plugins I don't think it makes sense to list it for every plugin, so ... remove? :)

package/ci/appveyor-desktop-mingw.bat
```diff
@@ -60,13 +60,16 @@ cd .. && cd ..
 
 rem Build. BUILD_GL_TESTS is enabled just to be sure, it should not be needed
 rem by any plugin.
+rem WITH_BASISIMPORTER is disabled, since Basis Universal currently does not
+rem support building on MinGW.
```
mosra: I just realized why https://ci.appveyor.com/project/mosra/magnum-plugins/builds/26945769/job/wk4pvpxw8d5s1y03#L1117 didn't build but https://ci.appveyor.com/project/mosra/magnum-plugins/builds/27017597/job/fdb7jx0qxspnox4p did -- originally you had the whole thing embedded and now you're building just the transcoder part. So that means the `Importer` could be enabled on MinGW, but the `ImageConverter` in #65 can't be. Enable this again then and move this comment to #65 then? :)

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -0,0 +1,315 @@
...
+    Error redirectError{&out};
+
+    CORRADE_VERIFY(!importer->openData(basisData.prefix(64)));
+
+    /* Corrupt the header */
+    basisData[100] = 100;
+    CORRADE_VERIFY(!importer->openData(basisData));
+
+    CORRADE_COMPARE(out.str(), "Trade::BasisImporter::openData(): invalid header\n"
+        "Trade::BasisImporter::openData(): bad basis file\n");
```
mosra: :heart: this is amazing, thank you!

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -0,0 +1,259 @@
...
+        return {};
+    }
+    const basist::transcoder_texture_format format =
+        TranscoderTextureFormat[Int(targetFormat)];
+    const UnsignedInt bytes_per_block = basis_get_bytes_per_block(format);
+
+    basist::basisu_image_info info;
+    if (!_state->_transcoder.get_image_info(_in.data(), _in.size(), info, index)) {
+        Error{} << "Trade::BasisImporter::image2D(): unable to get image info";
+    }
```
mosra: Missing `return {};`? :scream: 

It's like the Murphy's law -- the only code path we don't test because we think "it's good enough" will have a bug in it :D Looking at the source of `get_image_info()` I have no idea how to trigger this one (most of the sanity checks done there are checked already by `AbstractImporter` base implementation itself), so probably just add the `return` and pray :)

src/MagnumPlugins/BasisImporter/BasisImporter.h
```diff
@@ -0,0 +1,221 @@
...
+        */
+        void setTargetFormat(BasisTranscodingType format);
+
+        /**
+        * @brief The importers target format
+        * @param format The format to transcode Basis files to
+        *
+        * See @ref basisimporter-target-format.
+        */
+        BasisTranscodingType targetFormat();
```
mosra: Could be `const`.

package/ci/appveyor-desktop-mingw.bat
Squareys: I think I had enabled this in the third or second to last build though to make sure ... And it didn't build...

mosra: But for a different reason ;) https://ci.appveyor.com/project/mosra/magnum-plugins/builds/27048449/job/ayx4ravm2almu2s8 ... because the YML doesn't download the library for MinGW.

Squareys: Oh! I thought I disabled that for that build 🙈 good catch, should be done around 4pm today at the latest

Squareys:
Done once more :)
(Waiting for the CI still, though)

mosra:
Merged in 627beffeb585599629540cd1d006e8280772d5e1, 0c238b80e69b9efb33e2b47279b653002fb72d3e, cd9d838b8bd4a9e099b4c88241826e4e5d19728b and 3c0041272f4fa158a82e2efc232783109078a21c; updated to ASTC and uncompressed formats in fcb1439da1fdbc537b746e66c402cc0199b6740c. Docs are [up now as well](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1BasisImporter.html).

There's still stuff to be done on Basis side, see #70.

Thanks for working on this!

PR #64, opened 2019-08-22
AssimpImporter: fix material loading

ikalevatykh:
Fix material loading when the material has ambient texture


mosra:
Wow, thank you! 

Somehow I missed the *main* case where you would actually want the ambient color to be white. Will merge tomorrow.

mosra:
Merged in b1102cb38e7ce86ae52896411acc9a564dadaab3 and added a regression test case in a23e6f0e28737613726521a96ccd09c6d6ed1421 (took a while, because Assimp is ... special). Thanks for the fix!

PR #65, opened 2019-08-29
BasisImageConverter

Squareys:
[BasisImporter](https://github.com/mosra/magnum-plugins/pull/62) | >BasisImageConverter< | [AnyImageImporter/Converter](https://github.com/mosra/magnum/pull/370)
------------ | ------------- | -------------

Hey @mosra !

Here's step 2 of Basis support, based on the `basis-importer` branch, so will need to rebase once #62 is merged :)

**TODOs:**

- [x] Fix SEGFAULT on fileData return of `doExportData`
- [x] Test
    - [x] RGB
    - [x] RGBA
    - [x] Invalid/Failure
- [x] Additional config options?
- ~~[ ] API for config options?~~
- [x] Enable on CI
    - [x] Travis
    - [x] AppVeyor
    - [x] CI Green
- [x] Disable on all packages
- [x] Documentation
    - [x] License
    - [x] Configuration
    - [x] Add everywhere


Squareys:
Rebased onto `next` :)

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,216 @@
...
+        case basisu::basis_compressor::error_code::cECFailedFontendExtract:
+            /* This error will actually never be raised from basis_universal code */
+            CORRADE_ASSERT_UNREACHABLE();
+        case basisu::basis_compressor::error_code::cECFailedWritingOutput:
+            /* We do not write any files, just data */
+            CORRADE_ASSERT_UNREACHABLE();
+        default:
+            Error{} << "Trade::BasisImageConverter::exportToData(): export failed with unknown error code.";
+            return {};
+        /* LCOV_EXCL_STOP */
```
mosra:
```cpp
        /* LCOV_EXCL_START */
        case basisu::basis_compressor::error_code::cECFailedFontendExtract:
            /* This error will actually never be raised from basis_universal code */
        case basisu::basis_compressor::error_code::cECFailedWritingOutput:
            /* We do not write any files, just data */
        default:
            CORRADE_ASSERT_UNREACHABLE();
        /* LCOV_EXCL_STOP */
```

If this blows up, we can add the new error codes to the list.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    basisu::basis_compressor_params params;
+
+    const bool multithreading = configuration().value<bool>("multithreading");
+    params.m_multithreading = multithreading;
+    basisu::job_pool jpool(threadCount);
+    params.m_pJob_pool = &jpool;
+
+#define PARAM_CONFIG(name, type) params.m_##name = configuration().value<type>(#name)
```
mosra: Indent the preprocessor, please :)

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
```diff
@@ -0,0 +1,111 @@
...
+namespace Magnum { namespace Trade {
+
+/**
+@brief BASIS image converter plugin
+
+Creates Portable Network Graphics (`*.png`) files from images with format
+@ref PixelFormat::R8Unorm / @ref PixelFormat::R16Unorm,
+@ref PixelFormat::RG8Unorm / @ref PixelFormat::RG16Unorm,
+@ref PixelFormat::RGB8Unorm / @ref PixelFormat::RGB16Unorm or
+@ref PixelFormat::RGBA8Unorm / @ref PixelFormat::RGBA16Unorm.
```
mosra: I spent quite some time looking into how Basis handles 16-bit depth.... just to notice this is just a copypaste error.

`libBASIS` ... what, did you read this at least once? :joy:

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
```diff
@@ -0,0 +1,111 @@
...
+@snippet MagnumPlugins/BasisImporter/BasisImporter.conf configuration
+
+@subsection Trade-BasisImporter-magnum-configuration Magnum-specific configuration
+
+The following options can be configured in addition to the ones that are
+exposed `basisu` parameters.
+
+`num_threads` - configure the number of threads basis should use during compression
+
+`enable_debug_printf` - enable verbose logging in basis
```
mosra: Ummmmm ..... a list, maybe? :)

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,216 @@
...
+    if(image.format() != PixelFormat::RGB8Unorm
+        && image.format() != PixelFormat::RGBA8Unorm) {
+        Error() << "Trade::BasisImageConverter::exportToData(): unsupported pixel format" << image.format();
+        return {};
+    }
+
+    /* Magnum specific configuration */
+    const Int threadCount = configuration().value<Int>("num_threads");
+    if(threadCount < 1) {
+        Error() << "Trade::BasisImageConverter::exportToData(): invalid configuration: num_threads is expected to be at least 1, but was:" << threadCount;
```
mosra:
```suggestion
        Error() << "Trade::BasisImageConverter::exportToData(): invalid configuration: num_threads is expected to be at least 1, but was" << threadCount;
```

(No colon at the end.) Also, if `0`, use [std::thread::hardware_concurrency](http://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency)? That's the usual thing elsewhere.

Also, `threads` instead of `num_threads`? Since that's our parameter name anyway?

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,216 @@
...
+        return {};
+    }
+
+    basisu::basis_compressor_params params;
+
+    const bool multithreading = configuration().value<bool>("multithreading");
+    params.m_multithreading = multithreading;
```
mosra: What happens if 

```ini
multithreading=false
num_threads=8
```

? Shouldn't `multithreading` be implicit based on the number of threads specified (i.e., enabled if `threads != 1`)?

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
```diff
@@ -0,0 +1,177 @@
...
+    importer->configuration().setValue("format", "Etc2");
+    CORRADE_VERIFY(importer->openData(data));
+    Containers::Optional<Trade::ImageData2D> converted = importer->image2D(0);
+    CORRADE_VERIFY(converted);
+
+    CORRADE_COMPARE(converted->size(), Vector2i(2, 3));
+    /* Ensure there is an alpha channel */
+    CORRADE_COMPARE(converted->compressedFormat(), CompressedPixelFormat::Etc2RGBA8Unorm);
+}
+
```
mosra: Is there any chance to trigger the encoder failures? Otherwise the error code enum is never covered.

package/archlinux/PKGBUILD-release
```diff
@@ -25,6 +25,7 @@ build() {
         -DBUILD_GL_TESTS=ON \
         -DWITH_ASSIMPIMPORTER=ON \
         -DWITH_BASISIMPORTER=ON \
+        -DWITH_BASISIMAGECONVERTER=OFF \
```
mosra: Where Importer is `ON`, converter could be `ON` also :) I guess you ran into conflicts with all those?

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,216 @@
...
+    PARAM_CONFIG(mip_filter, std::string);
+
+    PARAM_CONFIG(max_endpoint_clusters, int);
+    PARAM_CONFIG(max_selector_clusters, int);
+    PARAM_CONFIG(quality_level, int);
+#undef PARAM_CONFIG
+
+    /* Without this, basis tries to find some global selector cookbook
+       and crashes on an assert that it does not exit */
+    params.m_no_hybrid_sel_cb = true;
```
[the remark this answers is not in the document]
mosra: Related: https://github.com/BinomialLLC/basis_universal/issues/72

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,216 @@
...
+        image.dataProperties();
+    Containers::ArrayView<const char> inputData = image.data().suffix(
+        dataProperties.first.sum());
+
+    basisu::image& basisImage = params.m_source_images.back();
+
+    /* Do Y-flip and tight packing of image data */
+    const std::size_t rowSize = image.size().x()*image.pixelSize();
+    const std::size_t rowStride = dataProperties.second.x();
+    const std::size_t packedDataSize = rowSize*image.size().y();
```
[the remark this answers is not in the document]
Squareys: We discussed this on gitter quite a bit, but the additional flipping of the input data adds another nuance of course. I expect it to be like this:

 - If the Y flipped flag is set on the basis image, it should display correctly in OpenGL without need to change UVs.
 - If the Y flipped flag is not set, we need to flip.
 - Since basis expects the data to be "unflipped", but I assume that Magnum Images are flipped in general, because of OpenGL, we need to flip the magnum image data. Simply inverting the converter parameter is insufficient, because that would set the incorrect flag in the resulting basis image.

If my expectations are correct, the comment should stay and instead I should be flipping the image. @mosra What about point 2, is that correct? I'll check the others.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    /* basis image is always RGBA, fill in alpha if necessary */
+    if (image.format() == PixelFormat::RGB8Unorm) {
+        const std::size_t width = image.size().x();
+        auto basisView = Containers::arrayCast<Color4ub>(
+            Containers::arrayView<basisu::color_rgba>(basisImage.get_ptr(), width));
+
+        for(Int y = 0; y < width; ++y) {
```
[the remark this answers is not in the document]
Squareys: hmmmmm 🤔 😅

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,216 @@
...
+            const auto src = Containers::arrayCast<const Color3ub>(inputData
+                .suffix(y*rowStride).prefix(rowSize));
+
+            Color4ub* destPixel = basisView.suffix(width).begin();
+            for(const Color3ub& pixel: src) {
+                destPixel->rgb() = pixel;
+                destPixel->a() = 255;
+            }
```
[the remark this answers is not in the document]
Squareys: I need better tests 🙈

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
Squareys: Would `0` be a better default then, maybe?

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
Squareys: Would make sense, yeah

Squareys: In that case `threads=1` is a good default.

package/archlinux/PKGBUILD-release
Squareys: You mean during rebase? Yes 😅

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+    } else {
```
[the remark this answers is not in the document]
Squareys: > as a sanity check in case the format support gets extended for HDR formats and such.

Uhm, but I have a check for the format at the very top? Is that insufficient?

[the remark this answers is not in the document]
Squareys: Re the 1/2-component images: checking basis code etc, I think this can be achieved by converting the input to RRR(G) and compressing that.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
Squareys: Oh no, I must have forgotten about this entirely documentation Block entirely, it even sill says that it creates PNG files

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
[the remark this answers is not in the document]
mosra: Hah, you're right about the third point :see_no_evil: I assumed the Y flip was done in their frontend (`basisu_tool.cpp`) and the encoder just consumed the flag. That would be a way to do this properly, I think, since it would allow pre-flipped images to be put there. (Another PR flying their way, then.)

> instead I should be flipping the image

No, I think a better way would be to "post-patch" the file and add this flag to their header. Should be quite robust as long as we test properly.

For the second point -- for Vulkan, which has again Y at top AFAIK, I'll need to equip all importers with a configurable flag for Y flipping. So the final condition would be rather `if(fileYflipped != apiNeedsImageFlip) flip()`. But until then, yes, what you say is correct.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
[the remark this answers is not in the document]
mosra: > Uhm, but I have a check for the format at the very top? Is that insufficient?

Those two checks are (relatively) far from each other, so if you extend the one at the top but forget this one at the bottom, it'll silently produce wrong results (instead of blowing up). That's what I meant.

> I think this can be achieved by converting the input to RRR(G) and compressing that.

Hmm, that sounds like I should extend the importer tests with those as well.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -50,26 +52,31 @@ auto BasisImageConverter::doFeatures() const -> Features { return Feature::Conve
...
-    const Int threadCount = configuration().value<Int>("num_threads");
-    if(threadCount < 1) {
-        Error() << "Trade::BasisImageConverter::exportToData(): invalid configuration: num_threads is expected to be at least 1, but was:" << threadCount;
+    Int threadCount = configuration().value<Int>("threads");
+    if(threadCount == 0) {
+        threadCount = std::thread::hardware_concurrency();
+    } else if(threadCount < 1) {
+        Error() << "Trade::BasisImageConverter::exportToData(): invalid configuration: threads is expected to be at least 0, but was"
+                << threadCount;
         return {};
```
mosra: You could fetch the `threads` as `UnsignedInt` and don't do any range checking at all, then people who accidentally say `-1` would spin up 4 billion threads and fry their machines, which is not our problem. Same would happen if people enter `threads=400`.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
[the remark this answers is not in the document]
Squareys: > No, I think a better way would be to "post-patch" the file and add this flag to their header.

This will result in a bit of convoluted logic, because as the user I would try to produce the "y-flipped" flag by setting `y-flipped`.
Hence, if the user sets y-flipped, we need to pass the image to the basis converter as is and then set the y-flip flag. (Without actually passing `y-flipped` parameter to basis, to perserve the data as-is!)
But when the user requests non-y-flip flag, we need to enable y-flipped on the converter, so that it "unflips" the image, we would still be inverting the flag afterwards.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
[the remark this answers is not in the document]
Squareys: Ah! Makes a lot of sense now, thanks for explaining, will add the assert 👍

package/archlinux/PKGBUILD-release
Squareys: I enabled the converter on archlinux/* PKGBUILD files except emscripten and mingw.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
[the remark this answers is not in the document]
mosra: Yeah, it would be basically passing the inverse of the Y-flip flag to the converter and then inverting the flag again on the resulting file.

.... until there's globally controllable Y-flip or the client-side flip I did for the transcoder it'll be kinda shitty like this, yeah.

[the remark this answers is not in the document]
Squareys: Actually, we have to copy every image row by row anyway, for RGB we need to re-pack to RGBA, RG => RRRG, R => RRR1, RGBA also needs to be copied because of skip and all.
Flipping of the input image is implemented at the current state of the PR.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
Squareys: It seems it's not great user-experience to let them figure out that "-1" was casted to unsigned int, probably because they accidentally put a `-` there?

mosra: i don't think it's our responsibility to check for that, same happens with `new int[-1];`, `malloc(-1)`, `vec.resize(-1)`...

Squareys: Hehe :D Fancy, aparently in my over then years with C++ I never actually came across this mistake 😂

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
Squareys: `cECFailedReadingSourceImages` is output on dimension issues (zero, which we catch before and larger than max dimensions, which would cause a pretty large allocation during test?), too many output slices, mipmap generation failure and during final sanity checks. I didn't find a obvious way to provoke any of those, though

`cECFailedValidating` could be provoked by settings texture type to cube map and passing non divisible by 6 amount of images.

`cECFailedFrontEnd` can be triggered with `m_params.m_max_endpoint_clusters >`basisu_frontend::cMaxEndpointClusters`, in case that isn't also checked beforehand.

`cECFailedBackend` has loads of code paths that lead to an error, but this code is accordingly complex and most code paths seem more like sanity checks than actual errors, maybe through the selector codebook parameter or something, but I think most of that is checked beforehand.

`cECFailedCreateBasisFile` checksum or transcoding failures, not sure how we'd produce one of these without messing with Basis' memory while it's running or creating a custom alloc and feeding that in or something 🤔

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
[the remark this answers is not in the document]
Squareys: So what's the conclusion here? Still do the flag fixup rather than flipping the image on copy?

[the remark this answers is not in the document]
mosra: Flag fixup, yeah. I'll submit a PR to Basis with this and it'll be easier to update that way.

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
mosra: For the first we could create an impossibly large ImageView with sized-but-null data view. That could work I think?

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.conf
```diff
@@ -0,0 +1,44 @@
...
+no_auto_global_sel_pal=false
+perceptual=true
+no_endpoint_rdo=false
+no_selector_rdo=false
+read_source_images=false
+write_output_basis_files=false
+compute_stats=false
+check_for_alpha=true
+force_alpha=false
+multithreading=true
```
mosra: Leftover? :) It's implicit from the `threads`, right?

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+Containers::Array<char> BasisImageConverter::doExportToData(const ImageView2D& image) {
+    /* Check input */
+    if(image.format() != PixelFormat::RGB8Unorm
+        && image.format() != PixelFormat::RGBA8Unorm
+        && image.format() != PixelFormat::RG8Unorm
+        && image.format() != PixelFormat::R8Unorm) {
+        Error() << "Trade::BasisImageConverter::exportToData(): unsupported format" << image.format();
+        return {};
+    }
```
mosra: I can add `Srgb` formats afterwards, but a related thing -- I don't see the `-linear` option being exposed, which switches between sRGB (the default) and linear.

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
```diff
@@ -0,0 +1,276 @@
...
+#include <Magnum/Trade/AbstractImageConverter.h>
+#include <Magnum/Trade/AbstractImporter.h>
+#include <Magnum/Trade/ImageData.h>
+#include <Magnum/Math/Color.h>
+
+#include <Corrade/Utility/Directory.h>
+
+#include "configure.h"
+
+using namespace Corrade;
```
mosra: :scream:

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
```diff
@@ -0,0 +1,276 @@
...
+    pngImporter->openFile(Utility::Directory::join(BASISIMPORTER_TEST_DIR, "rgb-27x63.png"));
+    const auto originalImage = pngImporter->image2D(0);
+
+    /* Create image with skip and padding for aligning rows to 4 bytes */
+    const Containers::Array<UnsignedByte> imageDataWithSkip{Containers::ValueInit,
+        ((27+8) + 1)*(63+7)};
+    const ImageView2D imageWithSkip{PixelStorage{}.setSkip({0, 1, 0}),
+        PixelFormat::R8Unorm, originalImage->size(), imageDataWithSkip};
+
+    auto sourcePixels = originalImage->pixels<Color4ub>();
```
mosra:
```suggestion
    auto sourcePixels = Containers::arrayCast<UnsignedByte>(originalImage->pixels<Color4ub>());
```

and then you can just assign without doing `.r()`. That's the "canonical" workflow -- have the loops as dumb as possible :)

Similar in the two cases below -- there it seems to still be the original range-loop.

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
```diff
@@ -0,0 +1,276 @@
...
+    auto destPixels = imageWithSkip.pixels<UnsignedByte>();
+    for(std::size_t y = 0; y != sourcePixels.size()[0]; ++y)
+        for(std::size_t x = 0; x != sourcePixels.size()[1]; ++x)
+            destPixels[y][x] = sourcePixels[y][x].r();
+
+    const auto compressedData = _converterManager.instantiate("BasisImageConverter")->exportToData(imageWithSkip);
+    CORRADE_VERIFY(compressedData);
+
+    if(_importerManager.loadState("BasisImporterRGBA8") == PluginManager::LoadState::NotFound)
+        CORRADE_SKIP("BasisImporterRGBA8 plugin not found, cannot test");
```
mosra: Just check for `BasisImporter` alone, this error message may be confusing for people who don't know the aliases. Same below.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -239,9 +239,7 @@ Containers::Optional<ImageData2D> BasisImporter::doImage2D(UnsignedInt index) {
        blows up for someone, we can reconsider. */
     CORRADE_INTERNAL_ASSERT_OUTPUT(_state->transcoder->get_image_level_desc(_state->in.data(), _state->in.size(), index, level, origWidth, origHeight, totalBlocks));
 
-    /* cDecodeFlagsPVRTCWrapAddressing is the default used by
-       transcode_image_level() */
-    const std::uint32_t flags = basist::basisu_transcoder::cDecodeFlagsPVRTCWrapAddressing;
+    const std::uint32_t flags = 0;
```
mosra: Yeah, Rich removed this two hours after I added that here :D I'll commit this soon into `master` (and bump the revisions for CIs).

mosra: Done: 66f73571db6b47c0e322cca0717f56cdce2de3df

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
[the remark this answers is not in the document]
Squareys: Because of the CRC/Checksum we decided to go with flipping on copy.

src/MagnumPlugins/BasisImageConverter/Test/CMakeLists.txt
```diff
@@ -44,6 +46,11 @@ if(NOT BUILD_PLUGINS_STATIC)
     if(WITH_BASISIMPORTER)
         set(BASISIMPORTER_PLUGIN_FILENAME $<TARGET_FILE:BasisImporter>)
     endif()
+    if(WITH_PNGIMPORTER)
+        set(PNGIMPORTER_PLUGIN_FILENAME $<TARGET_FILE:PngImporter>)
+    elseif(WITH_STBIMAGEIMPORTER)
+        set(PNGIMPORTER_PLUGIN_FILENAME $<TARGET_FILE:StbImageImporter>)
+    endif()
```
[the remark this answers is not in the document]
Squareys: Oh, I see 🙈

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.conf
Squareys: Yes

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
Squareys: no? 😅

mosra: it's implicit when including `Magnum/Magnum.h`, no?

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
Squareys: This leads to

~~~
D:\GitHub\vcpkg\installed\x64-windows\include\Corrade/Containers/StridedArrayView.h(823): error C2440: 'initializing': cannot convert from 'init
ializer list' to 'Corrade::Containers::StridedArrayView<2,Magnum::UnsignedByte>'
D:\GitHub\vcpkg\installed\x64-windows\include\Corrade/Containers/StridedArrayView.h(823): note: No constructor could take the source type, or co
nstructor overload resolution was ambiguous
..\src\MagnumPlugins\BasisImageConverter\Test\BasisImageConverterTest.cpp(136): note: see reference to function template instantiation 'Corrade:
:Containers::StridedArrayView<2,Magnum::UnsignedByte> Corrade::Containers::arrayCast<Magnum::UnsignedByte,2,const Magnum::Color3ub>(const Corrad
e::Containers::StridedArrayView<2,const Magnum::Color3ub> &)' being compiled
ninja: build stopped: subcommand failed.
~~~

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
Squareys: But I guess I would check for `BasisImporterRGBA8`, but just change the error message, right?

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
Squareys: Move it into the namespace?

Squareys: Oh, Magnum already has it, right, so this is simply superfluous

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
Squareys: > https://github.com/BinomialLLC/basis_universal/blob/master/basisu_tool.cpp#L286

It's "perceptual"

Squareys: Document / rename or add a parameter?

package/archlinux/PKGBUILD-emscripten-wasm-webgl2
```diff
@@ -32,6 +32,7 @@ build() {
         -DCMAKE_CXX_FLAGS="-DBASISD_SUPPORT_BC7=0 -DBASISD_SUPPORT_ATC=0 -DBASISD_SUPPORT_FXT1=0 -DBASISD_SUPPORT_PVRTC2=0" \
         -DBASIS_UNIVERSAL_DIR=/home/mosra/Code/basis_universal \
         -DWITH_BASISIMPORTER=ON \
+        -DWITH_BASISIMAGECONVERTER=OFF \
```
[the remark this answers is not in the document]
Squareys: And mingw. Done.

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
Squareys: That crashes before it even reaches .process() :D Especially since the data needs to be copied anyway

Squareys: `cECFailedValidating` => we don't expose texture type because we can't input more than one image anyway. I don't think there is a way to provoke this one either.

`cECFailedFrontEnd` => ✅ 

And I think that's the only one.

Squareys:
@mosra Rebased, let's see what the CIs say!

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
mosra: ookay, skip that, then :)

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
mosra: wtf

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
mosra: eh :/ yeah, can you document that? directly in the `BasisImageConverter.conf`

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
mosra: If there's `BasisImporter` and it doesn't provide a `BasisImporterRGBA8` alias, then there's something very off and it should blow up (while this would lead to just `SKIP`) ... so pick the loudest alternative :D

But this is a minor thing, nothing critical.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
Squareys: Nice, so awesome to be able to be able to document that directly there, done 👍

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
Squareys: Well, not necessarily, could just be compiled out or an older version 🤷‍♂ 

Not sure what your comment means in practice, shall I remove the RGBA8 prefix from both the skip message and the loadState call?

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
Squareys: `Containers::arrayCast<const UnsignedByte>(...` was the solution, just a missing `const`.

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
mosra: Yes, remove both.

mosra:
Merged, thank you! :tada:

PR #66, opened 2019-09-20
AssimpImporter: populate right texture indices

ikalevatykh:
Ambient/diffuse/specular texture indices are not populated in the current version. They are always 0 even a material has several textures or a model has several materials.

### DIFF
--- a/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
+++ b/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
@@ -550,15 +550,26 @@ Containers::Pointer<AbstractMaterialData> AssimpImporter::doMaterial(const Unsig
     }
 
     PhongMaterialData::Flags flags;
-    aiString texturePath;
+    UnsignedInt textureIndices[3];
     aiColor3D color;
 
-    if(mat->Get(AI_MATKEY_TEXTURE(aiTextureType_AMBIENT, 0), texturePath) == AI_SUCCESS)
+    /* Start from first possible texture index for this material */
+    auto textureIndex = _f->textureIndices[mat];
+    if (textureIndex < _f->textures.size() &&
+        _f->textures[textureIndex] == std::make_pair(mat, aiTextureType_AMBIENT)) {
         flags |= PhongMaterialData::Flag::AmbientTexture;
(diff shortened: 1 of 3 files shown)


mosra:
Hahah guess what -- I *just* pushed the *same* fix, twenty minutes ago.

I wonder how did this happen, am I your evil twin? :D

ikalevatykh:
Hhaha, should I already be scared? 8)

mosra:
No, seriously, are we working on the same projects? :D Because your other bugfix, few days ago, got reported to me recently as well and there I could say "oh, it got *just* fixed, pull latest master".

ikalevatykh:
Heh, in any case, it is nice that it is already fixed, thank you for your titanic work) 

Sorry for off-topic, have you plan to do something with this: https://github.com/mosra/magnum-plugins/blob/b9c454af2d16484f2f0e9ad2b004cb94ba46a108/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp#L304 . Maybe add dummy child nodes for each of meshes? On another case, I will propose some fix.

mosra:
The fixes I'm doing are for various 3D-scanned datasets used by [AI Habitat](https://aihabitat.org) and I guess that's where we overlap :)

> Sorry for off-topic

Not a problem -- I actually wanted to propose that next time you discover some bug or missing feature in the Assimp importer, let me know before you start implementing. In case I'm already on it, it'll save you time, if I'm not yet, then you'll save time me :)

Multiple meshes per node -- I did a similar fix for TinyGltfImporter some time ago in da587349599f976894f5b96b10dc44bab0921d82, and I think it could be done very similarly here -- see [the overview here](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1TinyGltfImporter.html#Trade-TinyGltfImporter-limitations-meshes). I did it this way to have animations attached to particular nodes still work properly, inserting dummy children would break those.

mosra:
So, my fix is in 503a6d93edc88e3634e3b2de00159540da9bc0f4 and I had to give you a mention as well (8e6f49946e2de39b03eff85cf7a2a3c5d42b175c) :) Both pushed to `master` now.

Closing this one as resolved, but feel free to reuse this PR for further questions about the multi-mesh support (and thanks in advance for looking into it).

ikalevatykh:
Ok, thank you!

PR #67, opened 2019-09-22
Support all formats, cubemaps and arrays in DdsImporter, make it more robust

mosra:
This was all fun and games until I tried to open real-world files and realized it doesn't work at all. What needs to be done:

- [x] Turn the switch into a table -- 401ea7778c040dd2152d3a7ae3a23fe710414350
- [x] GLI supports ASTC formats in DXGI since 2015, how standard is that?! https://github.com/g-truc/gli/commit/0ef1b27124da5671b0043e26c2007394ffc4b7a0 ... apparently it was "added by microsoft and then removed again"? haha https://forums.developer.nvidia.com/t/nv-tt-exporter-astc-compression/122477
  - [x] There's an ASTC DDS file at https://zenhax.com/viewtopic.php?t=8684 but it spectacularly fails to load with current code. Try to get NVidia Texture Tools Exporter somehow (needs an account) and make my own files for testing.
- [x] Reduce existing excessive tests a bit (gives a false impression that it's tested A LOT, while not really) -- a64351af71497de5500c009dedb77152c0edbaa1
  - [x] Test also files with incomplete compressed blocks -- 3a9f67bf02ac20fe120f7c89a885f9b27f5aec28
  - [x] Test cube maps and arrays! and cube map arrays!
- [x] Better handling of short files -- d9617726f6a818496d87ce210475be915b0aef21
- [x] Actually cover RGBA, BGRA and greyscale handling with tests -- bfb89306a6bf436dbeed9407c9ccbfd9fcbbc68b, 28a2fa3af5e2bdd2a4fd4b8249c65189ce9deb38
- [x] Support DXT10 swizzled formats -- 0be45f3afa7f7d90ab844e8805de8b8ab1a218c5
- [x] Support depth/stencil formats as we have them in `PixelFormat` now -- 2e715481a61b8ab1c1f628c6408f2e38a03e6edc
- [x] Support remaining compressed formats -- a477af52e8cb40d2875a5ea3205b8d9e18041f87
  - [x] crafting *proper* test files in random compression types (especially varying compressed block size / block data size), because that's currently totally untested -- using an external tool such as AMD Compressonator
- [x] Y-flip for uncompressed data, warning for compressed. All other plugins do this, this one somehow never did. The horror!
  - [x] ~~Does DDS have any way to describe origin/coordinate system? I don't think so.~~ No.
  - [x] Provide an `assumeYUp` setting to not flip / hide the warning
- [x] building up a "knowledge base" table with compressed block sizes and data sizes for each generic format we have directly in magnum -- mosra/magnum@75bfd41baf1831f4597607a375271c4f77ce4edd
- [x] ~~reworking `CompressedImage(View|Data)` to check against those, similarly to what `Image` does for pixels already~~ not critically needed for this PR


mosra:
Major part of the changes is merged in 65381f8e6db9a0ed1f5187c296667a7159c4ae93...ee51ae45ffad0d6c884fd14781adc464bd735bb2. This again exploded way more than expected and motivation dried out.

What remains is looking closer at ASTC and verifying with a second independent source, and cubemaps/arrays. Support for cubemaps, arrays, mips and especially combinations of those is either nonexistent or broken in majority of tools I tried, so it's not that important to be able to read them. When such a self-contained file is needed, it can be always easily extracted out of separately compressed DDS files and then combined into a KTX using `magnum-imageconverter --layers`.

PR #68, opened 2019-09-24
AssimpImporter: Support multiple meshes per node

ikalevatykh:
If a node has several meshes, creates one Object3D per mesh.
 
object3DCount() query returns a number of all nodes extended with a number of extra nodes for each additional mesh.
object3DForName() points to the first object containing the first mesh, 
object3DName() returns the same name for all objects in the given sequence.
doObject3D() for multi-meshes returns node for first mesh as usual but also returns indices of extra meshes nodes as children.


mosra:
@ikalevatykh any chance you could have a look at the tests & docs? It would be great to have this in for the release, but I probably won't be able to find time to add those myself.

Thank you!

ikalevatykh:
@mosra I had to switch to another task but I will do it Thursday

mosra:
You're great, thanks!

mosra:
Merged a variant of this in #81. Thanks for providing the initial version! :+1:

PR #72, opened 2019-10-30
Make FindBasisUniversal search libraries instead of pkgconfig

Squareys:
Hi @mosra !

As promised, here's a version of the FindBasisUniversal that works with vcpkg.
I have a maybe unconventional use of `find_package_handle_standard_args` in there for setting the `BasisUniversal_(Encoder|Transcoder_FOUND` vars. It causes output when both are found and I'm interested in your judgment whether that is a good trade-off for the more concise code.

Best,
Jonathan

### DIFF
--- a/modules/FindBasisUniversal.cmake
+++ b/modules/FindBasisUniversal.cmake
@@ -51,48 +51,6 @@
 #   DEALINGS IN THE SOFTWARE.
 #
 
-# Vcpkg distributes imgui as a library with a config file, so try that first --
-# but only if BASIS_UNIVERSAL_DIR wasn't explicitly passed, in which case we'll
-# look there instead
-find_package(basisu CONFIG QUIET)
-if(basisu_FOUND AND NOT BASIS_UNIVERSAL_DIR)
-    if(NOT TARGET BasisUniversal::Transcoder)
-        add_library(BasisUniversal::Transcoder INTERFACE IMPORTED)
-        set_property(TARGET BasisUniversal::Transcoder APPEND PROPERTY
-            INTERFACE_LINK_LIBRARIES basisu_transcoder)
-
-        # Retrieve include directory for FindPackageHandleStandardArgs later
-        get_target_property(BasisUniversalTranscoder_INCLUDE_DIR basisu_transcoder
-            INTERFACE_INCLUDE_DIRECTORIES)
-    endif()
-
-    if(NOT TARGET BasisUniversal::Encoder)
-        add_library(BasisUniversal::Encoder INTERFACE IMPORTED)
-        set_property(TARGET BasisUniversal::Encoder APPEND PROPERTY
-            INTERFACE_LINK_LIBRARIES basisu_encoder)
-
-        # Retrieve include directory for FindPackageHandleStandardArgs later
-        get_target_property(BasisUniversalEncoder_INCLUDE_DIR basisu_encoder
-            INTERFACE_INCLUDE_DIRECTORIES)
-    endif()
-
-# Otherwise find the source files and compile them as part of the library they
-# get linked to
-else()
-    # Disable the find root path here, it overrides the
-    # CMAKE_FIND_ROOT_PATH_MODE_INCLUDE setting potentially set in
-    # toolchains.
-    find_path(BasisUniversalEncoder_INCLUDE_DIR NAMES basisu_frontend.h HINTS
-        HINTS "${BASIS_UNIVERSAL_DIR}" "${BASIS_UNIVERSAL_DIR}/encoder"
-        NO_CMAKE_FIND_ROOT_PATH)
-    mark_as_advanced(BasisUniversalEncoder_INCLUDE_DIR)
-
-    find_path(BasisUniversalTranscoder_INCLUDE_DIR NAMES basisu_transcoder.h
-        HINTS "${BASIS_UNIVERSAL_DIR}/transcoder" "${BASIS_UNIVERSAL_DIR}"
-        NO_CMAKE_FIND_ROOT_PATH)
-    mark_as_advanced(BasisUniversalTranscoder_INCLUDE_DIR)
-endif()
-
 list(FIND BasisUniversal_FIND_COMPONENTS "Encoder" _index)
 if(${_index} GREATER -1)
     list(APPEND BasisUniversal_FIND_COMPONENTS "Transcoder")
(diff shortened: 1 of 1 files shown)


modules/FindBasisUniversal.cmake
```diff
@@ -115,100 +73,162 @@ macro(_basis_setup_source_file source)
...
 # Find components
 foreach(_component ${BasisUniversal_FIND_COMPONENTS})
     if(_component STREQUAL "Encoder")
         if(NOT TARGET BasisUniversal::Encoder)
-            find_path(BasisUniversalEncoder_DIR NAMES basisu_frontend.cpp
-                HINTS "${BASIS_UNIVERSAL_DIR}" "${BASIS_UNIVERSAL_DIR}/encoder"
-                NO_CMAKE_FIND_ROOT_PATH)
-            if(BasisUniversalEncoder_DIR)
-                set(BasisUniversal_Encoder_FOUND TRUE)
+            # Try to find compiled library, e.g. when installed through vcpkg
```
mosra: It's not `e.g.`, but rather *specific* to vcpkg -- there's no other case where this would be true.

Except ... if jherico's fork would be built and installed manually. But in that case, the two `if()`s below should check for debug `OR` release libs, as one can build & install just one of them.

modules/FindBasisUniversal.cmake
```diff
@@ -115,100 +73,162 @@ macro(_basis_setup_source_file source)
...
+                    IMPORTED_LOCATION_DEBUG ${BasisUniversalEncoder_LIBRARY_DEBUG})
+                set_property(TARGET BasisUniversal::Encoder APPEND PROPERTY
+                    IMPORTED_LOCATION_RELEASE ${BasisUniversalEncoder_LIBRARY_RELEASE})
+                set_property(TARGET BasisUniversal::Encoder APPEND PROPERTY
+                    INTERFACE_INCLUDE_DIRECTORIES ${BasisUniversalTranscoder_INCLUDE_DIR})
+
+                find_package_handle_standard_args(BasisUniversal_Encoder
+                    BasisUniversalEncoder_LIBRARY_DEBUG
+                    BasisUniversalEncoder_LIBRARY_RELEASE
+                    BasisUniversalEncoder_INCLUDE_DIR)
```
mosra: Calling this here would mean the `-- Found BasisUniversal` message gets printed multiple times. This should be setting the `BasisUniversal_Encoder_FOUND` variable instead if all the relevant variables get found, like it does in the `else()` branch. For example,

```cmake
if((BasisUniversalEncoder_LIBRARY_DEBUG OR BasisUniversalEncoder_LIBRARY_RELEASE) AND BasisUniversalEncoder_INCLUDE_DIR)
    set(BasisUniversal_Encoder_FOUND TRUE)
endif()
```

Same for the transcoder below.

modules/FindBasisUniversal.cmake
```diff
@@ -115,100 +73,162 @@ macro(_basis_setup_source_file source)
...
-                INTERFACE_SOURCES "${BasisUniversalTranscoder_SOURCES}")
         else()
             set(BasisUniversal_Transcoder_FOUND TRUE)
         endif()
     endif()
 endforeach()
 
-include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(BasisUniversal
     REQUIRED_VARS BasisUniversalTranscoder_INCLUDE_DIR BasisUniversalEncoder_INCLUDE_DIR
```
mosra: This started failing in the MinGW CI build (which doesn't build the encoder AFAIK), moreover this fails in case only either the encoder or transcoder is found (for example, for Habitat, I'm bundling just the transcoder sources instead of the whole thing).

I think keeping just `HANDLE_COMPONENTS` would work, as the component `_FOUND` variables are set properly above (but please verify that it actually does fail properly when nothing is found :)).

modules/FindBasisUniversal.cmake
Squareys: Well, should I create my own cmake project on basis, then that is another example for when this would be used.

mosra: Yeah, then the `OR` note applies as well. Can you do this change, then (and then wrapping the `set_property()` in appropriate `if()`s as well)?

modules/FindBasisUniversal.cmake
Squareys: You mean `HANDLE_COMPONENTS` without the two include dirs?

mosra: yeah,

```cmake
find_package_handle_standard_args(BasisUniversal
    HANDLE_COMPONENTS)
```

modules/FindBasisUniversal.cmake
```diff
@@ -119,96 +77,163 @@ endmacro()
...
 endforeach()
 
+if(BasisUniversalEncoder_INCLUDE_DIR)
+    set(_BASISUNIVERSAL_VARS BasisUniversalEncoder_INCLUDE_DIR)
+else()
+    set(_BASISUNIVERSAL_VARS BasisUniversalTranscoder_INCLUDE_DIR)
+endif()
+
 include(FindPackageHandleStandardArgs)
+set(DUMMY_INCLUDE_DIR ${BasisUniversalTranscoder_INCLUDE_DIR} ${BasisUniversalEncoder_INCLUDE_DIR})
```
mosra: this is not used anywhere

modules/FindBasisUniversal.cmake
```diff
@@ -119,96 +77,163 @@ endmacro()
...
         if(NOT TARGET BasisUniversal::Transcoder)
-            find_path(BasisUniversalTranscoder_DIR NAMES basisu_transcoder.cpp
-                HINTS "${BASIS_UNIVERSAL_DIR}/transcoder" "${BASIS_UNIVERSAL_DIR}"
-                NO_CMAKE_FIND_ROOT_PATH)
-            if(BasisUniversalTranscoder_DIR)
-                set(BasisUniversal_Transcoder_FOUND TRUE)
+            find_library(BasisUniversalTranscoder_LIBRARY_RELEASE basisu_transcoder)
+            find_library(BasisUniversalTranscoder_LIBRARY_DEBUG basisu_transcoder HINTS "debug")
+
+            if(BasisUniversalTranscoder_LIBRARY_DEBUG AND BasisUniversalTranscoder_LIBRARY_RELEASE)
```
mosra: `OR` here, and then wrap the `set_property()` for the libraries below in corresponding `if()`s

modules/FindBasisUniversal.cmake
```diff
@@ -119,96 +77,163 @@ endmacro()
...
-            find_path(BasisUniversalEncoder_DIR NAMES basisu_frontend.cpp
-                HINTS "${BASIS_UNIVERSAL_DIR}" "${BASIS_UNIVERSAL_DIR}/encoder"
-                NO_CMAKE_FIND_ROOT_PATH)
-            if(BasisUniversalEncoder_DIR)
-                set(BasisUniversal_Encoder_FOUND TRUE)
+            # Try to find compiled library when installed through vcpkg
+            find_library(BasisUniversalEncoder_LIBRARY_RELEASE basisu_encoder)
+            find_library(BasisUniversalEncoder_LIBRARY_DEBUG basisu_encoder HINTS "debug")
+
+            if(BasisUniversalEncoder_LIBRARY_DEBUG AND BasisUniversalEncoder_LIBRARY_RELEASE)
```
mosra: `OR` here, and then wrap the `set_property()` for the libraries below in corresponding `if()`s

mosra:
All my local tests worked, so merged, thank you!

PR #73, opened 2019-11-11
StbTrueTypeFont: Use fabsf instead of fabs

Squareys:
Hey @mosra !

As discussed via gitter, the performance of StbTrueTypeFont on MSVC Debug builds is crazy bad. Through profiling I figured out that all of the time is spent in  `fabs()` during rasterization of the Glyphs in `fillGlyphCache`.
In embedded stbfreetype in ImGui this is not an issue for some reason, and I found that they override `STBTT_fabs(x)` with `fabsf(x)`, the `float` override of `fabs`.
This didn't entirely fix the issue (imgui is still over 10x faster at loading a font), but at least 30% speedup.

I tried to figure out if there are different compile flags applied during compilation of imgui, but that doesn't seem to be the case.

Best, 
Jonathan.

### DIFF
--- a/src/MagnumPlugins/StbTrueTypeFont/StbTrueTypeFont.cpp
+++ b/src/MagnumPlugins/StbTrueTypeFont/StbTrueTypeFont.cpp
@@ -35,6 +35,10 @@
 
 #define STB_TRUETYPE_IMPLEMENTATION
 #define STBTT_STATIC
+#include <math.h>
+/* Use fabsf instead of fabs (double version) for 30% performance
+   improvement on MSVC Debug builds */
+#define STBTT_fabs(x) fabsf(x)
 #include "stb_truetype.h"
 
 namespace Magnum { namespace Text {


src/MagnumPlugins/StbTrueTypeFont/StbTrueTypeFont.cpp
```diff
@@ -35,6 +35,10 @@
 
 #define STB_TRUETYPE_IMPLEMENTATION
 #define STBTT_STATIC
+#include <math.h>
+/* Use fabsf instead of fabs (double version) for 30% performance
+   improvement on MSVC Debug builds */
+#define STBTT_fabs(x) fabsf(x)
```
mosra: :+1: for the comment explaining why

mosra:
Thank you!

mosra:
Hmm. This causes GCC to emit a warning:

```
../src/external/stb/stb_truetype.h: In function ‘void stbtt__rasterize_sorted_edges(stbtt__bitmap*, stbtt__edge*, int, int, int, int, void*)’:
../src/MagnumPlugins/StbTrueTypeFont/StbTrueTypeFont.cpp:41:30: warning: use of old-style cast to ‘float’ [-Wold-style-cast]
   41 | #define STBTT_fabs(x) fabsf(x)
      |                              ^
```

I'll need to investigate why it's using doubles at all. Maybe that's responsible for the remaining perf drop.

PR #77, opened 2019-12-19
TinyGltfImporter: Update to current HEAD and KHR_lights_punctual

Squareys:
Hi @mosra !

This was surprisingly easy, I wonder what I missed... I had to update tiny_gltf to get support for `KHR_lights_punctual`, in the process I adapted the importer to API changes there. Since the new light extension is pretty similar to the old, this was pretty straight forward.

I made a minor adaption to the test, from the documentation `Linear` was the default all along and therefore I wonder why the test was checking for `Nearest` as the minFilter was not set in the file.

Cheers,
Jonathan.


src/MagnumExternal/TinyGltf/tiny_gltf.h
```diff
@@ -40,16 +52,46 @@
...
+  x(x &&) TINYGLTF_NOEXCEPT = default; \
+  x &operator=(const x &) = default;   \
+  x &operator=(x &&) TINYGLTF_NOEXCEPT = default;
+
 namespace tinygltf {
 
 #define TINYGLTF_MODE_POINTS (0)
 #define TINYGLTF_MODE_LINE (1)
 #define TINYGLTF_MODE_LINE_LOOP (2)
+#define TINYGLTF_MODE_LINE_STRIP (3)
```
mosra: The importer code could be updated to use the constant now (it used just 3 before).

src/MagnumPlugins/TinyGltfImporter/Test/TinyGltfImporterTest.cpp
```diff
@@ -1945,7 +1945,7 @@ void TinyGltfImporterTest::texture() {
     CORRADE_COMPARE(texture->type(), TextureData::Type::Texture2D);
 
     CORRADE_COMPARE(texture->magnificationFilter(), SamplerFilter::Linear);
-    CORRADE_COMPARE(texture->minificationFilter(), SamplerFilter::Nearest);
+    CORRADE_COMPARE(texture->minificationFilter(), SamplerFilter::Linear);
```
mosra: A more explicit test for this would be nice (testing all defaults as mentioned in the plugin docs).

src/MagnumExternal/TinyGltf/tiny_gltf.h
```diff
@@ -3686,46 +6141,91 @@ static void SerializeNumberProperty(const std::string &key, T number,
...
+  JsonReserveArray(ary, value.size());
+  for (auto &s : value) {
+    JsonPushBack(ary, JsonFromString(s.c_str()));
   }
-
-  obj[key] = vals;
+  JsonAddMember(obj, key.c_str(), std::move(ary));
 }
 
 static bool ValueToJson(const Value &value, json *ret) {
```
mosra: Too bad tinygltf doesn't have a define for compiling out all the serialization stuff we don't need :/ I assume that is also what needed the json.hpp update?

src/MagnumExternal/TinyGltf/json.hpp
```diff
@@ -1,7 +1,7 @@
 /*
     __ _____ _____ _____
  __|  |   __|     |   | |  JSON for Modern C++
-|  |  |__   |  |  | | | |  version 3.3.0
+|  |  |__   |  |  | | | |  version 3.5.0
```
mosra: We need less of this thing, not more 😅. Any comment about why is it needed?

Reminds me of https://twitter.com/CaliaDomenico/status/1207922171902906369

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
```diff
@@ -1287,18 +1285,26 @@ Containers::Optional<TextureData> TinyGltfImporter::doTexture(const UnsignedInt
             minFilter = SamplerFilter::Linear;
             mipmap = SamplerMipmap::Linear;
             break;
-        default: CORRADE_ASSERT_UNREACHABLE(); /* LCOV_EXCL_LINE */
+        default:
```
mosra: Maybe add `case -1` right below `case TINYGLTF_TEXTURE_FILTER_LINEAR_MIPMAP_LINEAR` and keep an unreachable default instead? So we stay as strict as before.

Same below, I think this could also fix the minor coverage loss.

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
```diff
@@ -104,9 +104,7 @@ bool loadImageData(tinygltf::Image* image, std::string*, std::string*, int, int,
 }
 
 std::size_t elementSize(const tinygltf::Accessor& accessor) {
-    /* GetTypeSizeInBytes() is totally bogus and misleading name, it should
-       have been called GetTypeComponentCount but who am I to judge. */
```
mosra: 😆

src/MagnumExternal/TinyGltf/json.hpp
Squareys: I just updated it to the latest version that tiny_gltf uses (the exact file bundled with tiny_gltf). I can revert this, though

src/MagnumPlugins/TinyGltfImporter/Test/TinyGltfImporterTest.cpp
Squareys: There is `textureDefaultSampler already` 🤔

mosra: Oh. Okay then (sorry, should have checked better).

Squareys:
Alright, updated. The defaults test was there already, I am really unsure how this passed before (looking at the diff, I didn't touch the tested file).

I was able to omit the json.hpp update (which [was only needed for clang with c++17](https://github.com/syoyo/tinygltf/commit/c0d0251e2c7f2273a894aa1b125af9313a88eae4)).

mosra:
Ah! I see now, there's two cases -- one is with `sampler` missing completely (which is tested properly), second is with `sampler` parameters missing (right now only wrapping seems to be tested, not min/mag filters at all).

Could you expand the `texture.gltf` to list all properties, and then add a `texture-empty-sampler.gltf` that has just an empty sampler, verifying the defaults are picked correctly? No need to add the glb/packed variants in this case I think. Thanks!

Squareys:
Crazy that you figured that out so quickly. I made the requested changes.

src/MagnumPlugins/TinyGltfImporter/Test/texture-empty-sampler.gltf
```diff
@@ -0,0 +1,183 @@
...
+            "sampler": 0,
+            "source": 0
+        },
+        {
+            "name": "Texture",
+            "sampler" : 0,
+            "source" : 0
+        }
+    ]
+}
```
mosra: This could rather be adapted from `texture-default-sampler.gltf` as there's much less unnecessary cruft :) Just add a samplers array to it and reference it from the texture.

Squareys: Ah, right, sorry

Squareys: Fixed! Sorry again, didn't pay attention there

PR #79, opened 2020-02-11
New plugin: IcoImporter

bowling-allie:


src/MagnumPlugins/IcoImporter/IcoImporter.h
```diff
@@ -0,0 +1,92 @@
...
+        MAGNUM_ICOIMPORTER_LOCAL Features doFeatures() const override;
+        MAGNUM_ICOIMPORTER_LOCAL bool doIsOpened() const override;
+        MAGNUM_ICOIMPORTER_LOCAL void doClose() override;
+        MAGNUM_ICOIMPORTER_LOCAL void doOpenData(Containers::ArrayView<const char> data) override;
+
+        MAGNUM_ICOIMPORTER_LOCAL UnsignedInt doImage2DCount() const override;
+        MAGNUM_ICOIMPORTER_LOCAL Containers::Optional<ImageData2D> doImage2D(UnsignedInt id) override;
+
+        bool _isOpened = false;
+        ImageType _imageType{};
```
mosra: This isn't used in any way, right? Only written to.

At some point in the future this importer could get used to load cursors also (and then the origin/hotspot coordinate would be needed), but until then I don't think we need to bother with those (also it seems CURs can't be PNGs anyway, which is another reason to ignore them).

src/MagnumPlugins/IcoImporter/IcoImporter.h
```diff
@@ -0,0 +1,92 @@
...
+        UnsignedShort _imageCount{};
```
mosra: This is available through `_imageDataArray.size()` already, no need to store the same state twice I think.

I'd say `_isOpened` could be also inferred from there (file is opened if `_imageDataArray` is non-null), but this handles the pathological case of zero-image `*.ico` properly, so leave it as-is I guess :)

CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ option(WITH_DRWAVAUDIOIMPORTER "Build DrWavAudioImporter plugin" OFF)
 option(WITH_FAAD2AUDIOIMPORTER "Build Faad2AudioImporter plugin" OFF)
 cmake_dependent_option(WITH_FREETYPEFONT "Build FreeTypeFont plugin" OFF "NOT WITH_HARFBUZZFONT" ON)
 option(WITH_HARFBUZZFONT "Build HarfBuzzFont plugin" OFF)
+option(WITH_ICOIMPORTER "Build IcoImporter plugin" OFF)
```
mosra: Would be great if you could enable this in all `*.bat` and `*.sh` file in `package/ci/` so we can see that it builds & runs properly everywhere.

src/MagnumPlugins/IcoImporter/IcoImporter.cpp
```diff
@@ -0,0 +1,256 @@
...
+bool IcoImporter::doIsOpened() const { return _isOpened; }
+
+void IcoImporter::doClose() {
+    _isOpened = false;
+    _imageDataArray = nullptr;
+}
+
+namespace {
+    constexpr std::initializer_list<const char> bmpFileHeaderTemplate{'B', 'M', 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+    constexpr std::initializer_list<const char> pngMagicNum{static_cast<char>(0x89), 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a};
```
mosra: IIRC `initializer_list` isn't C++11, so this won't work (the CIs will complain). What about a plain `const char pngMagicNum[] { ... }`, and then 

```cpp
std::equal(std::begin(pngMagicNum), std::end(pngMagicNum), imageDataBegin)
```

in the code below?

bowling-allie: hmm, https://en.cppreference.com/w/cpp/utility/initializer_list says it is c++11

CMakeLists.txt
bowling-allie: ah, forgot about CI!

src/MagnumPlugins/IcoImporter/IcoImporter.cpp
mosra: Blargh, sorry, forgot a word, it should have said "isn't **constexpr** in C++11".

src/MagnumPlugins/IcoImporter/Test/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+else()
+    set(ICOIMPORTER_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+endif()
+
+# CMake before 3.8 has broken $<TARGET_FILE*> expressions for iOS (see
+# https://gitlab.kitware.com/cmake/cmake/merge_requests/404) and since Corrade
+# doesn't support dynamic plugins on iOS, this sorta works around that. Should
+# be revisited when updating Travis to newer Xcode (xcode7.3 has CMake 3.6).
+if(NOT BUILD_PLUGINS_STATIC)
+    set(ICOIMPORTER_PLUGIN_FILENAME $<TARGET_FILE:IcoImporter>)
```
mosra: To make the CI pass also for dynamic builds, you'll need to add this here:

```cmake
    if(WITH_STBIMAGEIMPORTER)
        set(STBIMAGEIMPORTER_PLUGIN_FILENAME $<TARGET_FILE:StbImageImporter>)
    endif()
```

and then this to `Test/configure.h.cmake`:

```cpp
#cmakedefine STBIMAGEIMPORTER_PLUGIN_FILENAME "${STBIMAGEIMPORTER_PLUGIN_FILENAME}"
```

and finally load the plugin in the test, right after loading the IcoImporter:

```cpp
    #ifdef STBIMAGEIMPORTER_PLUGIN_FILENAME
    CORRADE_INTERNAL_ASSERT(_manager.load(STBIMAGEIMPORTER_PLUGIN_FILENAME) & PluginManager::LoadState::Loaded);
    #endif
```

Hope this helps :)

bowling-allie:
Ok, I'm not sure what causes the failure.

Tests run on emscripten pass on my machine... except when i use the ci script

mosra:
I'm getting those too (and only on the CI), somehow the previous emscripten versions don't handle unaligned reads/stores well and the new ones do. That's a minor thing tho, I can look at that when merging.

What's left to do here? IIRC it's just removing the commented-out BMP code (or should we keep it?) and adding tests for the (currently uncovered) error handling code paths.

mosra:
(Finally) merged in 8761c97556df412dd4227722c60c6e566a4d5355. I removed the commented-out parts that handle BMP loading and extended the tests a bit to verify also the "file too short" cases. Turns out the DevIlImageImporter can handle BMP ICOs so there's at least something for those (OTOH it crashes with *some* PNG ICOs and produces bad output for other, heh.)

Thanks for pushing this 90% of the way :)

PR #81, opened 2020-03-12
AssimpImporter: support multiple meshes per Assimp node

xqms:
Since `Magnum::MeshObjectData3D` supports only one mesh instance, the importer has until now silently ignored any additional meshes per scene node.

We now add additional child objects for any extra meshes. This adds some ugliness to `doObjectData3D()` and friends, because we need to differentiate between "original" nodes referring to Assimp nodes and the new "additional mesh" objects.

This also adds a unit test using a hybrid quad/line mesh exported from blender.

### DIFF
--- a/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
+++ b/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
@@ -33,6 +33,7 @@
 #include <Corrade/Utility/ConfigurationGroup.h>
 #include <Corrade/Utility/DebugStl.h>
 #include <Corrade/Utility/Directory.h>
+#include <Corrade/Utility/FormatStl.h>
 #include <Corrade/Utility/String.h>
 #include <Magnum/FileCallback.h>
 #include <Magnum/Mesh.h>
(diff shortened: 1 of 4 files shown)


mosra:
Hi, happy to see you around :)

Before I look at the code -- there was a variant of this submitted a while ago (in #68, without tests tho), is the handling in this PR different compared to that one? TinyGltfImporter implements this handling as well (see its [docs](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1TinyGltfImporter.html#Trade-TinyGltfImporter-limitations-meshes)), so maybe having the same behavior in both could make sense? (A potentially time-saving idea would be to reuse the same glTF test file for both and verifying both give the same result, but I have no idea how stable Assimp's glTF support is at this point.)

Thank you, as always!

xqms:
Thanks for the hint, I'll take a look at #68 and TinyGltfImporter :)

mosra:
Btw., I'm not saying the approach implemented there is the best possible -- if you see any shortcomings in it, I'm open for suggestions how to make it better :)

xqms:
Well, at first glance it looks very similar to what I came up with, so of course it must be a great idea :D
The biggest difference seems to be the ordering of the objects and I'll adapt that to be consistent with TinyGltfImporter.

xqms:
Okay, I adapted the logic to be consistent with TinyGltfImporter and updated the documentation in AssimpImporter.h.

Unit test: Assimp (at least my stock version on Ubuntu 18.04, which is admittedly a bit old) won't load the glTF test file from TinyGltfImporter (just returns an empty scene, no meshes). I'll try with a more recent Assimp version tomorrow. Currently, we only have my more basic Collada test with quad/line primitives.

xqms:
Updated again with a port of the TinyGltfImporter test case. I had to convert the example to COLLADA, because Assimp would just give me an empty scene from the gltf file.
I also had to add at least one triangle / line to each mesh, since Assimp strips out empty meshes.
Finally, the point primitive is not supported by COLLADA as far as I know, so I removed the point mesh.

I guess this is ready for review again, but there's no rush on my end ;)

mosra:
Thank you! 

The test was failing on Assimp 3.2 but I couldn't be bothered trying to find a variant that works there, so it's disabled on that ancient version (on the macOS CI runner it's 4.1 and there it's tested).

PR #83, opened 2020-05-02
Material texture layer support for TinyGltfImporter and AssimpImporter

Squareys:
Hi @mosra !

Here's the accompanying PR for mosra/magnum#438. 

**TODOs:**

 - [x] Compile check
 - [x] Add configuration value for textures sets
 - [x] Testing (A gltf+glb file with a different layer for each, reusing that for assimp, too)
 - [x] Advertise support

### DIFF
--- a/src/MagnumPlugins/AssimpImporter/AssimpImporter.conf
+++ b/src/MagnumPlugins/AssimpImporter/AssimpImporter.conf
@@ -29,6 +29,11 @@ provides=XglImporter
 # AI_CONFIG_* values, can be changed only before the first file is opened
 ImportColladaIgnoreUpDirection=false
 
+# Allow non-default texture coordinate sets. If disabled, materials with
+# non-zero texture coordinate sets (which need explicit support in shaders)
+# will fail to import.
+allowMaterialTextureCoordinateSets=false
+
 # aiPostProcessSteps, applied to each opened file
 [configuration/postprocess]
 JoinIdenticalVertices=true
--- a/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
+++ b/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
@@ -3,7 +3,7 @@
 
     Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019
               Vladimír Vondruš <mosra@centrum.cz>
-    Copyright © 2017 Jonathan Hale <squareys@googlemail.com>
+    Copyright © 2017, 2020 Jonathan Hale <squareys@googlemail.com>
     Copyright © 2018 Konstantinos Chatzilygeroudis <costashatz@gmail.com>
     Copyright © 2019, 2020 Max Schwarz <max.schwarz@ais.uni-bonn.de>
 
@@ -786,10 +786,23 @@ Containers::Pointer<AbstractMaterialData> AssimpImporter::doMaterial(const Unsig
        something else, so we just don't import it at all. */
     aiColor3D ambientColor;
     UnsignedInt ambientTexture{};
+    UnsignedInt ambientCoordinateSet{};
     mat->Get(AI_MATKEY_COLOR_AMBIENT, ambientColor);
-    if(flags & PhongMaterialData::Flag::AmbientTexture)
+    if(flags & PhongMaterialData::Flag::AmbientTexture) {
         ambientTexture = firstTextureIndex++;
-    else {
+        mat->Get(AI_MATKEY_UVWSRC(aiTextureType_AMBIENT, 0), ambientCoordinateSet);
+
+        /* Couldn't convince Assimp to import non-zero values from a COLLADA
+         * file, so this code path isn't tested */
+        if(ambientCoordinateSet != 0) {
+            if(!configuration().value<bool>("allowMaterialTextureCoordinateSets")) {
+                Error{} << "Trade::AssimpImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them";
+                return {};
+            }
+
+            flags |= PhongMaterialData::Flag::TextureCoordinateSets;
+        }
+    } else {
         /* Assimp 4.1 forces ambient color to white for STL models. That's just
            plain wrong, so we force it back to black (and emit a warning, so in
            the very rare case when someone would actually want white ambient,
@@ -804,25 +817,68 @@ Containers::Pointer<AbstractMaterialData> AssimpImporter::doMaterial(const Unsig
     /* Key always present, default black */
     aiColor4D diffuseColor;
     UnsignedInt diffuseTexture{};
+    UnsignedInt diffuseCoordinateSet{};
     mat->Get(AI_MATKEY_COLOR_DIFFUSE, diffuseColor);
-    if(flags & PhongMaterialData::Flag::DiffuseTexture)
+    if(flags & PhongMaterialData::Flag::DiffuseTexture) {
         diffuseTexture = firstTextureIndex++;
+        mat->Get(AI_MATKEY_UVWSRC(aiTextureType_DIFFUSE, 0), diffuseCoordinateSet);
+
+        /* Couldn't convince Assimp to import non-zero values from a COLLADA
+         * file, so this code path isn't tested */
+        if(diffuseCoordinateSet != 0) {
+            if(!configuration().value<bool>("allowMaterialTextureCoordinateSets")) {
+                Error{} << "Trade::AssimpImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them";
+                return {};
+            }
+
+            flags |= PhongMaterialData::Flag::TextureCoordinateSets;
+        }
+    }
 
     /* Key always present, default black */
     aiColor4D specularColor;
     UnsignedInt specularTexture{};
+    UnsignedInt specularCoordinateSet{};
     mat->Get(AI_MATKEY_COLOR_SPECULAR, specularColor);
-    if(flags & PhongMaterialData::Flag::SpecularTexture)
+    if(flags & PhongMaterialData::Flag::SpecularTexture) {
         specularTexture = firstTextureIndex++;
+        mat->Get(AI_MATKEY_UVWSRC(aiTextureType_SPECULAR, 0), specularCoordinateSet);
+
+        /* Couldn't convince Assimp to import non-zero values from a COLLADA
+         * file, so this code path isn't tested */
+        if(specularCoordinateSet != 0) {
+            if(!configuration().value<bool>("allowMaterialTextureCoordinateSets")) {
+                Error{} << "Trade::AssimpImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them";
+                return {};
+            }
+
+            flags |= PhongMaterialData::Flag::TextureCoordinateSets;
+        }
+    }
 
     UnsignedInt normalTexture{};
-    if(flags & PhongMaterialData::Flag::NormalTexture)
+    UnsignedInt normalCoordinateSet{};
+    if(flags & PhongMaterialData::Flag::NormalTexture) {
         normalTexture = firstTextureIndex++;
+        mat->Get(AI_MATKEY_UVWSRC(aiTextureType_NORMALS, 0), normalCoordinateSet);
+
+        /* Couldn't convince Assimp to import non-zero values from a COLLADA
+         * file, so this code path isn't tested */
+        if(normalCoordinateSet != 0) {
+            if(!configuration().value<bool>("allowMaterialTextureCoordinateSets")) {
+                Error{} << "Trade::AssimpImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them";
+                return {};
+            }
+
+            flags |= PhongMaterialData::Flag::TextureCoordinateSets;
+        }
+    }
 
     Containers::Pointer<PhongMaterialData> data{Containers::InPlaceInit, flags,
-        Color3{ambientColor}, ambientTexture,
-        Color4{diffuseColor}, diffuseTexture,
-        Color4{specularColor}, specularTexture, normalTexture, Matrix3{},
+        Color3{ambientColor}, ambientTexture, ambientCoordinateSet,
+        Color4{diffuseColor}, diffuseTexture, diffuseCoordinateSet,
+        Color4{specularColor}, specularTexture, specularCoordinateSet,
+        normalTexture, normalCoordinateSet, Matrix3{},
         MaterialAlphaMode::Opaque, 0.5f, shininess, mat};
 
     /* Needs to be explicit on GCC 4.8 and Clang 3.8 so it can properly upcast
--- a/src/MagnumPlugins/AssimpImporter/AssimpImporter.h
+++ b/src/MagnumPlugins/AssimpImporter/AssimpImporter.h
@@ -213,6 +213,9 @@ verbosity levels in each instance.
     @cpp 0xffffff_srgbf @ce, which causes all other color information to be
     discarded. If such a case is detected, the ambient is forced back to
     @cpp 0x000000_srgbf @ce. See also [assimp/assimp#2059](https://github.com/assimp/assimp/issues/2059).
+-   Unless explicitly enabled with the @cb{.ini} allowMaterialTextureCoordinateSets @ce
+    @ref Trade-AssimpImporter-configuration "configuration option", only the
+    first set of texture coordinates is supported.
 
 @subsection Trade-AssimpImporter-limitations-lights Light import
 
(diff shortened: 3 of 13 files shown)


mosra:
> gltf+glb

Just FYI I don't think you need both, just gltf alone is simple (and you can cut that part out of the `material-invalid.gltf` where it checks all cases).

For Assimp I'm afraid a glTF won't work (at least not on the ass-old versions on the CI), but copying and modifying one of  COLLADA files could be easy enough I hope.

src/MagnumPlugins/AssimpImporter/AssimpImporter.conf
```diff
@@ -29,6 +29,9 @@ provides=XglImporter
 # AI_CONFIG_* values, can be changed only before the first file is opened
 ImportColladaIgnoreUpDirection=false
 
+# Import non-default texture coordinate sets
```
mosra, line 32:
```suggestion
# Allow non-default texture coordinate sets. If disabled, materials with 
# non-zero texture coordinate sets (which need explicit support in shaders) 
# will fail to import.
```

Be a bit more helpful here ;)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -786,10 +786,19 @@ Containers::Pointer<AbstractMaterialData> AssimpImporter::doMaterial(const Unsig
...
+    UnsignedInt ambientCoordinateSet{};
     mat->Get(AI_MATKEY_COLOR_AMBIENT, ambientColor);
-    if(flags & PhongMaterialData::Flag::AmbientTexture)
+    if(flags & PhongMaterialData::Flag::AmbientTexture) {
         ambientTexture = firstTextureIndex++;
-    else {
+        mat->Get(AI_MATKEY_UVWSRC(aiTextureType_AMBIENT, 0), ambientCoordinateSet);
+
+        if(!configuration().value<bool>("allowMaterialTextureCoordinateSets") && ambientCoordinateSet != 0) {
+            Error{} << "Trade::AssimpImporter::material(): multiple texture coordinate sets are not supported";
```
mosra, line 796:
```suggestion
            Error{} << "Trade::AssimpImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them";
```

Again, try to be as helpful as possible with each message. Same elsewhere

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -786,10 +786,19 @@ Containers::Pointer<AbstractMaterialData> AssimpImporter::doMaterial(const Unsig
...
+            return {};
+        } else if(ambientCoordinateSet != 0) {
+            flags |= PhongMaterialData::Flag::TextureCoordinateSets;
+        }
```
mosra, line 800: Actually, why not

```suggestion
        if(ambientCoordinateSet != 0) {
            if(!configuration().value<bool>("allowMaterialTextureCoordinateSets")) {
                Error{} << "Trade::AssimpImporter::material():  multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them";
                return {};
            }
            
            flags |= PhongMaterialData::Flag::TextureCoordinateSets;
        }
```

? Seems more straightforward to me. Same elsewhere

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -3,7 +3,7 @@
 
     Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019
               Vladimír Vondruš <mosra@centrum.cz>
-    Copyright © 2017 Jonathan Hale <squareys@googlemail.com>
+    Copyright © 2017, 2020 Jonathan Hale <squareys@googlemail.com>
```
mosra, line 6: :+1: Do that for AssimpImporter.cpp also, please :)

src/MagnumPlugins/AssimpImporter/Test/material-coordinate-sets.dae
```diff
@@ -0,0 +1,154 @@
...
+            <source>diffusenormal_texture-surface</source>
+          </sampler2D>
+        </newparam>
+        <technique sid="common">
+          <phong>
+            <emission>
+              <color sid="emission">0 0 0 1</color>
+            </emission>
+            <!-- COLLADA supports either texture or color but not both, so
+              texture and color is tested with material-color-texture.obj -->
```
mosra, line 40: This comment isn't relevant here.

src/MagnumPlugins/TinyGltfImporter/Test/TinyGltfImporterTest.cpp
```diff
@@ -2517,6 +2519,33 @@ void TinyGltfImporterTest::materialTexCoordFlip() {
     }), TestSuite::Compare::Container);
 }
 
+void TinyGltfImporterTest::materialTextureCoordinateSets() {
+    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("TinyGltfImporter");
+    importer->configuration().setValue("allowMaterialTextureCoordinateSets", true);
```
mosra, line 2524: I would make two variants of this test, the other testing the default behavior, that the import fails for each of the materials.

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
```diff
@@ -1400,13 +1400,15 @@ std::string TinyGltfImporter::doMaterialName(const UnsignedInt id) {
...
         return false;
     }
 
-    if(texCoord != 0) {
+    if(configuration().value<bool>("allowMaterialTextureCoordinateSets")) {
+        coordinateSet = texCoord;
+    } else if(texCoord != 0) {
         Error{} << "Trade::TinyGltfImporter::material(): multiple texture coordinate sets are not supported";
         return false;
     }
```
mosra, line 1414:
```suggestion
    if(texCoord != 0) {
        if(!configuration().value<bool>("allowMaterialTextureCoordinateSets")) {
            Error{} << "Trade::TinyGltfImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them";
            return false;
        }
        
        coordinateSet = texCoord;
    }
```

Basically the same as in the case of Assimp.

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.conf
```diff
@@ -34,4 +34,7 @@ textureCoordinateYFlipInMaterial=false
 # The non-standard MeshAttribute::ObjectId is by default recognized under this
 # name. Change if your file uses a different identifier.
 objectIdAttribute=_OBJECT_ID
+
+# Import non-default texture coordinate sets
```
mosra, line 38:
```suggestion
# Allow non-default texture coordinate sets. If disabled, materials with 
# non-zero texture coordinate sets (which need explicit support in shaders) 
# will fail to import.
```

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -625,6 +627,18 @@ void AssimpImporterTest::materialMultipleTextures() {
...
+    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("AssimpImporter");
+    CORRADE_VERIFY(importer->openFile(Utility::Directory::join(ASSIMPIMPORTER_TEST_DIR, "material-coordinate-sets.dae")));
+
+    auto& mat = static_cast<PhongMaterialData&>(*importer->material(0));
+
+    // Assimp ignores ambient texture: CORRADE_COMPARE(mat.ambientCoordinateSet(), 1);
+    CORRADE_COMPARE(mat.diffuseCoordinateSet(), 2);
+    CORRADE_COMPARE(mat.ambientCoordinateSet(), 3);
+    CORRADE_COMPARE(mat.normalCoordinateSet(), 2);
+}
```
mosra, line 640: Same as in case of the TinyGltf test -- ideally two variants, testing that the default behavior correctly fails. This also means you need to have ~~three~~ four different materials in order to verify each of the diffuse/specular/ambient/normal failures is done correctly. Instanced tests might (or might not) be helpful.

(What about a specular texture here?)

src/MagnumPlugins/AssimpImporter/AssimpImporter.conf
Squareys: Nice, yes 👍

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
Squareys: I actually solved this before reading your comment... the second ambient should have been specular (copy pasta), and the failure is now also tested, but not separate. Maybe check really quick if the new version satisfies your requirements :)

Squareys:
@mosra FYI, this is ready and I'm currently waiting for feedback on this :)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -804,25 +815,63 @@ Containers::Pointer<AbstractMaterialData> AssimpImporter::doMaterial(const Unsig
     /* Key always present, default black */
     aiColor4D diffuseColor;
     UnsignedInt diffuseTexture{};
+    UnsignedInt diffuseCoordinateSet{};
     mat->Get(AI_MATKEY_COLOR_DIFFUSE, diffuseColor);
-    if(flags & PhongMaterialData::Flag::DiffuseTexture)
+    Debug() << "Hello Material";
```
mosra, line 820: Hello Debug! :wave:

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -625,6 +627,38 @@ void AssimpImporterTest::materialMultipleTextures() {
...
+        CORRADE_COMPARE(out.str(), "Trade::AssimpImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them\n");
+    }
+
+    importer->configuration().setValue("allowMaterialTextureCoordinateSets", true);
+
+    CORRADE_VERIFY(importer->openFile(Utility::Directory::join(ASSIMPIMPORTER_TEST_DIR, "material-coordinate-sets.dae")));
+    auto abstractMat = importer->material(0);
+    auto& mat = static_cast<PhongMaterialData&>(*abstractMat);
+
+    CORRADE_EXPECT_FAIL("Assimp ignores ambient texture and does not import coordinate sets");
```
mosra, line 649: Just this `CORRADE_EXPECT_FAIL` is enough here -- it works until the end of the scope, so the following ones are unnecessary.

src/MagnumPlugins/AssimpImporter/AssimpImporter.h
```diff
@@ -213,6 +213,9 @@ verbosity levels in each instance.
     @cpp 0xffffff_srgbf @ce, which causes all other color information to be
     discarded. If such a case is detected, the ambient is forced back to
     @cpp 0x000000_srgbf @ce. See also [assimp/assimp#2059](https://github.com/assimp/assimp/issues/2059).
+-   Unless explicitly enabled by setting the configuration option
+    "allowMaterialTextureCoordinateSets" to @cpp true @ce, only the first set
+    of texture coordinates is supported. See details @ref Trade-AssimpImporter-configuration "below".
```
mosra, line 218:
```suggestion
-   Unless explicitly enabled with the @cb{.ini} allowMaterialTextureCoordinateSets @ce 
    @ref Trade-AssimpImporter-configuration "configuration option", only the 
    first set of texture coordinates is supported.
```

Links that are just `here`, `click here`, `see below` are an antipattern and hurt readability ;)

src/MagnumPlugins/TinyGltfImporter/Test/TinyGltfImporterTest.cpp
```diff
@@ -2517,6 +2519,56 @@ void TinyGltfImporterTest::materialTexCoordFlip() {
...
+        CORRADE_VERIFY(!importer->material(0));
+        CORRADE_VERIFY(!importer->material(1));
+        CORRADE_VERIFY(!importer->material(2));
+        CORRADE_VERIFY(!importer->material(3));
+        CORRADE_COMPARE(out.str(),
+            "Trade::TinyGltfImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them\n"
+            "Trade::TinyGltfImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them\n"
+            "Trade::TinyGltfImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them\n"
+            "Trade::TinyGltfImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them\n");
+    }
```
mosra, line 2539: I would split the first part into a separate test, `materialTextureCoordinateSetsDefault()`

src/MagnumPlugins/TinyGltfImporter/Test/TinyGltfImporterTest.cpp
```diff
@@ -286,13 +287,13 @@ constexpr struct {
...
     {"invalid texture index pbrSpecularGlossiness diffuse", "diffuseTexture index 2 out of bounds for 2 textures"},
     {"invalid texture index pbrSpecularGlossiness specular", "specularGlossinessTexture index 2 out of bounds for 2 textures"},
-    {"invalid texCoord index pbrMetallicRoughness", "multiple texture coordinate sets are not supported"},
-    {"invalid texCoord index pbrSpecularGlossiness diffuse", "multiple texture coordinate sets are not supported"},
-    {"invalid texCoord index pbrSpecularGlossiness specular", "multiple texture coordinate sets are not supported"},
+    {"invalid texCoord index pbrMetallicRoughness", "multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them"},
+    {"invalid texCoord index pbrSpecularGlossiness diffuse", "multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them"},
+    {"invalid texCoord index pbrSpecularGlossiness specular", "multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them"},
```
mosra, line 296: Since those are now handled in a dedicated test case, remove all four from there (and also from the `material-invalid.gltf` file)

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -625,6 +627,38 @@ void AssimpImporterTest::materialMultipleTextures() {
...
+    {
+        std::ostringstream out;
+        Error redirectError{&out};
+        CORRADE_VERIFY(importer->openFile(Utility::Directory::join(ASSIMPIMPORTER_TEST_DIR, "material-coordinate-sets.dae")));
+
+        CORRADE_EXPECT_FAIL("Material is loaded since coordinate sets are not imported");
+        CORRADE_VERIFY(!importer->material(0));
+
+        CORRADE_COMPARE(out.str(), "Trade::AssimpImporter::material(): multiple texture coordinate sets are not allowed by default, enable allowMaterialTextureCoordinateSets to import them\n");
+    }
```
mosra, line 641: As with TinyGltf, I would split the first part into a separate test, `materialTextureCoordinateSetsDefault()`

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -804,25 +815,63 @@ Containers::Pointer<AbstractMaterialData> AssimpImporter::doMaterial(const Unsig
...
+    if(flags & PhongMaterialData::Flag::DiffuseTexture) {
         diffuseTexture = firstTextureIndex++;
+        mat->Get(AI_MATKEY_UVWSRC(aiTextureType_DIFFUSE, 0), diffuseCoordinateSet);
+
+        if(diffuseCoordinateSet != 0) {
```
mosra, line 825:
```suggestion
        /* Couldn't convince Assimp to import non-zero values from a COLLADA 
           file, so this code path isn't tested */
        if(diffuseCoordinateSet != 0) {
```

Add an explanatory comment so we know *why* this isn't covered. Same in the two cases below.

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.h
```diff
@@ -269,7 +269,10 @@ fail.
 -   Subset of all material specs is currently imported as @ref PhongMaterialData,
     including normal textures
 -   Ambient color is always @cpp 0x000000_rgbf @ce (never a texture)
--   Only the first set of texture coordinates is supported
+-   Unless explicitly enabled by setting the configuration option
+    "allowMaterialTextureCoordinateSets"
+    to @cpp true @ce, only the first set of texture coordinates is supported. See details
+    @ref Trade-TinyGltfImporter-configuration "below".
```
mosra, line 275:
```suggestion
-   Unless explicitly enabled with the @cb{.ini} allowMaterialTextureCoordinateSets @ce 
    @ref Trade-TinyGltfImporter-configuration "configuration option", only the 
    first set of texture coordinates is supported.
```

Basically same as in the case of Assimp.

mosra:
Thx!

PR #84, opened 2020-05-08
TinyGltfImporter: Import Weights and JointIds mesh attributes

Squareys:
Hi @mosra !

Building on top of https://github.com/mosra/magnum/pull/441, here's the import of the new mesh attributes in TinyGltfImporter.

Best,
Jonathan

**TODO**
 - [ ] Docs: Advertise support
 - [ ] Tests
 - [ ] What are `"skins"`? 
 - [ ] Is this sufficient to import skinned mesh animations?

### DIFF
--- a/src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
+++ b/src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
@@ -1171,6 +1171,41 @@ Containers::Optional<MeshData> TinyGltfImporter::doMesh(const UnsignedInt id, Un
                 return Containers::NullOpt;
             }
 
+        /* Joint attribute ends with _0, _1 ... */
+        } else if(Utility::String::beginsWith(attribute.first, "JOINTS")) {
+            name = MeshAttribute::JointIds;
+
+            if(accessor.type != TINYGLTF_TYPE_VEC4) {
+                Error{} << "Trade::TinyGltfImporter::mesh(): unexpected JOINTS type" << accessor.type;
+                return Containers::NullOpt;
+            }
+
+            if(!(accessor.componentType == TINYGLTF_COMPONENT_TYPE_UNSIGNED_BYTE) &&
+               !(accessor.componentType == TINYGLTF_COMPONENT_TYPE_UNSIGNED_SHORT)) {
+                Error{} << "Trade::TinyGltfImporter::mesh(): unsupported JOINTS component type"
+                    << (accessor.normalized ? "normalized" : "unnormalized")
+                    << accessor.componentType;
+                return Containers::NullOpt;
+            }
+
+        /* Weights attribute ends with _0, _1 ... */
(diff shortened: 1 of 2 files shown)


Squareys:
From gitter to archive here:

> It appears that in addition to the Skin Data, we will need to import some form of "Skin" or "Skeleton", which requires extension of the Trade:: Apis.
I.e. this could be done with SkeletonData / SkinData or a Trade::SkinnedMeshObjectData
There are inverseBindMatrices that are stored per Bone in there (Some Matrix accessor in gltf), a reference to the root node and the nodes that are used as bones/joints
Since ObjectType::Light will redirect to LightData, I assume we'd have ObjectType::SkinnedMesh and redirect to skin, mesh and possibly nodes of the skeleton

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
```diff
@@ -1171,6 +1171,41 @@ Containers::Optional<MeshData> TinyGltfImporter::doMesh(const UnsignedInt id, Un
                 return Containers::NullOpt;
             }
 
+        /* Joint attribute ends with _0, _1 ... */
+        } else if(Utility::String::beginsWith(attribute.first, "JOINTS")) {
```
mosra, line 1175: TODO for testing: the code relies on the attributes being sorted (so JOINTS_0 before JOINTS_1) and without gaps. I think sorting is ensured implicitly due to how JSON dict data is stored, but the gaps need to be checked.

src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.cpp
```diff
@@ -1171,6 +1171,41 @@ Containers::Optional<MeshData> TinyGltfImporter::doMesh(const UnsignedInt id, Un
...
+            name = MeshAttribute::JointIds;
+
+            if(accessor.type != TINYGLTF_TYPE_VEC4) {
+                Error{} << "Trade::TinyGltfImporter::mesh(): unexpected JOINTS type" << accessor.type;
+                return Containers::NullOpt;
+            }
+
+            if(!(accessor.componentType == TINYGLTF_COMPONENT_TYPE_UNSIGNED_BYTE) &&
+               !(accessor.componentType == TINYGLTF_COMPONENT_TYPE_UNSIGNED_SHORT)) {
```
mosra, line 1184:
```suggestion
            if(!(accessor.componentType == TINYGLTF_COMPONENT_TYPE_UNSIGNED_BYTE && !accessor.normalized) &&
               !(accessor.componentType == TINYGLTF_COMPONENT_TYPE_UNSIGNED_SHORT && !accessor.normalized)) {
```

IDs can't be normalized. Also needs a test ;)

mosra:
Huh, what's this, completely forgot this was even attempted in the old plugin :laughing: Two complete plugin reimplementations later, this is done in `GltfImporter` in cf9c300a3063c832c31ed7a72f5a805c2b93c358.

PR #94, opened 2020-12-17
[StanfordImporter] Support vertex_index for face list

jlaxson:
Assimp (at least) exports ply files with faces as `property list uchar int vertex_index` whereas the code here supports only `vertex_indices` as a name (assimp parses both).  The ostensible [PLY reference](http://paulbourke.net/dataformats/ply/) seems to conflate them, implying to me it was silently changed at some point.  This change adds support for both.


mosra:
Oh! Thank you. Didn't run into this when verifying with my own model zoo :)

Will adapt the tests a bit and merge.

mosra:
Thank you, merged! Took a while as I had to solve the sudden clang-cl build failure as well, everything should be green on `master` again.

jlaxson:
👍

PR #97, opened 2021-06-18
AssimpImporter: import animations

pezcode:
This PR adds support for importing node animations with `AssimpImporter`.

`doAnimation` imports translation vectors, rotation quaternions and scale vectors in separate tracks for each animation channel. Interpolation is assumed to always be linear as Assimp doesn't support any other interpolation types, and it's impossible to get the source data.

I reused the `TinyGltfImporter` config values to perform the same sanity fixups. While *some* Assimp importers perform these themselves (e.g. the FBX importer makes sure to get the shortest rotation path), not all of them do.

What's missing:
- [x] tests - a majority can be reused from TinyGltfImporter
- [x] check against older Assimp versions as well as latest master
- [x] ~~there's a bug with uninitialized quaternions when using `magnum-player`, have to investigate~~ see comments about spline-interpolated glTF
- [x] mention animation support in the docs

---

Commissioned by [Wonderland](https://wonderlandengine.com).


pezcode:
To expand a bit on the interpolation issues:

In the case of glTF cubic spline animations, Assimp simply takes the spline knots. While this is not *terribly* wrong, it's still wrong, but there is no way to get the interpolation function out of Assimp. Same thing applies to step interpolation, although that's probably less relevant.

There's also a nasty bug that only [got fixed](https://github.com/assimp/assimp/commit/e3083c21f0a7beae6c37a2265b7919a02cbf83c4) less than a year ago (so way after the latest tagged release 5.0.1): glTF spline tangents are read as key frame values instead of skipping them, producing unusable garbage and, worst of all, leaving no way to reliably detect that we got garbage.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -341,6 +350,14 @@ void AssimpImporter::doOpenData(const Containers::ArrayView<const char> data) {
 
     CORRADE_INTERNAL_ASSERT(_f->scene);
 
+    /* Get name of importer. Useful for workarounds based on importer/file type. */
+    _f->importerName = "unknown";
+    const int importerIndex = _importer->GetPropertyInteger("importerIndex", -1);
```
mosra, line 355: Oh, this is very useful, good to know. I need this exact thing to implement Assimp "PBR" material support in #91, where a "unknown" texture in one format is a material/roughness texture in glTF and there's no other way to differentiate the two than checking the file format :laughing:

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1118,6 +1135,235 @@ Containers::Optional<ImageData2D> AssimpImporter::doImage2D(const UnsignedInt id
...
+                return Containers::NullOpt;
+            }
+
+            /* Assimp only supports linear interpolation. For glTF splines
+               it simply uses the spline control points. */
+            /** @todo Before https://github.com/assimp/assimp/commit/e3083c21f0a7beae6c37a2265b7919a02cbf83c4
+             *        Assimp incorrectly read the spline tangents as values.
+             *        Is there a way to detect this so we can work around it,
+             *        or at least emit a warning?
+             */
```
mosra, line 1243: Sigh :) When all sane attempts failed, this is how I was detecting what library version I was linking to, by checking for a presence of an API that got added in given version: https://github.com/mosra/magnum-plugins/blob/ae1278de39a2b8b8f46dd590d39a15d0b410f18e/src/MagnumPlugins/AssimpImporter/Test/CMakeLists.txt#L46-L52

An even more insane variant of this is for SPIRV-Tools, which for some reason don't provide *any* version info whatsoever, so I have to detect each API addition with that: https://github.com/mosra/magnum-plugins/blob/ae1278de39a2b8b8f46dd590d39a15d0b410f18e/src/MagnumPlugins/SpirvToolsShaderConverter/CMakeLists.txt#L68-L88

- - -

Looking around, https://github.com/assimp/assimp/pull/3510 seems to be a PR that got merged around the same time as the above commit (few days earlier), so i think you could ignore the few commits in between and assume that if `aiScene` has the name member, it should also have non-fucked-up tangents? (I didn't really look further, maybe there's a better candidate tho.) The `CheckAssimpVersion5.cpp` file is currently in the test directory, I think it would make sense to move it one level up next to the plugin and add this other check to it similarly as with SPIRV-Tools.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1118,6 +1135,235 @@ Containers::Optional<ImageData2D> AssimpImporter::doImage2D(const UnsignedInt id
...
+    std::size_t dataOffset = 0;
+    std::size_t trackId = 0;
+    Containers::Array<Trade::AnimationTrackData> tracks{trackCount};
+    for(std::size_t a = animationBegin; a != animationEnd; ++a) {
+        const aiAnimation* animation = _f->scene->mAnimations[a];
+        for(std::size_t c = 0; c != animation->mNumChannels; ++c) {
+            const aiNodeAnim* channel = animation->mChannels[c];
+            const UnsignedInt target = doObject3DForName(channel->mNodeName.C_Str());
+            if(target == -1) {
+                Error{} << "Trade::AssimpImporter::animation(): unknown target node " << channel->mNodeName.C_Str();
```
mosra, line 1233: Minor thing (a test would catch this, I suppose) -- the Debug classes implicitly add spaces between printed values, so

```suggestion
                Error{} << "Trade::AssimpImporter::animation(): unknown target node" << channel->mNodeName.C_Str();
```

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1118,6 +1135,235 @@ Containers::Optional<ImageData2D> AssimpImporter::doImage2D(const UnsignedInt id
...
+
+            /* Key times are given as ticks, convert to milliseconds. Default value of 25 is
+               taken from the assimp_view tool. */
+            double ticksPerSecond = animation->mTicksPerSecond != 0.0 ? animation->mTicksPerSecond : 25.0;
+
+            /* For glTF files mTicksPerSecond is completely useless before
+               https://github.com/assimp/assimp/commit/7a8b7ba88d6d84dde7fd43419ac2b022c9887856
+               but can be assumed to always be 1000. */
+            /** @todo Check if this is broken in other importers, too */
+            const bool isGltf = std::strcmp(_f->importerName, "glTF2 Importer") == 0;
```
mosra, line 1258: FYI, the [StringView](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1BasicStringView.html) class should be able to do this in a less error-prone but similarly efficient way:

```cpp
using namespace Containers::Literals;

const bool isGltf = _f->importerName == "glTF2 Importer"_s;
```

String views will eventually replace std::string everywhere anyway, so including it just for this one comparison is fine right now.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1118,6 +1135,235 @@ Containers::Optional<ImageData2D> AssimpImporter::doImage2D(const UnsignedInt id
...
+            if(isGltf) ticksPerSecond = 1000.0;
```
mosra, line 1259: ~~For shits and giggles~~ To inform the user, I would print a warning pointing out to assimp's brokeness:

```cpp
if(isGltf && !Math::equals(animation->mTicksPerSecond, 1000.0)) {
    Warning{} << "Trade::AssimpImporter::animation():" << animation->mTicksPerSecond << "ticks per second is incorrect for glTF, patching to 1000.0";
    ticksPerSecond = 1000.0;
}
```

At first I thought this could be only a verbose warning (`if(flags() & ImporterFlag::Verbose)`) but since it seems to be fixed in master and is a pretty important fix, I think it makes sense to print it always.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1118,6 +1135,235 @@ Containers::Optional<ImageData2D> AssimpImporter::doImage2D(const UnsignedInt id
...
+    return _f->scene->mAnimations[id]->mName.C_Str();
+}
+
+namespace {
+
+Animation::Extrapolation extrapolationFor(aiAnimBehaviour behaviour) {
+    switch (behaviour) {
+        case aiAnimBehaviour_DEFAULT:
+            /** @todo emulate this correctly by inserting keyframes
+                with the default node transformation */
```
mosra, line 1173: Hmm, or the Animation lib could support this directly? 

IIRC I was considering this option but it felt kinda useless to me (couldn't think of a practical use case, since "default transformation" is kinda arbitrary and can mean anything) so I didn't add it.

mosra (commented):
Thanks for this! 

Some ideas that could hopefully unblock the glTF import disaster :)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: First things first: Thanks for the feedback so far. Always a pleasure getting great, detailed technical (and friendly!) feedback from you 🤗

On to the topic: The problem I see here is that even if we find out the Assimp version, I don't see a way (or maybe I missed it) to figure out if the animation data is taken from cubic splines in the first place. The only thing Assimp gives us is the keys and values, but no info about the interpolation function. A **terrible** workaround would be to manually parse the GLTF enough to find out if cubic spline interpolation is used, and then at least emit a warning.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: Done. Will push when I finished the tests.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: Done 🙇

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: I suppose you could add the concept of "border" values for extrapolation. For this particular thing it seems a bit useless, since the default node translation/rotation/etc. is well-defined and the model *should* have a key-frame at 0.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: Good catch. I tripped over this a few times before, probably won't be the last time 😄

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
mosra: Border values sound useful, but from a technical perspective the more flexible the extrapolation gets, the slower the animation processing will be due to all the special-casing... so if it's doable, I think adding extra keyframes at the beginning/end is the better solution :)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
mosra: > I don't see a way to figure out if the animation data is taken from cubic splines in the first place.

Ahah, damn, I see. This sucks. So, if I understand correctly, on older versions this will give back a total nonsense, and on newer versions it'll ignore the tangent part of the cubic splines, right? I would definitely mention this in the limitations/caveats section of the docs.

I'd say in older versions you could grep the input file for a presence of `"CUBICSPLINE"` and then emit a warning in each `doAnimation()` if that's the case. I.e., doing something like this in `doOpenData()`:

```cpp
if(_f->importerIsGltf)
    _f->animationsProbablyBroken = Containers::StringView{data}.contains("\"CUBICSPLINE\""_s);
```

It's the laziest possible solution but should do the trick. It's probably also quite slow for large files, but the bar set by Assimp wasn't really high either. We're not here to fix the world, anyway.

In practice, I'm telling everyone to never use AssimpImporter for glTF because the material, light and mesh import sucks as well, so *optimistically* the only place this would affect is the tests, where you'll get various levels of bullshit from the TinyGltfImporter test files. And *hopefully* FBX animation support is not this bad, as that's what this importer is mainly for :)

pezcode: I went ahead and implemented the check, but then realized that `doOpenFile` fills the assimp scene data and then calls `doOpenData({})`, so for files this doesn't work. I don't feel like going full clown mode and trying to intercept this (custom IO callback? don't even know if this works) is worth it in the slightest, but I'm open to other suggestions.

mosra: Sigh, again. Then ... I'd say print the warning on older versions always? It would mean some false positives, but better than being completely silent while giving out bad data.

There already is a callback implemented, for Magnum file callbacks, but I don't feel the extra complexity is worth it. Just warn always.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: Would you be OK with leaving the todo in there for now and postponing this to a later fix? I haven't found a real-world file that even relies on this behaviour, and it complicates the import code and tests quite a bit.

mosra: Absolutely :) A TODO is fine.

src/MagnumPlugins/AssimpImporter/CheckAssimpVersion.cpp
```diff
@@ -0,0 +1,64 @@
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019,
+                2020, 2021 Vladimír Vondruš <mosra@centrum.cz>
```
mosra, line 5: Add yourself here, to all touched files :)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -359,22 +382,22 @@ void AssimpImporter::doOpenData(const Containers::ArrayView<const char> data) {
         /* Store first possible texture index for this material, next textures
            use successive indices. */
         _f->textureIndices[mat] = textureIndex;
-        for(std::size_t i = 0; i != mat->mNumProperties; ++i) {
+        for(std::size_t j = 0; j != mat->mNumProperties; ++j) {
```
mosra, line 385: Huh. I wonder what kind of bugs this caused :O

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -448,6 +471,18 @@ void AssimpImporter::doOpenData(const Containers::ArrayView<const char> data) {
...
+    /* Before https://github.com/assimp/assimp/commit/e3083c21f0a7beae6c37a2265b7919a02cbf83c4
+       Assimp incorrectly read spline tangents as values in glTF animation tracks.
+       Quick and dirty check to see if we're dealing with a possibly affected file
+       and Assimp version. This might produce false-positives on files without
+       spline-interpolated animations, but for doOpenState and doOpenFile we 
+       have no access to the file content to check if the file contains "CUBICSPLINE". */
+    if (_f->scene->HasAnimations() && _f->importerIsGltf && ASSIMP_VERSION < 20201123) {
+        Warning{} << "Trade::AssimpImporter::open(): spline-interpolated animations imported" <<
+            "from this file are most likely broken using this version of Assimp. Consult the" <<
+            "importer documentation for more information.";
```
mosra, line 484:
```suggestion
        Warning{} << "Trade::AssimpImporter::openData(): spline-interpolated animations imported "
            "from this file are most likely broken using this version of Assimp. Consult the "
            "importer documentation for more information.";
```

(Using a single string literal instead.)

src/MagnumPlugins/AssimpImporter/AssimpImporter.h
```diff
@@ -197,13 +197,47 @@ with @ref InputFileCallbackPolicy::LoadTemporary and
...
+    the @cb{.ini} mergeAnimationClips @ce option in order to for example
+    preserve cinematic animations when using the Blender glTF exporter (as it
+    otherwise outputs a separate clip for each object). When this option is
+    enabled, @ref animationCount() always report either @cpp 0 @ce or
+    @cpp 1 @ce and the merged animation has no name. With this option enabled,
+    however, it can happen that multiple conflicting tracks affecting the same
+    node are merged in the same clip, causing the animation to misbehave.
+-   Assimp versions before commit [e3083c21f0a7beae6c37a2265b7919a02cbf83c4](https://github.com/assimp/assimp/commit/e3083c21f0a7beae6c37a2265b7919a02cbf83c4)
+    read spline-interpolated glTF animation tracks incorrectly and produce broken
+    animations. If this is detected, the importer produces a warning.
```
mosra, line 239:
```suggestion
    animations. A warning is printed when importing glTF animations on these versions,
    however because it's impossible to detect the actual brokeness, the warning is 
    printed even if the imported data may be correct.
```

Rephrase as you see fit, apparently I'm unable to form a clear sentence today :sweat_smile:

src/MagnumPlugins/AssimpImporter/AssimpImporter.h
```diff
@@ -307,6 +341,9 @@ enum. Some of them are enabled by default, some not; options for not yet
 supported features are omitted. These are passed to Assimp when opening a file,
 meaning a change in these will be always applied to the next opened file.
 
+Additional options exist to configure animation import. See
+below for all options and their default values.
```
mosra, line 345: Since the options are already referenced from the "Animation import" section above, I don't think this paragraph is needed.

mosra (commented):
Got just a few minutes today, so only a bunch of random comments :)

Looks great so far! :+1:

pezcode:
> Got just a few minutes today, so only a bunch of random comments :)
> 
> Looks great so far! 👍

Thanks for taking the time to check in even if you're busy, appreciate it. I adressed your recent feedback locally, I'll push it along with all the tests probably tomorrow. Assimp is still abusing me with a funny bug, gotta fix that first.

src/MagnumPlugins/AssimpImporter/AssimpImporter.h
pezcode: Done

src/MagnumPlugins/AssimpImporter/AssimpImporter.h
pezcode: Done

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: Done

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: Probably nothing, since it only shadowed `i` and the original `i` wasn't being used from what I can tell. I forgot to update the comments in the loop, will fix that in another commit.

mosra: > the original i wasn't being used from what I can tell.

Ah! I thought the inner loop was incrementing the original variable (which would be crazy), missed the redeclaration. :)

pezcode:
Here are the tests 🐌 A few related notes:
- `TinyGltfImporterTest` files and cases are re-used where possible. The animation*.gl* needed changes, because Assimp doesn't import animations from glTF files that don't have a scene (there's also a test for that behavior). So I copied them over and modified them in this importer test directory.
- Re-use doesn't work for the spline-interpolated data (see discussion above), so I simply test that the data makes it there as linearly interpolated animations.
- Trying to convert the existing animated glTF files into FBX/Collada (.dae) made me lose so much hair I decided to create a simple test scene in Blender, add key-frames and export them directly. This kind of worked but not well enough to be able to add tests of the same fidelity as the glTF ones. Every exporter seems to have a different way to bake and merge samples, so the outputs are wildly different. In `animation()` (as opposed to `animationGltf()`), these exported files are loaded and then some sanity checks are performed. It's probably possible to extend this to check certain key-frames but it will be painful, see the `@todo` comment at the end of it.
- There is no test for `Animation::Extrapolation` beyond expecting `Constant` in tests. There is currently no code in Assimp that sets this except for the .irr importer, and I couldn't even find a sample file using animations, so...

Some general comments on format support:
- glTF is **poisoned**. I had to remove the invalid animation test, because out of bounds indices in the .gltf ended up in wild asserts in the underlying json library, so there's no way to even get useful error output. There is also a mind-blowing bug that reads beyond key/value buffers and you end up with extraneous, bogus data and there is no way to detect this. I would almost go as far as putting a big red warning on the importer docs à la _**"DON'T EVEN TRY TO USE THIS WITH GLTF"**_.
- FBX seems to work fine, but I couldn't for the life of me figure out how to export files correctly for better tests. Either Blender exported no animations, or it exported all animations targetting _all_(??) objects. I tested a whole slew of matching FBX/glTF files off [Sketchfab](https://sketchfab.com) with [`magnum-player`](https://magnum.graphics/showcase/player/) and they all seemed to work. The ones that were broken were still broken with `TinyGltfImporter`.
- Collada works judging by the test files and a few random scenes off the internet, but I didn't put *that* much effort into checking

I wanted to test with older versions, but various vcpkg shenanigans threw a wrench in my plans. Not sure how much of this is tested on CI, but I could give this another try with WSL (or as a CMake subdirectory - is that supported for Assimp?). So for now, consider the test extent to be 5.0.1 and master.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1118,6 +1155,312 @@ Containers::Optional<ImageData2D> AssimpImporter::doImage2D(const UnsignedInt id
...
+            const aiNodeAnim* channel = animation->mChannels[c];
+
+            UnsignedInt translationKeyCount = channel->mNumPositionKeys;
+            UnsignedInt rotationKeyCount = channel->mNumRotationKeys;
+            UnsignedInt scalingKeyCount = channel->mNumScalingKeys;
+
+            /* Assimp adds useless 1-key tracks with default node
+               translation/rotation/scale if the channel doesn't
+               target all 3 of them. Ignore those. */
+            if(translationKeyCount == 1 || rotationKeyCount == 1 || scalingKeyCount == 1) {
```
pezcode, line 1249: Should this be hidden behind a config value? Maybe something generic like `removeEmptyAnimations` that would also remove completely empty tracks (that would break the glTF test, but I don't see a good real-world use-case for that anyway).

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1118,6 +1155,312 @@ Containers::Optional<ImageData2D> AssimpImporter::doImage2D(const UnsignedInt id
...
+    for(std::size_t a = animationBegin; a != animationEnd; ++a) {
+        channelCount += _f->scene->mAnimations[a]->mNumChannels;
+    }
+
+    /* Presence of translation/rotation/scaling keys for each channel.
+       We might skip certain tracks (see comments in the next loop)
+       but we need to know the correct track count and data array size
+       up front, so determine this and remember it for the actual
+       loop that extracts the tracks. */
+    Containers::Array<std::tuple<bool, bool, bool>> channelTrackPresence{channelCount};
```
pezcode, line 1230: Is there a better way to do this? I tried `EnumSet` but `AnimationTrackTargetType` entries aren't binary exclusive.

src/MagnumPlugins/AssimpImporter/Test/CMakeLists.txt
```diff
@@ -87,7 +81,12 @@ corrade_add_test(AssimpImporterTest AssimpImporterTest.cpp
         y-up.dae
         z-up.dae)
 target_link_libraries(AssimpImporterTest PRIVATE Assimp::Assimp)
-target_include_directories(AssimpImporterTest PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>)
+target_include_directories(AssimpImporterTest PRIVATE
+    ${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>
+    # The test needs access to configureInternal.h written by
+    # AssimpImporter. The dynamic library doesn't get linked to
+    # and hence doesn't get the binary dir in the include dirs.
+    ${PROJECT_BINARY_DIR}/src)
```
pezcode, line 89: Not a huge fan of this but I didn't see a better way to let the tests access headers in the binary dir of `AssimpImporter`

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -287,6 +349,690 @@ void AssimpImporterTest::openDataFailed() {
...
+        return true;
+
+    /* splitExtension returns empty extensions for files starting with a dot */
+    const std::string extension = Utility::Directory::splitExtension("x" + fileName).second;
+    const bool isGltf = extension == ".gltf" || extension == ".glb";
+    if(isGltf) {
+        return false;
+    } else {
+        const unsigned int version = aiGetVersionMajor()*100 + aiGetVersionMinor();
+        CORRADE_ASSERT(extension == ".dae" || extension == ".fbx", "Unsupported file type", false);
```
mosra, line 368: Could be just this, asserts with messages are mainly targeted at library users so they get a human readable description of what they did wrong. Not the case here :)

```suggestion
        CORRADE_INTERNAL_ASSERT(extension == ".dae" || extension == ".fbx");
```

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -287,6 +349,690 @@ void AssimpImporterTest::openDataFailed() {
...
+
+void AssimpImporterTest::animationGltfNoScene() {
+    /* Assimp refuses to import glTF animations if the file
+       has no scenes. This reuses the TinyGltfImporter test
+       files, not the corrected ones used by other tests. */
+    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("AssimpImporter");
+    CORRADE_VERIFY(importer->openFile(Utility::Directory::join(TINYGLTFIMPORTER_TEST_DIR,
+        "animation.gltf")));
+
+    CORRADE_COMPARE(importer->animationCount(), 0);
```
mosra, line 589: I'd do this instead so it's clear that this is a known importer failure (and when Assimp eventually fixes itself, the test will say "unexpected pass" instead of "fail":

```cpp
    /* This reuses the TinyGltfImporter test files, not the corrected ones used by other tests. */
    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("AssimpImporter");
    CORRADE_VERIFY(importer->openFile(Utility::Directory::join(TINYGLTFIMPORTER_TEST_DIR,
        "animation.gltf")));

    CORRADE_EXPECT_FAIL("Assimp refuses to import glTF animations if the file has no scenes.");
    CORRADE_COMPARE(importer->animationCount(), 3);
```

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -287,6 +349,690 @@ void AssimpImporterTest::openDataFailed() {
...
+    CORRADE_VERIFY(importer->openFile(Utility::Directory::join(ASSIMPIMPORTER_TEST_DIR,
+        "animation.gltf")));
+
+    std::ostringstream out;
+    {
+        Warning redirectWarning{&out};
+        CORRADE_VERIFY(importer->animation(1));
+    }
+    CORRADE_VERIFY(out.str().find(" ticks per second is incorrect for glTF, patching to 1000\n")
+        != std::string::npos);
```
mosra, line 614:
```suggestion
    CORRADE_VERIFY(Containers::StringView{out.str()}.contains(" ticks per second is incorrect for glTF, patching to 1000\n"));
```

Probably needs a `#include <Corrade/Containers/StringStl.h>`. (And I should finally add "string contains" comparators for this kind of tests.)

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -287,6 +349,690 @@ void AssimpImporterTest::openDataFailed() {
...
+    CORRADE_COMPARE(animation->trackType(0), AnimationTrackType::Quaternion);
+    CORRADE_COMPARE(animation->trackResultType(0), AnimationTrackType::Quaternion);
+    CORRADE_COMPARE(animation->trackTargetType(0), AnimationTrackTargetType::Rotation3D);
+    CORRADE_COMPARE(animation->trackTarget(0), 3);
+    Animation::TrackView<const Float, const Quaternion> rotation = animation->track<Quaternion>(0);
+    CORRADE_COMPARE(rotation.interpolation(), Animation::Interpolation::Linear);
+    CORRADE_COMPARE(rotation.before(), Animation::Extrapolation::Constant);
+    CORRADE_COMPARE(rotation.after(), Animation::Extrapolation::Constant);
+    CORRADE_COMPARE_AS(rotation.keys(), (Containers::StridedArrayView1D<const Float>{keys}), TestSuite::Compare::Container);
+    if(!ASSIMP_HAS_BROKEN_GLTF_SPLINES) {
```
mosra, line 669:
```suggestion
    {
      #if ASSIMP_HAS_BROKEN_GLTF_SPLINES
      CORRADE_EXPECT_FAIL("Current version of assimp incorrectly imports glTF spline-interpolated animations.");
      #endif
```

Again, to show that this is a known importer failure in the test output :) The scope is important as it limits what comparisons the expected failure applies to.

Same below.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -131,6 +151,27 @@ constexpr struct {
...
+
+constexpr struct {
+    const char* name;
+    const char* suffix;
+} GltfAnimationFileData[]{
+    {"ascii external", ".gltf"},
+    {"ascii embedded", "-embedded.gltf"},
+    {"binary external", ".glb"},
+    {"binary embedded", "-embedded.glb"}
+};
```
mosra, line 173: Does Assimp handle all these correctly the same? If it does, then I would use just a single variant, the one that's the easiest to produce, which is I suppose "ASCII external"?

In case of TinyGltfImporter these were important because unfortunately tinygltf used to have bugs and failed to import certain variants, but for Assimp I don't think we need to care that much :)

mosra (commented):
The more you look at Assimp, the more bugs it has :see_no_evil: 

Thanks!

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
mosra: Config value (that's enabled by default) sounds like a good idea :+1: Turning the option off might also give the users a fallback option in case this postprocessing turns out to have some unhandled corner case and provides wrong output (or if they need to be bug-compatible with Assimp for whatever other reason).

About empty tracks, I can see a use case where some postprocessing/optimization pipeline removes all keyframes because they're redundant / identity but the end-user app still expects to look up that particular animation under a name and play it. So I would personally keep them.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
mosra: `AnimationTrackTargetType` could work with `BigEnumSet`, which only requires the enum to have some sane upper bound. 

Another alternative would be `BoolVector3` but its constructor is quite shitty and doesn't allow you to construct from bools directly (I suppose that's because I had a brain failure when trying to make that constexpr.)

src/MagnumPlugins/AssimpImporter/Test/CMakeLists.txt
mosra: This is fine, I'm doing this regularly :)

mosra:
>  I would almost go as far as putting a big red warning on the importer docs à la "DON'T EVEN TRY TO USE THIS WITH GLTF".

I already mark the support as red in the [file format tables](https://doc.magnum.graphics/magnum/file-formats.html), but perhaps this case would warrant a black "DANGEROUS" label :laughing: 

image

> There is no test for Animation::Extrapolation beyond expecting Constant in tests. There is currently no code in Assimp that sets this

Good to know. Can you add this as a comment next to the `switch` to document the lack of code coverage for these lines?

> Collada [...] I didn't put that much effort into checking

That's fine. People who still use Collada for animations are fully responsible for their fate, not us. The format is an overengineered insanity.

> I wanted to test with older versions. Not sure how much of this is tested on CI, but I could give this another try with WSL (or as a CMake subdirectory - is that supported for Assimp?)

What's tested is 3.2 on Ubuntu 16.04, whatever is the current version on Homebrew and [5.0.1 on Windows](https://github.com/mosra/magnum-ci#assimp). Assimp supports a CMake subproject ([recipe in the docs](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1AssimpImporter.html#Trade-AssimpImporter-usage)) but I don't know how far back this works, it's possible that early 4.x might have issues that my Find module isn't aware of. Personally I'd only manually test with 4.0 and let the CI take care of the other versions, or even don't bother with version 4 at all.

- - -

About the CI failures:

- macOS and MinGW CI failures unrelated to your code should get fixed if you rebase on top of current master
- for the Linux failures, `CheckAssimpVersion.cpp` is referenced as `checkAssimpVersion.cpp` in CMake, and thus works only on case-insensitive filesystems (IIRC, with Git on Windows the rename has to be done in two steps, `CheckAssimpVersion.cpp` -> `a.cpp` -> commit -> `a.cpp` -> `checkAssimpVersion.cpp`, and then squash the commits together)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: I added `removeDummyAnimationTracks` (+ test and docs) and kept empty tracks completely untouched. A catch-all config value for both seemed silly, and so did an extra one just for cleanup.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
pezcode: Done

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
pezcode: Done

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
pezcode: This is much nicer, good to know

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
pezcode: Done

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -287,6 +349,690 @@ void AssimpImporterTest::openDataFailed() {
...
+    CORRADE_COMPARE(scaling.after(), Animation::Extrapolation::Constant);
+    CORRADE_COMPARE_AS(scaling.keys(), (Containers::StridedArrayView1D<const Float>{keys}), TestSuite::Compare::Container);
+    if(!ASSIMP_HAS_BROKEN_GLTF_SPLINES) {
+        constexpr Vector3 scalingData[]{
+            {-2.0f, 1.1f, -4.3f},
+            {5.1f, 0.1f, -7.3f},
+            {3.0f, 0.1f, 2.5f},
+            {1.5f, 9.8f, -5.1f}
+        };
+        CORRADE_COMPARE_AS(scaling.values(), (Containers::StridedArrayView1D<const Vector3>{scalingData}), TestSuite::Compare::Container);
```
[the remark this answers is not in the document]
pezcode: Done. FYI I copied that from `TinyGltfImporterTest`, same change should probably be done there as well. I'll send a PR for that tomorrow.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
pezcode: No difference I can see, I removed everything except for `animation.gltf`.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1214,23 +1216,97 @@ Containers::Optional<AnimationData> AssimpImporter::doAnimation(UnsignedInt id)
...
+                /* This might not perfectly restore the T/R/S components, but it's okay
+                   if the comparison fails, this whole fix is a best-effort attempt */
+                node->mTransformation.Decompose(nodeScaling, nodeRotation, nodeTranslation);
+
+                const bool verbose{flags() & ImporterFlag::Verbose};
+
+                if(translationKeyCount == 1 && channel->mPositionKeys[0].mTime == 0.0) {
+                    const Vector3 value = Vector3::from(&channel->mPositionKeys[0].mValue[0]);
+                    const Vector3 nodeValue = Vector3::from(&nodeTranslation[0]);
+                    if(Math::TypeTraits<Vector3>::equals(value, nodeValue)) {
```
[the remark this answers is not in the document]
pezcode: 😱 Makes the code so much cleaner. Out of curiosity, how would I perform a test for binary equality?

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1241,16 +1317,14 @@ Containers::Optional<AnimationData> AssimpImporter::doAnimation(UnsignedInt id)
...
     for(std::size_t a = animationBegin; a != animationEnd; ++a) {
         const aiAnimation* animation = _f->scene->mAnimations[a];
         for(std::size_t c = 0; c != animation->mNumChannels; ++c) {
             const aiNodeAnim* channel = animation->mChannels[c];
             const UnsignedInt target = doObject3DForName(channel->mNodeName.C_Str());
-            if(target == -1) {
-                Error{} << "Trade::AssimpImporter::animation(): unknown target node" << channel->mNodeName.C_Str();
-                return Containers::NullOpt;
-            }
+            CORRADE_ASSERT(target != -1, "Trade::AssimpImporter::animation(): target node must exist", {});
```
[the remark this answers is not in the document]
pezcode: Not really, no. I changed it to an internal assert.

pezcode:
CI still fails, compiler doesn't like the tuple assignment, but I'll rework that bit of code in another commit.

Thanks for the detailed feedback with code snippets, makes it a breeze to implement any changes 😃 

Exercises left:

- [ ] show `std::tuple` the door
- [ ] file massaging and basic player output testing in `animation()` test
- [ ] try to compile with older Assimp, at least 4.0

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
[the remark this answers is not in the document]
mosra: Oh, haha, sorry. Most probably a historical reason, the helper functions got added after I realized this is too annoying to type :D

If you don't have time left for the PR I can do that myself in the next round of cleanups.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -286,6 +338,713 @@ void AssimpImporterTest::openDataFailed() {
...
+   tests on certain versions and test files. */
+bool supportsAnimation(const std::string& fileName) {
+    /* 5.0.0 supports all of Collada, FBX, glTF */
+    if(ASSIMP_IS_VERSION_5)
+        return true;
+
+    /* splitExtension returns empty extensions for files starting with a dot */
+    const std::string extension = Utility::Directory::splitExtension("x" + fileName).second;
+    const bool isGltf = extension == ".gltf";
+    if(isGltf) return false;
```
mosra, line 353:
```suggestion
    if(Utility::String::hasSuffix(fileName, ".gltf") return false;
```

Or switch the whole thing to a `StringView` and use [hasSuffix()](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1BasicStringView.html#a403edf0edbe56aa02add16ee25df91f1).

(Commenting just because I was doing a very similar thing elsewhere just now :))

pezcode: Perfect, that just removed 2 lines of code and a variable 👌

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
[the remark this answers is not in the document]
pezcode: I'll just take you up on that offer then :eyes: This is taking much longer than expected already. Thanks!

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: I settled for `BigEnumSet`, works great

pezcode:
I fixed the CI failure on Linux locally (loading .gltf without checking for the Assimp version), but the Mac/Mingw one needs some more debugging. Track order seems to be mixed up there.

Final stretch 🙏

mosra:
> Track order seems to be mixed up there.

Sigh, FFS. I think this is the problem: https://github.com/assimp/assimp/blob/d855ec2472e82066aedb6d43efd63355d2eb2ef2/code/AssetLib/glTF2/glTF2Importer.cpp#L1364 There's no guarantee about iteration order for a `std::unordered_map`, so this stupid thing just gives out something totally different on every platform. 

So... I guess the test will need to do some mapping / reordering first (based on target type and target ID, maybe?) and then compare the ordered IDs. Probably not worth the extra maintenance hell to do this directly in the importer code, just in the test.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -482,9 +482,9 @@ void AssimpImporterTest::animation() {
 
     constexpr Vector3 rotationData[KeyCount]{
         {0.0f, 0.0f, 0.0f},
-        {0.0f, 0.0f, Float{Rad(-90.0_degf)}},
-        {Float{Rad(90.0_degf)}, 0.0f, Float{Rad(-90.0_degf)}},
-        {Float{Rad(135.0_degf)}, 0.0f, Float{Rad(-90.0_degf)}}
```
pezcode: Any idea why GCC 4.8 doesn't like the curly braces here?

> error: cannot convert 'Magnum::Rad {aka Magnum::Math::Rad<float>}' to 'Magnum::Float {aka float}' in initialization
         {Float{Rad(135.0_degf)}, 0.0f, Float{Rad(-90.0_degf)}}

pezcode:
Look at all these green checkmarks 😱🎉

I did have to skip checking expected rotation values for Collada files, but everything else is properly checked now. Note that this is not an issue with Assimp, but purely with Blender's Collada exporter being complete rubbish.

Pending any review changes (and adding copyright data), this is complete from my side.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
mosra: As far as I know, early C++11 spec didn't allow using `{}` for conversion operators, only for constructors. Old Clang (3.8) has the same issue. Similar issue is with initializing references, e.g. `int& a{b}`doesn't compile there. Newer compilers are more consistent and allow `{}` for all those cases. 

C++17 then allows using `{}` for enum initialization as well (the main reason was to be able to do e.g. `std::byte{56}`).

mosra:
I squashed the commits slightly, keeping the major parts separate, and merged as 9baaadb66f61b6b95d0a0e0f21564abfa309095c...e733222974b8bdf5589314280e2fd149a01a9463. Apart from the "integration tests" with [Littlest Tokyo](https://www.artstation.com/artwork/AJGbV) glTF and [Steam Locomotive](https://sketchfab.com/3d-models/br52-steam-locomotive-animation-852f498c26ee457d933f7fbeeb50c8ab) using magnum-player, which went great, I checked locally with AddressSanitizer and there were no [out-of-bounds reads or memory issues](https://github.com/mosra/magnum-plugins/pull/97#discussion_r657987988), only a silly

> ==579355==ERROR: AddressSanitizer: alloc-dealloc-mismatch (operator new vs operator delete []) on 0x60300001daa0

inside `~aiScene()`, but I don't care that much, really. Maybe it's already fixed in Assimp master, whatever.

Thanks for a great piece of work, and sorry for having to deal with this buggy :shit:

PR #98, opened 2021-06-20
TinyGltfImporter: Update docs about animation

pezcode:
Found these two small fixes while comparing what to copy over to `AssimpImporter`. Skin data seems to be imported, and the importer state for `AnimationData` wasn't being mentioned.

### DIFF
--- a/src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.h
+++ b/src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.h
@@ -131,7 +131,7 @@ calls with @ref InputFileCallbackPolicy::LoadTemporary and
 @ref InputFileCallbackPolicy::Close is emitted right after the file is fully
 read.
 
-Import of skeleton, skin and morph data is not supported at the moment.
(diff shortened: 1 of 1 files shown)


src/MagnumPlugins/TinyGltfImporter/TinyGltfImporter.h
```diff
@@ -131,7 +131,7 @@ calls with @ref InputFileCallbackPolicy::LoadTemporary and
 @ref InputFileCallbackPolicy::Close is emitted right after the file is fully
 read.
 
-Import of skeleton, skin and morph data is not supported at the moment.
+Import of skeleton and morph data is not supported at the moment.
```
mosra, line 134: I removed the "skeleton" from here as well and merged as f9e41c62ad43c8b314bfad7180d413032fdb5d15. This was originally here only because I had very little idea about what's the difference between a skin and a skeleton, and given that it's just a single ID that (apparently) almost nobody uses, I don't think it needs to be mentioned here.

Thanks for the PR! :+1:

PR #99, opened 2021-07-03
AssimpImporter: import skin data

pezcode:
This PR adds support for importing skins as well as joint weight mesh attributes with AssimpImporter.

`doSkin3D` imports skins (joint node ids + corresponding inverse bind matrices) for each mesh with skin data. Assimp doesn't expose any of the original (possibly multi-mesh) skins and instead copies joint data into each mesh, duplicating bones if necessary. The new `mergeSkins` option is meant to allow users to merge all skins in the scene into one, removing duplicate joints and adjusting mesh attributes.

`doMesh` imports joint ids and weights as custom mesh attributes (built-in support for those is WIP in https://github.com/mosra/magnum/pull/441) named "JOINTS" and "WEIGHTS". Multiple sets (4 weights per set) of these are imported as layers of the same custom mesh attribute id. This allows more than 4 weights per vertex but requires changing the new `maxJointWeights` option (default: `4`).

This is very much WIP, but I pushed my current progress so you can intervene early if anything is completely off-track 😄 

---

Commissioned by [Wonderland](https://wonderlandengine.com/).


pezcode:
Todo:

- [ ] `mergeSkins` is currently not de-duplicating bones
- [ ] Finish tests, a good part is already tested, but not everything
    - [ ] skin matrices, something off about rotations there
    - [ ] joint ids and weights from a common test file
    - [ ] `mergeSkins`
    - [ ] `maxJointWeights`, need some useful test data

src/MagnumPlugins/AssimpImporter/AssimpImporter.conf
```diff
@@ -49,6 +49,16 @@ mergeAnimationClips=false
 # key/value pair with the default node transformation
 removeDummyAnimationTracks=true
 
+# Maximum amount of joint weights per vertex to import as vertex attributes.
+# Requires the LimitBoneWeights option to be set as well to take effect.
+# This can be controlled separately for each mesh import.
+maxJointWeights=4
```
pezcode, line 55: This only sets `AI_CONFIG_PP_LBW_MAX_WEIGHTS` so I feel like I should list this in the AI_CONFIG_ section and name it appropriately, but `PpLbwMaxWeights` isn't really convincing me :eyes:

src/MagnumPlugins/AssimpImporter/AssimpImporter.h
```diff
@@ -287,9 +296,6 @@ verbosity levels in each instance.
 -   Custom mesh attributes (such as `object_id` in Stanford PLY files) are
     not imported.
 -   Texture coordinate layers with other than two components are skipped
--   For some file formats (such as COLLADA), Assimp may create a dummy
-    "skeleton visualizer" mesh if the file has no mesh data. For others (such
-    as glTF) not.
```
pezcode, line 292: This is fixed by setting `AI_CONFIG_IMPORT_NO_SKELETON_MESHES`, but there's another comment down below in the *Scene import* section ("[...] there's sometimes just a single root node") which I'm not sure about. The node collapsing doesn't happen anymore with this setting (the test got adapted accordingly), but I wonder if this is the only circumstance where this happens.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -1250,6 +1267,117 @@ void AssimpImporterTest::animationMerge() {
...
+    auto&& data = ExportedFileData[testCaseInstanceId()];
+    setTestCaseDescription(data.name);
+
+    if(!supportsSkinning(data.suffix))
+        CORRADE_SKIP("Skin data for this file type is not supported with the current version of Assimp");
+
+    /* Skinned mesh imported into Blender and then exported */
+
+    // TODO attribution:
+    // https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_019_SimpleSkin.md
```
pezcode, line 1285: Where would I best put attribution here? The tutorials are licensed under [CC-BY 4.0](https://creativecommons.org/licenses/by/4.0/) so anything allowed as long as we credit the author and add a link to the license. I'm afraid recreating a skinned mesh with Blender is a bit out of my depth.

pezcode:
There is a known test failure with one glTF file, but the CIs with old Assimp versions skip that. They still report failure with 0 failed tests, is that because of the warning about some tests not containing any checks?

src/MagnumPlugins/AssimpImporter/AssimpImporter.conf
pezcode: The comment about setting this per mesh import is also wrong, fixed locally.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -1757,6 +1889,68 @@ void AssimpImporterTest::lineMesh() {
...
+    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("AssimpImporter");
+    CORRADE_VERIFY(importer->openFile(Utility::Directory::join(ASSIMPIMPORTER_TEST_DIR,
+        "skin" + std::string{data.suffix})));
+    CORRADE_COMPARE(importer->meshCount(), 2);
+
+    /* Custom attributes should not be available until a mesh is imported.
+       These should return nothing (and not crash). */
+    CORRADE_COMPARE(importer->meshAttributeName(meshAttributeCustom(0)), "");
+    CORRADE_COMPARE(importer->meshAttributeName(meshAttributeCustom(564)), "");
+    CORRADE_COMPARE(importer->meshAttributeForName("thing"), MeshAttribute{});
```
mosra, line 1908: As per my comment above, `importer->meshAttributeForName("JOINTS")` *should* work here.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
```diff
@@ -1942,38 +2141,27 @@ void AssimpImporterTest::scene() {
...
-        const UnsignedInt version = aiGetVersionMajor()*100 + aiGetVersionMinor();
-        /** @todo Possibly works with other versions (definitely not 3.0) */
-        CORRADE_EXPECT_FAIL_IF(version <= 302,
-            "Assimp 3.2 and below doesn't use name of the root node for collapsed nodes.");
-        CORRADE_COMPARE(importer->object3DForName("Scene"), 0);
-        CORRADE_COMPARE(importer->object3DName(0), "Scene");
-    }
+    Containers::Pointer<ObjectData3D> parent = importer->object3D(0);
+    CORRADE_COMPARE(parent->children(), {1});
+    CORRADE_COMPARE(parent->instanceType(), ObjectInstanceType3D::Empty);
```
mosra, line 2164: Ummm ... I don't understand why is there such a weird interaction with `IMPORT_NO_SKELETON_MESHES`?

IIRC the original intent of this test was to verify that collapsing the whole file into a single node works, i.e. testing corner cases of [this block of code](https://github.com/mosra/magnum-plugins/blob/faa16d460d21dc8eb3682aec4335954db5dca758/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp#L409-L440) (and the COLLADA/skeleton-related comment there is probably outdated now?). 

I *think* in this case the expected behavior was that the importer skips the useless extra added root node and returns only its child, but I have no idea why it's no longer being done -- it should be.

The "only a single node" case in the code happens for PLY files for example (or some of the glTF files), but seems like the test doesn't explicitly verify that. I'll add a test case to check that code path.

src/MagnumPlugins/AssimpImporter/Test/skin.dae
```diff
@@ -0,0 +1,209 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mosra, line 1: The new files should get referenced in the test CMakeLists again :)

mosra:
> They still report failure with 0 failed tests, is that because of the warning about some tests not containing any checks?

Yes, that's why :)

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
mosra: Is that where the `skin.gltf` is from? Looks rather different, did you re-export with Blender? The test files from TinyGltfImporter could not be reused?

In any case, attribution should go [here](https://github.com/mosra/magnum-plugins/blob/faa16d460d21dc8eb3682aec4335954db5dca758/src/MagnumPlugins/AssimpImporter/AssimpImporter.h#L127-L131), something like

```patch
diff --git a/src/MagnumPlugins/AssimpImporter/AssimpImporter.h b/src/MagnumPlugins/AssimpImporter/AssimpImporter.h
index 0da7c4c3..2d03fa5c 100644
--- a/src/MagnumPlugins/AssimpImporter/AssimpImporter.h
+++ b/src/MagnumPlugins/AssimpImporter/AssimpImporter.h
@@ -128,7 +128,11 @@ This plugin provides `3dsImporter`, `Ac3dImporter`, `BlenderImporter`,
     licensed under @m_class{m-label m-success} **BSD 3-clause**
     ([license text](http://assimp.org/index.php/license),
     [choosealicense.com](https://choosealicense.com/licenses/bsd-3-clause/)).
-    It requires attribution for public use.
+    It requires attribution for public use. The tests use a file from the
+    [A Simple Skin](https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_019_SimpleSkin.md)
+    glTF tutorial, licensed under @m_class{m-label m-success} **CC-BY 4.0**
+    ([license text](https://github.com/KhronosGroup/glTF-Tutorials/blob/master/LICENSE),
+    [choosealicense.com](https://choosealicense.com/licenses/cc-by-4.0/)).
 
 @section Trade-AssimpImporter-usage Usage

```

src/MagnumPlugins/AssimpImporter/AssimpImporter.h
mosra: See [my comment here](https://github.com/mosra/magnum-plugins/pull/99#discussion_r663461312).

pezcode:
Thanks for taking the time to look at this on a Sunday, very much appreciated ☺️

src/MagnumPlugins/AssimpImporter/AssimpImporter.conf
pezcode: @mosra Not quite sure how to interpret the thumbs-up 😄 Do I keep it as it?

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
pezcode: Yeah, it's re-exported from Blender, mainly because I needed an extra un-skinned mesh in the scene. The *original* TinyGltfImporter skin test files can't be used with Assimp because they have no meshes. Assimp only exposes bones through meshes, so no meshes = no bones. Main reason I imported and re-exported that tutorial file is that it has accompanying joint weights, making it fairly simple to test both skins and vertex attributes with the same file.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -698,11 +760,56 @@ Containers::Optional<MeshData> AssimpImporter::doMesh(const UnsignedInt id, Unsi
...
+                maxJointCount = Math::max(jointCount, maxJointCount);
+            }
+        }
+
+        CORRADE_INTERNAL_ASSERT(!configuration().group("postprocess")->value<bool>("LimitBoneWeights") ||
+            maxJointCount <= configuration().value<UnsignedInt>("maxJointWeights"));
+
+        jointLayerCount = (maxJointCount + 3)/4;
+        attributeCount += jointLayerCount*2;
+        stride += jointLayerCount*(sizeof(Vector4ui) + sizeof(Vector4));
```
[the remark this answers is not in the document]
pezcode: Is there a way to keep GL from filling the last component with a 1? Purely from a user perspective, the majority of GLSL skinning code I could find (including the glTF docs) doesn't have any bone limit but simply uses the entire set in a uvec4+vec4. Asking people to add a check or modify attributes manually might be an annoying hurdle.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -698,11 +760,56 @@ Containers::Optional<MeshData> AssimpImporter::doMesh(const UnsignedInt id, Unsi
...
+           Joint ids and weights are the only custom attributes in this importer
+           so we can simply check the number of existing custom attributes. */
+        const char* attributes[]{"JOINTS", "WEIGHTS"};
+        CORRADE_INTERNAL_ASSERT(_f->meshAttributeNames.size() % Containers::arraySize(attributes) == 0);
+        const std::size_t firstAttributeLayer = _f->meshAttributeNames.size()/Containers::arraySize(attributes);
+        for(std::size_t layer = firstAttributeLayer; layer < jointLayerCount; ++layer) {
+            for(UnsignedInt i = 0; i < Containers::arraySize(attributes); ++i) {
+                const std::string attribute = Utility::formatString("{}_{}", attributes[i], layer);
+                const UnsignedShort attributeId = layer*Containers::arraySize(attributes) + i;
+                _f->meshAttributesForName.emplace(attribute, meshAttributeCustom(attributeId));
```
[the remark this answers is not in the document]
pezcode: Yeah... it's not ideal. Using `maxJointWeights` in `doOpenData()` won't quite work if someone disables the `LimitBoneWeights` postprocess option (we could ignore extra weights, but we'd be missing the weight prioritization and normalization step Assimp performs). One option is force-enabling `LimitBoneWeights` and letting users use `maxJointWeights` exclusively but then someone will inevitably set it to 99999999 🤡

Using the same attribute but different instances would make this a lot simpler, indeed. The main motivation for the unique attribute ids was for this to be as close to a drop-in replacement for `TinyGltfImporter` as possible, but if that PR gets merged in the foreseeable future I don't think the inconsistency is too bad.

src/MagnumPlugins/AssimpImporter/Test/skin.dae
pezcode: Whoops, putting this on my todo list so I can do this towards the end in case more test files are needed

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
pezcode: The comment in the test suggested it was Assimp doing the collapsing internally so I simply adapted the test. But if it tests some importer code then it might need some more scrutiny. I'm not entirely sure what's happening either, but I can do some more debugging.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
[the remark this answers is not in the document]
mosra: >  Using maxJointWeights

Dang, sorry, realized that we don't need to take that into account after writing the comment but then forgot to update it fully. Just a single attribute, yes.

> drop-in replacement for TinyGltfImporter

It's my fault that the essential functionality is not there yet, not yours :) I'll try to merge that PR and make this consistent as soon as possible.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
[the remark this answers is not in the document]
mosra: Yeah, true, not ideal... Keep it as is, then. Until there's a builtin shader that's flexible enough to handle such things it's best to not overcomplicate :)

src/MagnumPlugins/AssimpImporter/AssimpImporter.conf
mosra: Ah, oh, didn't realize it's about the `LimitBoneWeights` as well. This option alone makes sense to me, wouldn't need a separate `LimitBoneWeights` I think -- instead say that it sets `aiProcess_LimitBoneWeights` as well, and `0` disables any limit.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
pezcode: Mildly related question: what are *bone nodes*?

https://github.com/mosra/magnum-plugins/blob/faa16d460d21dc8eb3682aec4335954db5dca758/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp#L571

mosra: Outdated comment -- nodes associated with a skin. Basically what's provided by `SkinData` now, I suppose this TODO expected the reference would go the other way around. Remove :)

> Assimp doing the collapsing internally so I simply adapted the test.

IIRC Assimp does the collapsing internally if you enable PreTransformVertices, *but* still adds a silly empty root node with an identity transformation that the importer plugin then attempts to remove.

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
pezcode: @mosra turns out the original file is from the [glTF 2 sample repo](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/SimpleSkin), licensed under [CC0](https://creativecommons.org/choose/zero/). Do I still mention it in that case? It's not used verbatim, if that makes any difference.

pezcode:
Functionality is all there, I just need to finish a few tests...

Meanwhile, enjoy some animated, skinned [Mixamo](https://www.mixamo.com/) characters imported from FBX files into Wonderland Engine 👯

pezcode:
Bigger struggle than anticipated, but everything should be tested now, including:
- multiple sets of joint weight attributes
- the `maxJointWeights` option
- the `mergeSkins` option

For the last one I had to do some `openState` shenanigans because I couldn't for the life of me find a small file with joints shared between skins, or how to make such a file with Blender.

I took another look at Assimp code to understand the `AI_CONFIG_IMPORT_NO_SKELETON_MESHES` interaction with node collapsing but I'm no smarter than I was before. Is this blocking? I don't think turning `AI_CONFIG_IMPORT_NO_SKELETON_MESHES` back off again is great, but I also don't understand the implications for other file types too well.

mosra:
> Is this blocking?

I'll look into it during the merge.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1474,7 +1668,67 @@ Containers::Optional<AnimationData> AssimpImporter::doAnimation(UnsignedInt id)
...
+std::string AssimpImporter::doSkin3DName(const UnsignedInt id) {
+    /* If the skins are merged, don't report any names */
+    if(_f->mergeSkins) return {};
+    return _f->scene->mMeshes[_f->meshesWithBones[id]]->mName.C_Str();
+}
+
+Containers::Optional<SkinData3D> AssimpImporter::doSkin3D(const UnsignedInt id) {
+    /* Import either a single mesh skin or all of them together.
+       Since Assimp gives us no way to enumerate the original skins
+       and assumes that one mesh = one skins, we give the users
```
mosra, line 1707:
```suggestion
       and assumes that one mesh = one skin, we give the users
```

This is the only remaining issue I found by looking at the code :)

mosra (commented):
Your code looks fine. I still need to check it out, build and run locally... and look into that scene collapsing behavior.

The Windows builds fail because of [this](https://github.com/assimp/assimp/blob/57b1040a0c0dacfdf342b49c7a953468b47600d3/include/assimp/cexport.h#L53) that covers the entire file, which I [define for the Windows Assimp builds](https://github.com/mosra/magnum-ci/blob/7a4692170c868929380fbe442acd5a6bd81f681a/.github/workflows/build.yml#L38). So... I guess I should enable export functionality for these binaries? Another option would be to rewrite that test as a glTF file, I could attempt that.

Squareys:
@mosra asked me to hand-craft a gltf with two meshes that share a skin, here's the result:
[skin-shared.zip](https://github.com/mosra/magnum-plugins/files/6776397/skin-shared.zip)

src/MagnumPlugins/AssimpImporter/Test/AssimpImporterTest.cpp
mosra: I don't think that's needed, no.

mosra:
Wow, goddamit. So the node collapsing worked before *exactly because* the `AI_CONFIG_IMPORT_NO_SKELETON_MESHES` was not set, which added one extra mesh to the imported file, which then managed to circumvent this: https://github.com/assimp/assimp/issues/3820

After patching `scene.dae` to contain a mesh, the test case started passing again. Dumpster fire.

pezcode:
> Another option would be to rewrite that test as a glTF file instead of building it up using Assimp's own APIs, I could attempt that.

That's probably the better option because you don't have to deal with Assimp itself in the tests. I painfully edited an existing Collada file for the multiple weight attributes test, but constructing glTF buffers actually seems easier (and might uncover yet-unknown glTF Assimp bugs... I'm really suspicious I haven't hit any in the basic `skin()` test :eyes:).

mosra:
I managed to fix the Windows CI issue by using the underlying C++ APIs instead of the C wrappers, not sure why there's a difference. (Of course it's never simple, so I'm now testing a workaround workaround that negates the workaround for version 3.2.)

One last question, about the "ignore dummy joint weights" commit -- there doesn't seem to be any difference in the test output whether the code is there or not, and looking at https://github.com/assimp/assimp/blob/1d33131e902ff3f6b571ee3964c666698a99eb0f/code/AssetLib/glTF2/glTF2Importer.cpp#L1099 it seems this happens only when the glTF looks like this

```json
    "skins": [
        {
            "joints": []
        }
    ],
```

which is [invalid according to the spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-skin). So I wouldn't even bother adding a special case, and if Assimp crashes on such a file it's ... not our problem :)

pezcode:
> One last question, about the "ignore dummy joint weights" commit -- there doesn't seem to be any difference in the test output whether the code is there or not, and looking at https://github.com/assimp/assimp/blob/1d33131e902ff3f6b571ee3964c666698a99eb0f/code/AssetLib/glTF2/glTF2Importer.cpp#L1099 it seems this happens only when the glTF looks like this
> 
> ```json
>     "skins": [
>         {
>             "joints": []
>         }
>     ],
> ```
> 
> which is [invalid according to the spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-skin). So I wouldn't even bother adding a special case, and if Assimp crashes on such a file it's ... not our problem :)

I didn't look into it all that deeply, but I ran across it when testing this file exported from Blender:
https://discourse.threejs.org/t/vertex-groups-on-gltf-exported-model/4397

What happens is that the dummy weights lead to overflow in `jointCount` (no. of weights per vertex) because the code assumes that no two bones affect the same vertex, which isn't 100% correct if we're using the glTF spec as reference:

> No joint may have more than one **non-zero** weight for a given vertex

I suppose the correct (and slow) way to fix this would be to store everything in a map but that completely garbles the joint order (which Assimp sorts by weight, if it has to reduce the bone count in the postprocess step)

pezcode:
If I'm reading Assimp's code correctly, the dummy weights are added to all joints in the skin that aren't influencing any vertex in the mesh, which is entirely possible. Should I try to create a small test with a custom file for this?

mosra:
Ah, you're right, I misread the code.

I tried to craft a file by reducing `skins.gltf` by hand and then adding a few extra nodes not referenced from the mesh but got too impatient, this crap library a massive time sink every time I touch it :fire: 

If you're able to do that, that'd be great. Another thing is the leak inside `aiCopyScene()`, I don't know what to do about that, probably best would be to hand-craft another file there as well :/

pezcode:
Sweet Jesus! I think I found the dumbest Assimp bug yet.

I was checking if it correctly imports multiple sets of joint weights (what I originally wanted to do with the file linked earlier, but then forgot because it caused the bug with dummy weights), and not to much suprise, it [doesn't](https://github.com/assimp/assimp/blob/8f0c6b04b2257a520aaab38421b2e090204b69df/code/glTF2/glTF2Importer.cpp#L825).

Now comes the best part: it doesn't even import the first set of weights... The parsing code is so abysmal, they look for `JOINT` (note the missing trailing S), find it (because `JOINTS_0` exists), then try to extract the index by checking against JOINT_i which fails so each set gets index `i = 0`, overwriting each previous set. What you get is always the **last** set of joint weights. 🙄  

I added a check that finds non-normalized weights in glTF files and prints a warning (+ accompanying test), but there isn't much else we can do.

---

Now for better news: adding the other test was fairly simple, so the dummy weight removal has a proper test that fails if you remove the checks. For some reason Github doesn't show the commits in the main PR view, but they're there in the Commits tab. Together with https://github.com/mosra/magnum-plugins/pull/100 this should hopefully be the victory lap for this PR 🙏

mosra:
Thanks! Will get back to this hopefully this evening, now I have to work on other things ;)

mosra:
Almost done putting everything together. Thanks to #100 it doesn't leak anymore, but the newly added test for dummy weights had an error in the `.bin` size (520, should be 620), which triggered an OOB read inside Assimp:

```
=================================================================
==1125177==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x616000028d88 at pc 0x5581dfff6298 bp 0x7ffd5b2951d0 sp 0x7ffd5b294980
READ of size 512 at 0x616000028d88 thread T0
    #0 0x5581dfff6297 in __interceptor_memcpy.part.0 (/home/mosra/Code/magnum-plugins/build-clang-addresssanitizer/Debug/bin/AssimpImporterTest+0xc0297)
    #1 0x7fd2d67aa39b  (/usr/lib/libassimp.so.5+0x71a39b)
    #2 0x7fd2d67ab3e4  (/usr/lib/libassimp.so.5+0x71b3e4)
    #3 0x7fd2d67abc51  (/usr/lib/libassimp.so.5+0x71bc51)
    #4 0x7fd2d6398eae in Assimp::BaseImporter::ReadFile(Assimp::Importer*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Assimp::IOSystem*) (/usr/lib/libassimp.so.5+0x308eae)
    #5 0x7fd2d63ab26c in Assimp::Importer::ReadFile(char const*, unsigned int) (/usr/lib/libassimp.so.5+0x31b26c)

SUMMARY: AddressSanitizer: heap-buffer-overflow (/home/mosra/Code/magnum-plugins/build-clang-addresssanitizer/Debug/bin/AssimpImporterTest+0xc0297) in __interceptor_memcpy.part.0
Shadow bytes around the buggy address:
  0x0c2c7fffd160: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2c7fffd170: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2c7fffd180: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2c7fffd190: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2c7fffd1a0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x0c2c7fffd1b0: 00[fa]fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2c7fffd1c0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2c7fffd1d0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2c7fffd1e0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2c7fffd1f0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2c7fffd200: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==1125177==ABORTING

```

TOP NOTCH error recovery right there. Very secure. Much hardening. I'll just pretend I didn't see anything. :fire:

mosra:
I squashed the major part into a single commit, with the fun and exciting new bug/misfeature discoveries kept as a separate commits, and merged as faa16d460d21dc8eb3682aec4335954db5dca758...d62d58c9250bcbc1a250d7c11590b589d62fb9b8.
 
Thanks a lot for having the perserverance to work on this :sweat_smile:

pezcode:
Thanks for cleaning up the debris, I didn't realize there was so much work left to merge 🙏

Issue #101, opened 2021-07-09
AssimpImporter: Should normalize filepaths for external image paths


Squareys:
I just encountered a FBX which had a filepath with `\` along the lines of:

```
[error] Trade::AbstractImporter::openFile(): cannot open file /a/unix/path/Assets/Something/Textures\[...]_Diffuse.png
```

It turns out that Assimp does not normalize filepaths before opening. A workaround is to [set a file callback](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1AbstractImporter.html#adc2889e9ba4c89df2f0f559c94f44c3c).

Might eventually get around to creating a PR.

mosra:
Fixed with #102. Yay, Assimp!

PR #102, opened 2021-07-10
AssimpImporter: normalize image paths

Squareys:
Hi @mosra !

As described in #101, I have an FBX file "from the wild", which has a relative windows path reference. This currently fails to load on Unix, but works on Windows.

[The first commit](6157ecb2db32663d8675e5bd3e7fe47d7c3485fe) runs the tests fine. My compiler complained that the symbol in StbImageImporter.cpp#99 does not exist, and since that made no sense whatsoever, I didn't want to waste time on that and I added support for PngImporter (it's might be faster, even).

[The second commit](97a7b34e6d58417859ae59b7e7b180ed7ac0b026) adds the test first with EXPECT_FAIL on non-Windows. This is to ensure that the CI would have indeed failed to import the image file.

The final commit fixes the test and removes the EXPECT_FAIL.

Best,
Jonathan

### DIFF
--- a/src/MagnumPlugins/AssimpImporter/Test/CMakeLists.txt
+++ b/src/MagnumPlugins/AssimpImporter/Test/CMakeLists.txt
@@ -73,6 +73,8 @@ corrade_add_test(AssimpImporterTest AssimpImporterTest.cpp
         exported-animation.gltf
         image-filename-trailing-space.mtl
         image-filename-trailing-space.obj
+        image-filename-backslash.mtl
+        image-filename-backslash.obj
         image-mips.mtl
         image-mips.obj
         light.dae
--- a/src/MagnumPlugins/AssimpImporter/Test/CMakeLists.txt
+++ b/src/MagnumPlugins/AssimpImporter/Test/CMakeLists.txt
@@ -4,6 +4,7 @@
 #   Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019,
 #               2020, 2021 Vladimír Vondruš <mosra@centrum.cz>
 #   Copyright © 2021 Pablo Escobar <mail@rvrs.in>
+#   Copyright © 2021 Jonathan Hale <squareys@googlemail.com>
 #
 #   Permission is hereby granted, free of charge, to any person obtaining a
 #   copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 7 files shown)


src/MagnumPlugins/AssimpImporter/Test/image-filename-non-normalized.mtl
```diff
@@ -0,0 +1,13 @@
+newmtl Material
+Ns 96.078431
+Ka 1.0 1.0 1.0
+Kd 0.64 0.64 0.64
+Ks 0.5 0.5 0.5
+Ke 0.0 0.0 0.0
+Ni 1.0
+d 1.0
+illum 2
+# There's a space here ---v
```
mosra, line 10: Leftover comment copied from the other file that tested trailing whitespace in filenames? :)

src/MagnumPlugins/AssimpImporter/Test/CMakeLists.txt
```diff
@@ -69,6 +73,8 @@ corrade_add_test(AssimpImporterTest AssimpImporterTest.cpp
         exported-animation.gltf
         image-filename-trailing-space.mtl
         image-filename-trailing-space.obj
+        image-filename-non-normalized.mtl
+        image-filename-non-normalized.obj
```
mosra, line 77: I'd rename these to `image-filename-backslash` since that's what it is.

mosra (commented):
Thanks!

:+1: for putting the test into an earlier separate commit. Good practice, appreciate that. It didn't need to have the EXPECT_FAIL tho, since it's fixed right after :)

> My compiler complained that the symbol in StbImageImporter.cpp#99 does not exist

Huh, can you elaborate? I have a fork of stb_image with https://github.com/nothings/stb/pull/950 applied, so chances are it used some other file in your case. The test CMakeLists are deliberately inflexible to keep them small and easy to maintain (in this case I'd need to add the `PngImporter` variant to all the other importer tests as well) so if it's okay with you I'd skip that one commit.

Squareys:
> It didn't need to have the EXPECT_FAIL tho, since it's fixed right after :)

Well, this makes bisecting easier...

> it used some other file in your case

Oh! That could be 🤔 

> it used some other file in your case

Sure, no problem.


Changes applied!

mosra:
> Well, this makes bisecting easier...

Depends what's the intent, I guess :) 

In my case the reason to commit the test first is to have a commit that *clearly* fails to either compile or to pass the test, to have a verification that there's indeed a problem. Then the immediately following commit has to fix the issue and make things compile / pass tests again. If both commit would pass, then it wouldn't be so obvious (this test has 100 XFAILs already, hard to spot there's one more) and could potentially hide issues like 

- the first commit not actually testing the right problem (and the second commit then having no effect)
- the second commit not actually fixing the problem (and the XFAIL accidentally kept there)

- - -

I dropped the PngImporter commit, finished the rename and merged as d0398ed34095350c9b1ca3f31430d1041f362bca...d6234a6239c46ac435364c8d292115411bc906ab. Thanks, and I hope that's everything for Assimp bugs, this week at least :D

PR #103, opened 2021-07-14
Trade: KTX importer

pezcode:
👋 

Here's a first draft for the [KTX2](https://github.khronos.org/KTX-Specification/) image importer. It supports block-compressed and uncompressed 1D, 2D, 3D images.

~~Array layers are imported as separate images, cube faces and mips are imported as image levels (in the same order as `DdsImporter`: all mips for face 0, then face 1, etc.).~~

Array layers and cubemap faces are imported as extra image dimensions (so 1D array images become `Image2D`, (layered) cube maps become `Image3D` with n*6 z slices). The exception are 3D array images, those are imported as separate `Image3D` for each layer. No API supports those and there is no `Image4D`.

I wanted to get some early feedback on a few design decisions, so there's still a long list of todos, roughly sorted by importance:
- [x] tests, tests, tests
  - [x] enable the plugin(s) in `package/ci/*.{bat,sh}` for the CIs to pick them up
- [x] accompanying `KtxImageConverter`
    - [x] support for block-compressed images
    - [ ] support for cube and layered image types once those are queryable
    - [x] tests
- [x] swizzling into existing `PixelFormat` where possible
- [ ] ~~arbitrary Vulkan formats, wrapped as custom `PixelFormat`~~ out of scope of this PR
- [x] make sure the cubemap order is somehow standardized (compare with `DdsImporter`) or at least documented
- [ ] ~~Basis compression~~ out of scope of this PR
- [ ] ~~supercompression (zstd etc.)~~ out of scope of this PR


pezcode:
The dependency on `Magnum::Vk` is not great for obvious reason, and I mainly used it to get something working quickly. My plan is to just copy the `VkFormat` definition from the Vulkan headers (which needs attribution I think, Apache 2 license) and create the mapping to `PixelFormat` using raw Vulkan values. What's a little more involved is getting the block size for all formats, and I was considering grabbing that from https://github.com/KhronosGroup/KTX-Specification/blob/master/formats.json.

Also, what Vulkan formats should we support? 1.0 only?

src/MagnumPlugins/KtxImporter/KtxImporter.h
```diff
@@ -0,0 +1,92 @@
...
+*/
+
+/** @file
+ * @brief Class @ref Magnum::Trade::KtxImporter
+ */
+
+#include <Corrade/Containers/Array.h>
+#include <Corrade/Utility/VisibilityMacros.h>
+
+#include "Magnum/Trade/AbstractImporter.h"
```
pezcode, line 35: I fixed these relative includes locally, those are remnants from using `TgaImporter` as a starting point.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,553 @@
...
+    const Vector3i size = Math::max(Vector3i{originalSize}, Vector3i{1});
+
+    /* Number of array layers, imported as separate images */
+    const UnsignedInt numLayers = Math::max(header.layerCount, 1u);
+
+    /* Cubemap faces, either 1 or 6. KTX allows incomplete cubemaps but those
+       have to be array layers with extra metadata inside key/value data. */
+    const UnsignedInt numFaces = header.faceCount;
+    if(numFaces != 1) {
+        CORRADE_INTERNAL_ASSERT(numFaces == 6);
```
pezcode, line 385: These are all bugs in exporters, is it enough to assert here? Or should all these asserts be `Error` output? I went with `Error` for unsupported but valid scenarios or missing data (file too short etc.).

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+    if(originalSize.z() > 0) {
+        CORRADE_INTERNAL_ASSERT(originalSize.y() > 0);
+        f->dimensions = 3;
+    } else if(originalSize.y() > 0)
+        f->dimensions = 2;
+    else
+        f->dimensions = 1;
+
+    /** @todo Assert that 3D images can't have depth format */
```
pezcode, line 372: I thought it would be handy to have a function that tells you if a `PixelFormat` is a depth and/or stencil format. Might have to do this manually in this plugin to support all Vulkan formats, but maybe an idea for your ever-growing todo list 👀

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,553 @@
...
+                f->imageData[layer][currentLevel++] = {mipSize, f->in.suffix(level.byteOffset).slice(imageOffset, imageOffset + length)};
+
+                /* Shrink to next power of 2 */
+                mipSize = Math::max(mipSize >> 1, Vector3i{1});
+            }
+        }
+    }
+
+    /** @todo Read KTXorientation and flip image? What is the default? */
+    /** @todo Read KTXswizzle and apply it. Can't do that for block-compressed formats, just ignore in that case? */
```
mosra, line 445: Hm, probably something that's worth exposing directly in image metadata, if only the (Web)GL support wasn't so bad for this feature...

I'd say expect an identity swizzle (or a swizzle that's implemented for the BGRA etc. variants) for now and fail the import otherwise.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,553 @@
...
+    /** @todo Read KTXanimData and expose frame time between images through
+              doImporterState (same as StbImageImporter). What if we need
+              doImporterState for something else, like other API format
+              enum values (KTXdxgiFormat__ etc.)? */
```
mosra, line 449: Huh, KTX supports animations, even? :O Cool.

The animation "support" in StbImageImporter was an ugly quick hack for the [Animated GIF example](https://doc.magnum.graphics/magnum/examples-animated-gif.html) that shouldn't even get made this way.

I think what could work is adding `Trade::AnimationTrackTargetType::ImageLayer` (better name?) and expose the keyframes via `animation()`... and do the same fix for `StbImageImporter` (although they [plan to drop support for animated GIFs](https://twitter.com/nothings/status/1411264318100807681) as it's too much of a nightmare). Though, this is all quite tangential and personally I wouldn't even bother with this unless you have ane asy way to generate the test data for this and want to use them somewhere :)

src/MagnumPlugins/KtxImporter/KtxHeader.h
```diff
@@ -0,0 +1,81 @@
...
+#pragma pack(1)
+
+/* KTX2 file header */
+struct KtxHeader {
+    UnsignedByte    identifier[12]; /* File identifier */
+    UnsignedInt     vkFormat;       /* VkFormat enum value, VK_FORMAT_UNDEFINED = custom */
+    UnsignedInt     typeSize;       /* Size of data type in bytes */
+    UnsignedInt     pixelWidth;     /* Image level 0 width */
+    UnsignedInt     pixelHeight;    /* Image level 0 height */
+    UnsignedInt     pixelDepth;     /* Image level 0 depth */
```
mosra, line 45: Could be a `Vector3ui` even?

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
mosra: The importer should be immune to malformed files (if you download a file and half of it is garbage, for example), so asserts should be only self-checking the code itself, not the data.

I think it could be relatively easy to test all of them using an instanced test where you have some valid file as a starting point but parametrize the file length, expected failure message and do things like "change byte at position X to value Y". Or if even that would become too involved, we can look into other ways.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
mosra: Yes, this makes sense to have directly in Magnum, `isDepthStencilPixelFormat()` for example. I have many such utils for `VertexFormat` already, so a similar set could be for pixel formats.

mosra:
> My plan is to just copy the VkFormat definition from the Vulkan headers

> what Vulkan formats should we support? 1.0 only?

Copy the enum from Magnum's own `flextVk.h`, that way it's in sync with the set of formats Magnum supports.

> which needs attribution I think, Apache 2 license

If you copy from the `flextVk.h` I think not anymore, and this file is credited elsewhere.

> **cube faces** and mips **are imported as image levels** in the same order as `DdsImporter`

Huh, not sure if that was a good decision back then. I'd import cube maps as 3D images instead, one reason is that the `GL::CubeMapTexture` APIs can operate with 3D images directly, which would be way more convenient than having to deal with 6 separate images. To distinguish between a 3D, array and cubemap image I think it could make sense to expose a `TextureData` via `texture()`, containing the corresponding `Type` enum (argh, I need to add `Texture1DArray` / `2DArray` / `CubeArray` fields there ASAP, TODO `#1` for me).

Actually, the KTX file is always just one image, right? It can have multiple layers (in which case it's a 1D / 2D array, 3D or a cubemap) and multiple mip levels (which are exposed under via the `level` parameter), but all in all it's always just one, right? So then for this `TextureData` the `textureCount()` would always give you a `1`, the `TextureData::Type` tells how the image is supposed to be interpreted and only one of `image1DCount()` / `image2DCount()` / `image3DCount()` would be `1`, with all other `0`.

> make sure the cubemap order is somehow standardized (compare with DdsImporter) or at least documented

Yeah, the weird GL coordinate system. For now (until the above-mentioned `ImporterFlag::FlipY` flag lands) the importer should ensure the cubemap is in a coordinate system matching GL, which, depending on how KTX defines that, may involve flips along both X and Y. What might help you a bit:

- I half-did this for `OpenExrImporter`, abusing their decoder to do the flips for me, but didn't finish yet. Related WIP change is in a branch: https://github.com/mosra/magnum-plugins/commit/429785dab62b2ad9ad69dc6f2bb844056e55137a
- And here's a (again WIP dirty) patch to the `magnum-cubemap` example to make it load single-file images, for easy visual testing: https://github.com/mosra/magnum-examples/commit/3bab3269646c2809bb8326ded46d03655dcca040

Honestly I have no idea how it is in DDS, I suspect nobody ever tested that. So don't take that as the ground truth :)

> tests, tests, tests

A problem is that the `AbstractImageConverter` doesn't support multiple mips right now, which means you can't use the `KtxImageConverter` to generate multi-mip-level test data. Two options:

- put the related tests on hold for a ~week until I add that (it's been on my list for several other reasons already anyway and should be rather easy, TODO `#2` for me)
- temporarily use some Khronos' own utility to generate those files ... but I suspect this would further complicate things instead of helping you

TODO `#3` for me would be making `CompressedImage` store the block size parameters (and *require* them), which (I hope) could help you when accepting Vk-specific formats in `KtxImageConverter`? TODO `#4` would be adding that `Trade::AnimationTrackTargetType::ImageLayer` value in case you really want to bother with animations.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
pezcode: Testing the errors should be fairly straight-forward with instanced tests, I'm mostly worried about messy code because of all the checks. I might just stick most of them into a `validateHeader` function so validation and import code are less intertwined.

CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ option(WITH_HARFBUZZFONT "Build HarfBuzzFont plugin" OFF)
 option(WITH_ICOIMPORTER "Build IcoImporter plugin" OFF)
 option(WITH_JPEGIMAGECONVERTER "Build JpegImageConverter plugin" OFF)
 option(WITH_JPEGIMPORTER "Build JpegImporter plugin" OFF)
+option(WITH_KTXIMPORTER "Build KtxImporter plugin" OFF)
```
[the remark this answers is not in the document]
pezcode: I'd vote to keep it as is, but clearly mention it and maybe try to detect KTX1 and spit out a useful error message (as opposed to "wrong file signature"). Bgfx (or rather bimg) [supports version 1.1](https://github.com/bkaradzic/bimg/blob/8355d36befc90c1db82fca8e54f38bfb7eeb3530/src/image.cpp#L3951), but I have absolutely no intuition about usage in the wild. I could check out a few other texture convertors to see if KTX1 is prevalent enough to warrant a name change.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+/** @todo Swizzle channels if necessary. Vk::PixelFormat doesn't contain any
+          of the swizzled formats like B8G8R8A8. */
+/** @todo Support all Vulkan formats allowed by the KTX spec. Create custom
+          PixelFormat with pixelFormatWrap and manually fill PixelStorage/
+          CompressedPixelStorage. We can take all the necessary info from
+          https://github.com/KhronosGroup/KTX-Specification/blob/master/formats.json
+          Do we also need this for the KtxImageConverter? This would allow
+          users to pass in images with implementation-specific PixelFormat
+          using the Vulkan format enum directly. */
```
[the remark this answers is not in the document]
pezcode: The DFD has pixel/block size, but I figured since we have to map this information for the converter anyway, we might reuse it as 'ground truth' and not even deal with the DFD at all, or if we choose to, have a way to validate the info.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,553 @@
...
+        return ImageData1D(_f->pixelFormat.compressed, levelData.size.x(), std::move(data));
+    }
+
+    /* Uncompressed */
+    if(_f->needsSwizzle) swizzlePixels(_f->pixelFormat.uncompressed, data,
+        flags() & ImporterFlag::Verbose ? "Trade::KtxImporter::image2D():" : nullptr);
+
+    /* Rows are tightly packed */
+    PixelStorage storage;
+    storage.setAlignment(1);
```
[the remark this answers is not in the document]
pezcode: What exactly does `alignment` do? My understanding was that it tells you about possible padding between rows. Since KTX stores rows tightly packed, I thought this should be 1. The only padding KTX has is for the start of an entire mip level, all subsequent images inside that mip level (array layers, cubemap faces) aren't explicitly aligned to anything but the block size, if compressed.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+                f->imageData[layer][currentLevel++] = {mipSize, f->in.suffix(level.byteOffset).slice(imageOffset, imageOffset + length)};
+
+                /* Shrink to next power of 2 */
+                mipSize = Math::max(mipSize >> 1, Vector3i{1});
+            }
+        }
+    }
+
+    /** @todo Read KTXorientation and flip image? What is the default? */
```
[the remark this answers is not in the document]
pezcode: `FlipY` sounds like it's an unconditional operation, so `Yup` or something like `OriginBottom` is probably better. KTX recommends supporting y-up and y-down, as well as y-up + z-out and y-down + z-in for 3D images. It has no hard default in the absence of metadata, but they make a recommendation for exporters:

> Where possible, the preferred default is to use [y-down]

So I'd assume we should expect that as the default value, too.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
pezcode: I won't insist on this being added 😄 Just wrote down all my thoughts while reading the spec.

pezcode:
> Actually, the KTX file is always just one image, right? It can have multiple layers (in which case it's a 1D / 2D array, 3D or a cubemap) and multiple mip levels (which are exposed under via the `level` parameter), but all in all it's always just one, right? So then for this `TextureData` the `textureCount()` would always give you a `1`, the `TextureData::Type` tells how the image is supposed to be interpreted and only one of `image1DCount()` / `image2DCount()` / `image3DCount()` would be `1`, with all other `0`.

Every single one of 1D, 2D, 3D, cubemap can have array levels in a KTX file, so for cubemap arrays and 3D arrays you'd still have `image3DCount()` >= 1. We could also just not support some of these cases, like 3D and cubemap arrays and spit out a warning like "only the first array layer was imported", but I don't see this being too tricky to not support. The only thing that might be confusing is that 1D array texture is a 2D image, etc. but then 3D array textures need to be spread across multiple images and `TextureData`. Maybe add `imageRange` to `TextureData`, then you could have the full family of `Type`, including `Texture3DArray` and `TextureCubeArray`?

>     * temporarily use some Khronos' own utility to generate those files ... but I suspect this would further complicate things instead of helping you

I'm okay with this option, at least a few tests should use external data to make sure the converter and importer don't have some shared faulty assumption about the format, writing and then accepting broken files :eyes:

> TODO `#3` for me would be making `CompressedImage` store the block size parameters (and _require_ them), which (I hope) could help you when accepting Vk-specific formats in `KtxImageConverter`?

If that isn't more work than me just getting that data in a switch, that would be handy. Do any other importers have a use for this?

CMakeLists.txt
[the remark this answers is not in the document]
pezcode: I did some quick googling:
- AMD's [Compressonator](https://gpuopen.com/compressonator/) supports KTX 1 and 2
- NVIDIA's [Texture Tools Exporter](https://developer.nvidia.com/nvidia-texture-tools-exporter) supports KTX 1 and 2
- [PVRTexTool](https://developer.imaginationtech.com/pvrtextool/) supports KTX 1 and 2

The others I found don't support KTX2, but are mostly all unmaintained:
- Intel's [ISPC Texture Compressor](https://github.com/GameTechDev/ISPCTextureCompressor) doesn't support KTX at all
- ARM's [astcenc](https://github.com/ARM-software/astc-encoder) supports KTX 1.1
- [etcpack](https://github.com/Ericsson/ETCPACK) only supports KTX 1.1 but it hasn't seen any meaningful updates since 2016
- Google's [etc2comp](https://github.com/google/etc2comp) only supports KTX1 but same thing, no updates for several years
- [etcpak](https://github.com/wolfpld/etcpak) only supports KTX 1.1
- [betsy, used by Godot](https://github.com/darksylinc/betsy) only supports KTX 1.1
- [NVTT](https://github.com/castano/nvidia-texture-tools) only supports KTX 1.1

pezcode:
Feel free to ignore commits for now, mainly just pushing things for backup reasons. When I have tests and/or the converter working, I'll ping you again.

CMakeLists.txt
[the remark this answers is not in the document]
mosra: Amazing, thank you. Let's keep it as is, so `KtxImporter` being KTX2 :) 

For the compressors my long-term plan is to wrap them in a way that makes them container-independent (the first one implemented as such is [StbDxtImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StbDxtImageConverter.html), added yesterday), and thus their limitation to KTX 1 shouldn't be a problem in practice. My main concern was if Blender or Photoshop would support only KTX1 with people *heavily* using that format, but that's fortunately not the case.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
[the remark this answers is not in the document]
mosra: For code size reasons I'd rather not have the whole table in the importer, only in the converter. 

> have a way to validate the info

I'd definitely do that, but only for pixel / block size, not the other fields like channel bits and whatever else is there. Magnum has [pixelSize()](https://doc.magnum.graphics/magnum/namespaceMagnum.html#a0b7bf2313cad99f16b84cd0614938273) / [compressedBlockSize()](https://doc.magnum.graphics/magnum/namespaceMagnum.html#adb6a6c9d15a04272bbeaa4f42e552f3e) / [compressedBlockDataSize()](https://doc.magnum.graphics/magnum/namespaceMagnum.html#aeed2b6db91a62ffcb28752a178fa52ad) so I think it could be just checking these against the pixel format and relevant fields in the DFD and failing if these don't match?

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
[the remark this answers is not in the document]
mosra: > possible padding between rows

Yes, but generally I'm striving for keeping it at 4 when the rows actually *are* 4-byte aligned, because that means less GL state changes when uploading the data, possibly a more optimal pixel copy path chosen in the driver etc.

> KTX stores rows tightly packed

Huh, for some reasons I thought KTX goes a *really long way* to ensure everything is 4-byte aligned, metadata, rows, all that. Probably mixed that up with glTF, sorry. Yes, in this case, can you then set the alignment to 1 only if it's really needed?

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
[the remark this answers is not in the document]
mosra: Good point about defining the Z axis direction as well. I suppose GL will want both `YUp` and `ZBackward` flags, while Vulkan, D3D and all other software needs neither (Y down, Z forward).

> So I'd assume we should expect that as the default value, too.

Yep.

While implementing the [StbDxtImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StbDxtImageConverter.html) I came across code that can perform Y-flip of BC blocks, so eventually flipping could get supported for at least some block-compressed formats. But until I have that implemented, the importer should warn / fail.

"Funnily" enough, `DdsImporter` didn't even bother flipping anything, and didn't even document such behavior. So there the code needs to get fixed.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
mosra: Okay, let's ignore animations for now, then :) For fun I might try porting StbImageImporter to use the animation APIs and then that could eventually get mirrored here.

mosra:
> Every single one of 1D, 2D, 3D, cubemap can have array levels in a KTX file

Cube map arrays *are* a thing though, and GL/Vulkan implements them simply as a bunch of cubemaps following each other in a 3D image. So in case of these I'd still make them a single `image3D(0)`, with the `texture(0)` specifying it's a `CubeMapArray` and not a 3D texture or a 2D array (..once I add that to the enum).

4D textures, on the other hand... [were a thing in 1997](https://www.khronos.org/registry/OpenGL/extensions/SGIS/SGIS_texture4D.txt) but I don't remember any recent driver / platform bothering with these anymore. Since these don't map to hardware anyway, I'd say having them special-cased with multiple `image3D()`s & multiple `texture()`s wouldn't really be a problem?

> Maybe add imageRange to TextureData

That's not really mapping to hardware either, TextureData primarily specifies texture type and sampler options in order to form a texture out of ImageData. Why forcing the users to split the ImageData3D to multiple 3D textures to be able to use them at all if the importer can do that directly?

> at least a few tests should use external data to make sure the converter and importer don't have some shared faulty assumption about the format

Great point, yep, that definitely needs to be done :)

> If that isn't more work than me just getting that data in a switch

Right now I got stuck in a "fixing a lightbulb" loop while adding mip level support to the image converter, hopefully getting unstuck soon. After that I'll see how involved this is.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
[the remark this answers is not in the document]
pezcode: KTX1 had a row alignment of 4, that might be why it was in the back of your head. KTX2 [says](https://github.khronos.org/KTX-Specification/#levelImages):

> Rows of uncompressed texture images must be tightly packed, equivalent to a GL_UNPACK_ALIGNMENT of 1.

The `DdsImporter` does what you suggested but I didn't realize it was for performance reasons and removed the check, makes perfect sense now 👌

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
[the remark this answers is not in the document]
pezcode: The flipping code is now all there (although missing a few more tests), and currently always uses right, up, out as the target direction. For block-compressed files it prints a warning and performs no flipping at all. Changing the target direction shouldn't be more involved than changing two lines in the code, so I'll resolve this conversation for now.

pezcode:
Made a bit of progress in the last few days 🤠 

- array layers and cubemap faces are now always exposed as extra dimensions of the imported `ImageData`, along with `texture()` giving the correct type (pending the additional enum types in `TextureData::Type` being added to Magnum)
- 3d array images are imported as separate images/textures per array layer
- image data is now flipped to match GL's texture origin, taking KTX orientation metadata into account
- a basic version of `KtxImageConverter` supporting 1D/2D/3D with *uncompressed* formats, but it shouldn't be too tricky to expand
- a first round of tests for `KtxImporter`, missing a few of the 'special' cases like 3D, all array images, cubemaps

Apart from what I mentioned, the major missing pain point is cubemaps. They're imported just fine, but there are still a few oddities with orientations and coordinate space I don't full understand. Going to have to do a bit more reading and testing to see how this interacts with GL and what the official Khronos tools do.

pezcode:
I was bored and spent a few minutes implementing the frame animation data reading, so I'll just put the snippet here if one of us goes back to implementing it on top of a new animation type:

`KtxHeader.h`:
```c++
/* KTX2 animation frame */
struct KtxAnimationFrame {
    UnsignedInt duration;  /* Number of time units per frame */
    UnsignedInt timescale; /* Number of time units per 1 second */
    UnsignedInt loopCount; /* Loop count, 0 = infinite */
};

static_assert(sizeof(KtxAnimationFrame) == 12, "Improper size of KtxAnimationFrame struct");
```
`KtxImporter::doOpenData()`:
```c++
    /* Read animation data */
    if(isLayered) {
        const auto found = keyValueMap.find("KTXanimData"_s);
        if(found != keyValueMap.end()) {
            if(numLayers * sizeof(Implementation::KtxAnimationFrame) <= found->second.size()) {
                const auto frames = Containers::arrayCast<const Implementation::KtxAnimationFrame>(found->second);
                Float key = 0.0f;
                for(UnsignedInt layer = 0; layer != frames.size(); ++layer) {
                    /** @todo import as animation track with {key, layer} */
                    Implementation::KtxAnimationFrame frame = frames[layer];
                    Utility::Endianness::littleEndianInPlace(frame.duration, frame.timescale);
                    Float duration = frame.duration;
                    if(frame.timescale > 0)
                        duration /= frame.timescale;
                    key += duration;
                }
            }
            else
                Warning{} << "Trade::KtxImporter::openData(): invalid animation data, ignoring";
        }
    }
```

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,891 @@
...
+}
+
+template<std::size_t Size> struct TypeForSize {};
+template<> struct TypeForSize<1> { typedef UnsignedByte  Type; };
+template<> struct TypeForSize<2> { typedef UnsignedShort Type; };
+template<> struct TypeForSize<4> { typedef UnsignedInt   Type; };
+template<> struct TypeForSize<8> { typedef UnsignedLong  Type; };
+
+/** @todo Can we perform endian-swap together with the swizzle? Might get messy
+          and it'll be untested... */
```
mosra, line 72: Sounds like a good idea for a utility directly in Magnum. Goes onto my TODO list :)

mosra (commented):
Went thoroughly through the importer code, great job so far :+1: 

I have some of the needed additions to Magnum halfway done, will try to push them ASAP and then will look at the converter.

mosra:
> a few oddities with orientations and coordinate space I don't full understand. Going to have to do a bit more reading and testing to see how this interacts with GL and what the official Khronos tools do.

I'd say "just" import (and store) them in a way that makes them work properly with the expaded cubemap example (link to the code above)? This will need to get rethought for Vulkan and the flipping flags later anyway, but until I have something tangible to test this with I don't think you should need to care beyond what GL needs. Or am I missing something?

mosra:
The new `TextureType` enums got added in mosra/magnum@4c4b2596763c99cb37ae8b472d61935b4af26052 (together with a bunch of deprecations/renames in mosra/magnum@046b9558b73d64292cf1af5111160b7a952dd4eb), but I realized there's one related issue on the converter side -- how to tell it how a certain 2D/3D image should be interpreted? Right now there's just 1D/2D/3D variants but without any builtin way to specify if it's an array or a cube map, and introducing plugin-specific stringly typed configuration options for this sounds dirty to me.

- One option would be to reuse `TextureData` there *somehow*, but since that converter deals with images only and not related texture/sampling info (which is a job for the scene converters instead), plus the interface is one-off functions, adding some extra function to specify a `TextureData` referencing some image ID doesn't feel like a good match. (On the other hand eventually I may need to be able to accept `AnimationData` for animated images (video?), but that needs an API for batch images anyway...)

- Another is adding a completely new enum and a `setImageType()` API, which would be orthogonal to the functions called (the goal is so one doesn't have to explicitly `setImageType(ImageType::Image1D)` in order to call `convert()` with a 1D image because that would be annoying). So, for example:

  ```cpp
  enum class ImageType { // naming pending™
      Default,
      Array, // 2D images get interpreted as 1D arrays, 3D as 2D arrays
      CubeMap, // 3D images get interpreted as cube maps
      CubeMapArray // 3D images get interpreted as cube map arrays
  };
  
  converter.setImageType(ImageType::Array);
  converter.convertImageToFile(ImageView3D{...}, "foo.ktx"); // interpreted as 2D array
  converter.setImageType(ImageType::Default);
  converter.convertImageToFile(ImageView3D{...}, "bar.ktx"); // as a 3D again
  ```

  This however means there's two different ways to encode the same thing (`TextureType` vs `ImageType`) and now the requirement of having to expose `TextureData` in order to be able to distinguish cube maps / arrays and 3D images feels unnecessarily complex... Or *maybe* the image classes should have a `flags()` field on their own that describes the image beyond just size and format (what's the Y and Z direction, if the alpha is premultiplied, ..., and what type of image it actually is) 

- An alternative to the `ImageType` enum is expanding the `ImageConverterFlags` with `Array` and `CubeMap` flags (so it's also possible to say `Array|CubeMap`) instead? Sounds like a better option to me, plus there will soon be various other flags like the `YUp` etc. having the same usage / effect.

Opinions / ideas? :)

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    Utility::Endianness::littleEndianInPlace(length);
+    Utility::Endianness::littleEndianInPlace(header.vendorId, header.descriptorType,
+        header.versionNumber, header.descriptorBlockSize);
+
+    CORRADE_INTERNAL_ASSERT(offset == dfdSize);
+
+    return data;
+}
```
mosra, line 369: Crazy idea (sorry in advance) -- I was wondering, instead of all the above code which seems rather involved and (to me at least) not trivial to extend for new formats, would it be practical to just store a huge table of pre-filled DFDs for each format we have, generated from the JSON for example? Assuming 256 formats (currently there's about 160) and (upper-bound) a 64 byte DFD for each (BC5 has 60), this would be 16 kB of data, which isn't too much I'd say. Then adding a new format would be just about regenerating from the JSON, and putting the DFD into the output would be just a copy from this array.

Apologies for the time spent on the code :sweat_smile:

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
```diff
@@ -0,0 +1,608 @@
...
+    /* Greatest common divisor */
+    while(b != 0) {
+        const UnsignedInt t = a % b;
+        a = b;
+        b = t;
+    }
+    const UnsignedInt gcd = a;
+
+    return product/gcd;
+}
```
mosra, line 441: Interesting that I still don't have such utils in the Math lib. TODO for me.

src/MagnumPlugins/KtxImporter/KtxHeader.h
```diff
@@ -0,0 +1,458 @@
...
+
+/* Taken from magnum/src/MagnumExternal/Vulkan/flextVk.h
+   (commit 9d4a8b49943a084cff64550792bb2eba223e0e03)
+
+   Contains all formats from:
+   - 1.2 core (without KHR aliases)
+   - EXT_texture_compression_astc_hdr
+   - IMG_format_pvrtc */
+enum VkFormat : UnsignedInt {
+    VK_FORMAT_UNDEFINED = 0,
```
mosra, line 46: Just a random though I had today -- since we don't really need to use the enum values anywhere, what if `formatMapping.hpp` had directly the numeric values, with the enum string perhaps only as a documentation comment? Since you're extracting the data from one enum already, why not go all the way and dereference it fully. That way updating for newly added formats would be just rerunning the `.py` script, without also having to copypaste this enum here.

src/MagnumPlugins/KtxImporter/Test/generate.sh
```diff
@@ -0,0 +1,50 @@
...
+toktx --t2 --input_swizzle bgra bgr.ktx2 pattern.png
+toktx --t2 --target_type RGBA --input_swizzle bgra bgra.ktx2 pattern.png
+
+# 2D
+toktx --t2 rgb.ktx2 pattern.png
+toktx --t2 --target_type RGBA rgba.ktx2 pattern.png
+
+# manual mipmaps
+toktx --t2 --mipmap rgb-mipmaps.ktx2 @@pattern-mips.txt
+
```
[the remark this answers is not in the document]
pezcode: `toktx` seems to choke on creating 3D images, not sure if I'm doing something wrong or that path is completely untested. I was thinking I'd just generate a 2D layered image and patch the header to make it look like 3D since the data layout is the same, and `toktx` doesn't support 3D layered images. For those we'd need something else, not sure if using `magnum-imageconverter` is considered cheating 👀

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,891 @@
...
+                    const UnsignedInt actual = (block.descriptorBlockSize - sizeof(block))/sizeof(Implementation::KdfBasicBlockSample);
+                    valid = valid && actual == expected;
+                }
+            }
+        }
+
+        if(!valid) {
+            Error{} << "Trade::KtxImporter::openData(): invalid data format descriptor";
+            return;
+        }
```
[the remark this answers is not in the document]
pezcode: I'm OK with dropping it, less code to test 😄 Mismatching `vkFormat`/DFD is simply a violation of the spec but there really isn't much we can do, and the `vkFormat` is our ground truth anyway. We *will* need rudimentary DFD parsing if we implement Basis support since vkFormat will be `UNDEFINED`.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,891 @@
...
+    CORRADE_INTERNAL_ASSERT(imageLength > 0);
+
+    /* Both lengths should be equal without supercompression. Be lenient here
+       and only emit a warning in case some shitty exporter gets this wrong. */
+    if(header.supercompressionScheme == Implementation::SuperCompressionScheme::None &&
+        level.byteLength != level.uncompressedByteLength)
+    {
+        Warning{} << prefix << "mismatching image data sizes, both compressed "
+            "and uncompressed should be equal but got" <<
+            level.byteLength << "and" << level.uncompressedByteLength;
```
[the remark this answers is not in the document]
pezcode: You're making me wonder if this a useful warning anyway. I kind of followed along with the KTX spec requirements, but what can anyone really do with that warning?

src/MagnumPlugins/KtxImporter/Test/generate.sh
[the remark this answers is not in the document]
mosra: > For those we'd need something else

For EXR I was generating the ground truth data with `./OpenExrImageConverterTest --save-diagnostics .` and then using those in `OpenExrImporterTest` (yeah and then cross-checking that these images are still readable by the official tools). The `imageconverter` doesn't even have 3D image support right now, I'd need to add that there first, going through the converter test seems like a lot easier option :)

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
pezcode: [formats.json](https://github.com/KhronosGroup/KTX-Specification/blob/master/formats.json) sadly won't help since it doesn't contain any info that Magnum doesn't already expose for existing `(Compressed)?PixelFormat`. I wanted to use that to get that info for arbitrary Vulkan formats, but for generating the DFD it's of no use.

One alternative would be using (parts of) the [Khronos DFD lib](https://github.com/KhronosGroup/KTX-Software/tree/master/lib/dfdutils), it looks self-contained and not terribly bloated.

src/MagnumPlugins/KtxImporter/KtxHeader.h
pezcode: The importer uses the original Vulkan enums to detect swizzled BGR/BGRA formats that don't exist in `PixelFormat`, so the enum is still needed. I could try to extract the enum from flextVk.h? Alternative B: a custom flextVk template, but that might go out of sync too. The last option would be amazing if we had the .xml that Khronos use to generate the vulkan_core.h used by dfdutils that has some of the unreleased (but supported by KTX) formats like 3D ASTC.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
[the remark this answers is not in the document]
mosra: For Basis I'd only detect it and then delegate the whole thing to `BasisImporter` (opening the data there), similarly as `AnyImageImporter` does. I think that shouldn't need any messing with DFD.

But in a followup step, let's keep this PR in a reasonable scope :)

pezcode:
Thanks for reviewing 👍 I adressed all the minor items locally, they'll be pushed along with the completed tests once I finished those.

> 
> > a few oddities with orientations and coordinate space I don't full understand. Going to have to do a bit more reading and testing to see how this interacts with GL and what the official Khronos tools do.
> 
> I'd say "just" import (and store) them in a way that makes them work properly with the expaded cubemap example (link to the code above)? This will need to get rethought for Vulkan and the flipping flags later anyway, but until I have something tangible to test this with I don't think you should need to care beyond what GL needs. Or am I missing something?

Mostly I need to make sure that image flips don't affect the coordinate space. The KTX spec mentioned e.g. having to flip faces for mismatching y-orientation. But there's no major code to write, just needs testing with GL.

> Opinions / ideas? :)

I'm a fan of the last option. Ideally, there'd also be a way to query those flags per image. Then we could even scrap exposing the image type through `texture()` for image-only importers. Something like `imageFlags2D()`? But I'm not sure if you want to mix this all with importer config or create a new enum.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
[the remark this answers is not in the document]
mosra: I'd say if the imported data end up being garbled (or if the import would crash :scream:), this warning could provide some hints why that might have happened. Better than ignoring this, I'd say (and compared to verifying the DFD which is just a block of rendundant information, I think this actually *is* useful to check).

src/MagnumPlugins/KtxImporter/KtxHeader.h
mosra: > swizzled BGR/BGRA formats

Right. What about having this VkFormat enum list only the formats we need to act on in some way, and the rest being encoded directly as numeric values in `formatMapping.hpp` (so extracted from both `flextVk.h` and `Vk/PixelFormat.h`)? That way updating would in most cases mean regenerating only the `formatMapping.hpp`, and the `VkFormat` enum will get extended only in rare scenarios.

> 3D ASTC

Completely forgot about these, sigh. [And seems like I asked already](https://github.com/KhronosGroup/KTX-Specification/pull/97#pullrequestreview-293626218), two years ago -- no way to get access to that XML I guess :(

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
mosra: Mmmm, C code generated by *Perl* scripts. Nope nope nope, that sounds like a *much bigger* maintenance hell than I'd want to suffer. Ah well, I thought they would have some *data* file enumerating all DFDs for all Vulkan types, but if I have to run code in order to (again) get a C code and not plain data, that's not the way I want to go.

Keep it like this, then.

mosra:
> I'm a fan of the last option. Ideally, there'd also be a way to query those flags per image.

Yep, I settled on the same conclusion, this is *essential* to have otherwise dealing with GL/Vulkan coordinate system differences would quickly turn into a nightmare. So basically:

- have `Image`, `ImageView` and `ImageData` a `flags()` field that can store `YUp` / `ZBackward` / `NonPremultipliedAlpha` and potentially other bits, where lack of the bit (the default) means the saner / more common variant (so Y down, Z forward and premultiplied alpha)
- these flags get filled by various APIs as needed, for example reading a GL framebuffer / texture would set the `YUp` bit
- image converters would have an option to *override* this when saving the image and encode the information into the file where possible (so e.g. `KtxImageConverter` would implicitly save with Y down because that's what everything else expects, but could be told otherwise with `ImageConverterFlag::YUp`, in which case it would flip the image if it doesn't have `YUp` already, and it also saves this orientation into file metadata; but `PngImageConverter` would always save with Y down as the file doesn't support arbitrary orientation)
- image importers would implicitly import the file with the orientation specified in the metadata (and would save that into `flags()` of the returned image), but again can be told to override this on import -- so for example when importing images for a GL pipeline it makes sense to implicitly enable the `YUp` bit on the importer

I still need to think about backwards compatibility to avoid breaking everyone's projects, but most probably the defaults would be "import with Y up" and "save with Y down", and somehow begging people to set these explicitly.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,891 @@
...
+                "right"_s, "down"_s, "forward"_s
+            };
+            Warning{} << "Trade::KtxImporter::openData(): missing or invalid "
+                "orientation, assuming" << ", "_s.join(Containers::arrayView(defaultDirections).prefix(f->numDimensions));
+
+            constexpr auto defaultOrientation = "rdi"_s;
+            const BoolVector3 flip = Math::notEqual(Math::Vector3<char>::from(defaultOrientation.data()),
+                Math::Vector3<char>::from(targetOrientation.data()));
+            for(UnsignedByte i = 0; i != f->numDimensions; ++i)
+                f->flip.set(i, flip[i]);
```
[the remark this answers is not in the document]
pezcode: I'm afraid that still sets it to `true` for higher dimensions that should be ignored. For e.g. a 1D image without metadata, `flip` would become `{false, true, true}`. This makes `any()`/`none()` useless, which is used right below. I added a comment to clarify since I couldn't find a code golf solution that didn't make the code unreadable.

src/MagnumPlugins/KtxImporter/KtxImporter.cpp
```diff
@@ -0,0 +1,891 @@
...
+            /* Point to the last item of the dimension */
+            srcData += strides[dimensions - 1 - i]*(sizes[dimensions - 1 - i] - 1);
+            strides[dimensions - 1 - i] *= -1;
+        }
+    }
+
+    Containers::StridedArrayView<dimensions + 1, const char> srcView{{srcData, src.size()}, sizes, strides};
+
+    Utility::copy(srcView, dstView);
+}
```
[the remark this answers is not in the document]
pezcode: Yeah, classic case of outsmarting myself. This works perfectly at less than half the code size 😄

pezcode:
Starting to get to the finish line 🥉 

I converted the cubemap example files to KTX and everything just seemlessly works when loaded in GL, so no special code needed there. If you want to try yourself, here's the output of PVRTexTool:
[cubemap.zip](https://github.com/mosra/magnum-plugins/files/6910407/cubemap.zip)
Colors are a bit messed up, but the orientations are fine. You'll also need https://github.com/mosra/magnum/pull/529.

Apart from that, I'm just finishing tests, a few esoteric cases need covering, but the majority is there.

For compressed formats, how do I best go about getting ground truth data? From a quick glance, `--save-diagnostics` requires `CompareImage`, which doesn't support compressed formats.

CIs are failing because I forgot to update the test files in CMake, but they do seem to compile, so there's that.

src/MagnumPlugins/KtxImporter/Test/KtxImporterTest.cpp
```diff
@@ -0,0 +1,1334 @@
...
+    CORRADE_COMPARE(image->compressedFormat(), data.format);
+    CORRADE_COMPARE(image->size(), data.size);
+
+    const CompressedPixelStorage storage = image->compressedStorage();
+    CORRADE_COMPARE(storage.rowLength(), 0);
+    CORRADE_COMPARE(storage.imageHeight(), 0);
+    CORRADE_COMPARE(storage.skip(), Vector3i{});
+
+    /** @todo Can we test ground truth data here? Probably have to generate it
+              using KtxImageConverter. */
```
mosra, line 659: To answer your other comment as well:

> For compressed formats, how do I best go about getting ground truth data? From a quick glance, `--save-diagnostics` requires `CompareImage`, which doesn't support compressed formats.

`CompareImage` is intended mainly for fuzzy comparison which for compressed formats would mean I need a software decoder for every one of them. It's planned, but pretty far ahead. Another option would be `TestSuite::Compare::StringToFile` which also implements `--save-diagnostic` (yes, a string is shitty, and the console output is also shitty, but it *can* save a file) but that would mean you'd need one ktx file and one raw block data file for each test case, which to me feels a bit excessive.

I'd say do this (since blocks are always 64 or 128 bits anyway, this will always work)

```cpp
    CORRADE_COMPARE_AS(
        Containers::arrayCast<const UnsignedLong>(image.data()),
        Containers::arrayView<UnsignedLong>({}), // probably stored in data.expected instead?
```

run the tests, and grab the ground truth data from the failure diagnostic? :) Assuming you briefly verified the plugin to work externally (opened the image through `magnum-player` for example), this would set things in stone, ensuring that no file reading errors or memory issues happen in the future.

Oh, actually, wouldn't you need this exact sequence of bytes for tests in KtxImageConverter anyway? There you would write it to a file (which you then compare against the ground truth result from PVRTexTool) and what's the input block data there is the output block data in this test.

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
```diff
@@ -0,0 +1,932 @@
...
+
+    Containers::Array<Implementation::KtxLevel> levelIndex{numMipmaps};
+
+    const std::size_t levelIndexSize = numMipmaps*sizeof(Implementation::KtxLevel);
+    std::size_t levelOffset = sizeof(Implementation::KtxHeader) + levelIndexSize +
+        dataFormatDescriptor.size() + keyValueData.size();
+
+    /** @todo Handle block-compressed formats. Need block size and block count
+              in each dimension. How should skip be handled, if it's not a
+              multiple of the block size? */
```
mosra, line 788: Same as I said above, can you for now just put an (internal) assert for skip being zero and row length / image height not defined? When I add what's needed to the CompressedImage APIs, I'll update the code here.

src/MagnumPlugins/KtxImporter/formatMapping.py
```diff
@@ -0,0 +1,138 @@
...
+
+with open(file_out, 'w', encoding='utf-8') as outfile:
+    compressed = lambda f : f.compressed
+    formats, compressed_formats = partition(compressed, formats)
+
+    outfile.write(header)
+
+    outfile.write('#ifdef _p /* PixelFormat */\n')
+    for format in formats:
+        outfile.write('_p(' + format.vulkan + ', ' + format.magnum + ', ' + format.type + ')\n')
```
mosra, line 132: To avoid accidents I think it could be useful to store the enum names as a comment? Like, if you'd save the `vulkan_name` in the above loop as well and then

```suggestion
        outfile.write('_p({}, {}, {}) /* {} */\n'.format(format.vulkan, format.magnum, format.type, format.vulkan_name)
```

Same for the compressed formats.

src/MagnumPlugins/KtxImporter/formatMapping.py
```diff
@@ -0,0 +1,138 @@
...
+    outfile.write('#endif\n')
+
+    outfile.write('#ifdef _c /* CompressedPixelFormat */\n')
+    for format in compressed_formats:
+        outfile.write('_c(' + format.vulkan + ', ' + format.magnum + ', ' + format.type + ')\n')
+    outfile.write('#endif\n')
```
mosra, line 138: Hmm, and instead of putting both regular and compressed pixel formats together, could this go into `compressedFormatMapping.hpp`? Feel free to just hardcode the output filenames here, the utility doesn't have to be *that* flexible.

src/MagnumPlugins/KtxImporter/Test/generate.sh
```diff
@@ -0,0 +1,99 @@
...
+# is identical to a 3D image, but the orientation metadata will become invalid.
+# This won't pass validation but KtxImporter should ignore invalid metadata.
+# However, because the z axis will be flipped, we need to pass the images
+# in reverse order here compared to 2d-layers.ktx2.
+PVRTexToolCLI -i black.png,pattern.png,pattern.png -o 3d.ktx2 -array -f r8g8b8,UBN,sRGB
+# depth = 3, numLayers = 0
+printf '\x03\x00\x00\x00\x00\x00\x00\x00' | dd conv=notrunc of=3d.ktx2 bs=1 seek=28
+# We can't patch a 2D array texture with mipmaps into a 3D texture because the
+# number of layers stays the same, unlike shrinking z in mipmap levels
+# TODO find another way to generate a test file
```
mosra, line 74: I'm going to extend `magnum-imageconverter` with cubemap-related functionality, then you could use that here? It's an extreme case of dogfooding tho.

Or... just generate it with the `KtxImageConverter` test, since I suppose there you'd need to test cube mipmaps as well?

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
```diff
@@ -0,0 +1,932 @@
...
+        case PixelFormat::Depth32F:
+        case PixelFormat::Depth24UnormStencil8UI:
+        case PixelFormat::Depth32FStencil8UI:
+            return 4;
+        default:
+            break;
+    }
+
+    CORRADE_ASSERT_UNREACHABLE("componentSize(): unsupported format" << format, {});
+}
```
mosra, line 170: This looks like something that could be useful straight in Magnum. 

Hmmm but I'm not sure about the output for combined depth/stencil formats, is this what KTX expects? Also what about potential new formats like `R11FG11FB10F`? Would it help anything if this would return a `Vector4ui` with a size per component and 0 for components that aren't present?

mosra (commented):
Great work, again.

Is this finishable without me implementing the `CubeMap` / `Array` / ... image and converter flags first? Anything else this is being blocked on?

I probably won't be able to finish `CompressedImage::blocks()` and related APIs anytime soon, but also I don't want to put this on hold just because of these -- I think it should be doable without these, if you won't support the pixel storage parameters for now.

src/MagnumPlugins/KtxImporter/Test/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
+#
+#   This file is part of Magnum.
+#
+#   Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019,
+#               2020, 2021 Vladimír Vondruš <mosra@centrum.cz>
+#   Copyright © 2021 Pablo Escobar <mail@rvrs.in>
```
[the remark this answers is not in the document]
pezcode: oops, good catch

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
```diff
@@ -0,0 +1,932 @@
...
+    }
+
+    /* Copy the pixels into output, dropping padding (if any) */
+    Utility::copy(image.pixels(), Containers::StridedArrayView<dimensions + 1, char>{pixels, sizes});
+}
+
+template<UnsignedInt dimensions>
+void copyPixels(const BasicCompressedImageView<dimensions>& image, Containers::ArrayView<char> pixels) {
+    /** @todo How do we deal with CompressedPixelStorage::skip?
+              ImageView::pixels handles this for non-compressed images. */
```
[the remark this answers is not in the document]
pezcode: Is it enough to assert that `image.storage() == CompressedPixelStorage{}`? Because the block (data) size in there is also ignored and a mismatch should be an error.

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
```diff
@@ -0,0 +1,932 @@
...
+    Utility::copy(image.data(), pixels);
+}
+
+/* Using a template template parameter to deduce the image dimensions while
+   matching both ImageView and CompressedImageView. Matching on the ImageView
+   typedefs doesn't work, so we need the extra parameter of BasicImageView. */
+template<UnsignedInt dimensions, typename T, template<UnsignedInt, typename> class View>
+Containers::Array<char> convertLevels(Containers::ArrayView<const View<dimensions, T>> imageLevels) {
```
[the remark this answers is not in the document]
pezcode: Didn't think of that, works just fine 👌

src/MagnumPlugins/KtxImageConverter/Test/KtxImageConverterTest.cpp
```diff
@@ -0,0 +1,376 @@
...
+    const UnsignedByte data[16]{};
+
+    std::ostringstream out;
+    Error redirectError{&out};
+    CORRADE_VERIFY(!converter->convertToData({
+        ImageView2D{PixelFormat::RGB8Unorm, {2, 2}, data},
+        ImageView2D{PixelFormat::RGB8Unorm, {1, 1}}
+    }));
+    CORRADE_COMPARE(out.str(),
+        "Trade::KtxImageConverter::convertToData(): image data for level 1 is nullptr\n");
```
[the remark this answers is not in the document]
pezcode: BasisImageConverter [checks for empty size and nullptr](https://github.com/mosra/magnum-plugins/blob/2d82ec817d7c72a21a6376858f6f22dcf2a8a43c/src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp#L132), that's why I figured I'd add those checks. But I'll gladly remove them 🙇

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
pezcode: Those sizes were taken straight from the KTX spec, [relevant paragraph](https://github.khronos.org/KTX-Specification/#_depth_and_stencil_formats). The `Vector4ui` with channel sizes would be very useful for dealing with the DFD, pretty handy way to get the channel count and bit size.

pezcode:
That was a great review round, mostly just deleting things 😆 

> Is this finishable without me implementing the `CubeMap` / `Array` / ... image and converter flags first? Anything else this is being blocked on?

Absolutely, once those are added it shouldn't take a whole lot more than setting the correct values in the KTX header. The layout of the level data won't change. Until then all data is exported as 1D/2D/3D.

> I probably won't be able to finish `CompressedImage::blocks()` and related APIs anytime soon, but also I don't want to put this on hold just because of these -- I think it should be doable without these, if you won't support the pixel storage parameters for now.

Yeah, I just fixed the last remaining issue with block-compressed image export, and if I ignore the storage parameters, this is 99% done. Just need some more code for depth/stencil format DFD writing, and then tests.

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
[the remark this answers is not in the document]
mosra: Yes, that should be enough :)

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
mosra: Ah, haha, a brainfart, sorry. Apparently I even documented it exactly like this, as [being consistent with the KTX spec](https://doc.magnum.graphics/magnum/namespaceMagnum.html#a0f3e774f6adcfb4b7d06df036c378c24a3771460406ccad7bebc0a4815e94d2e0).

Let's keep this utility here for now, when I have a second use case for it I'll move it to Magnum.

src/MagnumPlugins/KtxImageConverter/Test/KtxImageConverterTest.cpp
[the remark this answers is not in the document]
mosra: Good to know, thanks -- yes, remove them :)

I added checks for empty lists, zero size, null views and format consistency in base implementation in mosra/magnum@eea9d32959c5b150cd0836de19164f5abcd1d11b, and the tests in `ImageConverter` plugins got cleaned up in 6e8a22dd32c984788d98deadb3a61a7d96ac6590.

pezcode:
I had to force-push to get the tests to pass that got updated in https://github.com/mosra/magnum-plugins/commit/6e8a22dd32c984788d98deadb3a61a7d96ac6590, looks a bit messy now 🙊 

KtxImageConverter now supports 3 config values:

- **orientation**: Orientation string of the form _[rl][du][oi]_ to write into the header. Doesn't flip any data, but it's handy until the image flags for flipping are added, and it's required for comparing test output in the converter tests against files output by toktx/PVRTexTool.
- **swizzle**: Swizzle string to write into the header. Doesn't flip any data, but handy to have if the reader supports it.
- **writer**: Name of the writing software/library. Mainly used for tests, but could be useful for things like `magnum-imageconverter`.

Block-compressed and mipmap images are now exported correctly, along with depth formats. Everything except for depth and 3D mipmapped textures is compared bit-by-bit against ground output by toktx or PVRTexTool. The others are 'tested' by comparing to files extracted with `--save-diagnostic`. KtxImporter also reads these and compares the content, but for the depth textures, this means the DFD is entirely untested.

Regarding the DFD in general: I was wondering if it's a good idea to revisit simply bundling the entire DFD for all formats. With a tiny patch to [a tool in KTX-Software](https://github.com/KhronosGroup/KTX-Software/blob/master/lib/dfdutils/testbidirectionalmapping.c) we can get it to write the DFDs of all supported Vulkan formats to disk. We could bundle these as resource files, they weigh up at ~12kb total, maybe 10kb if we remove all the formats not supported by Magnum. It's not as simple as copying a .json and running a python script, but it's not much more tricky. It would remove a lot of messy, hard to maintain DFD writing code and means supporting new formats added to Magnum works almost out of the box. If not a good idea for the converter itself, we could bundle the files and use them in the tests. Opinions?

The remaining CI errors are funny GCC 4 syntax problems, and the Emscripten `KtxImageConverter` test not finding the `KtxImporter` test files. I'll take a look at those tomorrow 🤠

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
```diff
@@ -729,7 +729,7 @@ Containers::Array<char> KtxImageConverter::convertLevels(Containers::ArrayView<c
             return {};
         }
 
-        constexpr Containers::StringView validOrientations[3]{"rl"_s, "du"_s, "io"_s};
```
pezcode: Any idea what's happening here? GCC 4.8 kept complaining about:

> error: cannot initialize aggregate of type 'const StringView [3] {aka const Corrade::Containers::BasicStringView<const char> [3]}' with a compound literal

but the **exact** same code works perfectly fine in `KtxImporter::doOpenData`.

pezcode:
I added rudimentary documentation to both plugins. Let me know if anything is missing. There's surprisingly little to document because virtually all formats are supported.

package/ci/emscripten.sh
```diff
@@ -97,6 +97,8 @@ cmake .. \
     -DWITH_ICOIMPORTER=ON \
     -DWITH_JPEGIMAGECONVERTER=OFF \
     -DWITH_JPEGIMPORTER=OFF \
+    -DWITH_KTXIMAGECONVERTER=OFF \
```
mosra, line 100: Why is it `OFF` here? :eyes:

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
mosra: Heh. I suppose it's because the function is templated? Putting it outside (and into a `namespace {`) should make the `constexpr` work, I hope.

mosra:
Hmm, apart from the few minor comments above I think this is pretty much finished? Good job on the docs as well, I can't think of anything that would be missing there right now.

> I was wondering if it's a good idea to revisit simply bundling the entire DFD for all formats. With a tiny patch to a tool in KTX-Software we can get it to write the DFDs of all supported Vulkan formats to disk. [...] we could bundle the files and use them in the tests

I'd say let's merge the PR in the current form, and then maybe revisit this in a followup? Honestly I'm a bit skeptical about the "tiny patch", from my experience where Khronos software is a constant moving target I'm afraid I could end up in a situation where the original tool gets "refactored", the patch is no longer possible and I end up reverting back to your original code.

But for the tests I think it's a good idea (if it's not too much extra work for you), worst case potential new formats would have nothing to verify against.

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
```diff
@@ -0,0 +1,929 @@
...
+    const UnsignedByte typeSize = formatTypeSize(format);
+    /* Compressed integer formats must use 32-bit lower/upper */
+    const UnsignedByte mappingBitLength = (isCompressedFormat ? sizeof(UnsignedInt) : typeSize)*8;
+    /* @todo BC6h has unsigned floats, but the spec says to use a sampleLower
+             of -1.0. Is this an error?
+             https://www.khronos.org/registry/DataFormat/specs/1.3/dataformat.1.3.html#bc6h_channel
+             The signed channel format flag is still set, however. */
+    const auto lowerUpper = channelMapping(suffix, mappingBitLength);
+    const UnsignedByte formatFlags = channelFormat(suffix);
+    /* For non-compressed RGBA channels, we get */
```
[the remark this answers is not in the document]
pezcode: This happens far too often, but I usually catch them when committing 🙊

src/MagnumPlugins/KtxImageConverter/KtxImageConverter.cpp
```diff
@@ -0,0 +1,929 @@
...
+    const bool isDepthStencil = !isCompressedFormat &&
+        sampleData.second().front().id != Implementation::KdfBasicBlockSample::ChannelId::Red;
+    
+    const UnsignedByte typeSize = formatTypeSize(format);
+    /* Compressed integer formats must use 32-bit lower/upper */
+    const UnsignedByte mappingBitLength = (isCompressedFormat ? sizeof(UnsignedInt) : typeSize)*8;
+    /* @todo BC6h has unsigned floats, but the spec says to use a sampleLower
+             of -1.0. Is this an error?
+             https://www.khronos.org/registry/DataFormat/specs/1.3/dataformat.1.3.html#bc6h_channel
+             The signed channel format flag is still set, however. */
```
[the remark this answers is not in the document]
pezcode: I submitted an issue: https://github.com/KhronosGroup/DataFormat/issues/16

src/MagnumPlugins/KtxImageConverter/Test/KtxImageConverterTest.cpp
```diff
@@ -0,0 +1,1022 @@
...
+        CORRADE_INTERNAL_ASSERT(Containers::arraySize(bytes) >= compressedBlockDataSize(CompressedPixelFormat(format)));
+
+        std::ostringstream out;
+        Error redirectError{&out};
+        CORRADE_VERIFY(!converter->convertToData(CompressedImageView2D{format, {1, 1}, bytes}));
+
+        /** @todo Is there a better way to do this? */
+        std::ostringstream formattedOut;
+        Debug redirectDebug{&formattedOut};
+        Debug{} << "Trade::KtxImageConverter::convertToData(): unsupported format" << format;
```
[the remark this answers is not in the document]
pezcode: Well, the only unsupported ones are 3D ASTC, which Magnum supports, so the ball is mostly in Vulkan's court here 👀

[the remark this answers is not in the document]
mosra: I consider those my problem, not Vulkan problem :)

package/ci/emscripten.sh
mosra: Should be enough to list all files it needs in the `FILES` section of `corrade_add_test()` similarly as you do for the importer test, e.g.

```cmake
corrade_add_test(KtxImageConverterTest KtxImageConverterTest.cpp
    LIBRARIES Magnum::Trade
    FILES
        ../../KtxImporter/Test/1d.ktx2
        ...)
```

That'll make Emscripten embed the files in the virtual filesystem which then the test can directly read. Though not sure if the relative paths work, if not then `${PROJECT_SOURCE_DIR}/src/MagnumPlugins/KtxImporter/`Test` instead.

pezcode:
I'm going to strangle GCC 4.8 one of these days 🤡

pezcode:
I addressed all your review comments and fixed a few remaining DFD bugs, but the DFD is now tested against dfdutils output for all formats.

Emscripten CI is still failing, I'll do some investigating on my own machine. `Directory::list()` doesn't seem to interact well with the virtual filesystem, or I'm doing something wrong with the paths in CMakeLists.txt

mosra:
Okay, I did a final pass over the latest changes and this looks exceptionally good and the test coverage is amazing as well. I'll do a bunch of tests locally (playing with `magnum-imageconverter`, especially), wire up the remaining 1D and compressed cases to AnyImageImporter / AnyImageConverter and merge.

One remaining thing -- I was thinking to to put all the dfd blobs into a single file, having them separately feels a bit excessive :laughing: Thinking of rewriting the patch to open the file for append instead and then prefix each DFD blob with size + format ID as two 32bit numbers, that could hopefully do the trick. If you have time to do it yourself, great, if not, I'll do it during merge.

pezcode:
> If you have time to do it yourself, great, if not, I'll do it during merge.

I can do that, yeah. That would also fix an interesting error on Windows with Ninja about the max command line length being exceeded :eyes: Ran into that when testing Emscripten and I couldn't figure out how to get it to use a response file for command line arguments (and not just object files.)

mosra:
Squashed into a few commits (didn't really find a way how to break it up better, without leaving a lot of noise in the history) and merged as 2114bf2b65418cea88c9b3b63ff7a0a23b02f64c...bd3f16e60335e21d7bf071e011f382025af13856. Thank you for this work! :+1:

Once I manage to finish the leftovers (array/cube annotation flags for images), I might have some followup tasks for this, and the Basis detection/proxying as well.

pezcode:
> didn't really find a way how to break it up better, without leaving a lot of noise in the history

I tried to break it up into commits specific to each importer, but the test dependency (as well as the single commit for registering both plugins for CI) kinda defeated that plan. Hope it wasn't too much work to squash these.

Thanks for the smooth review process and speedy addition of required Magnum changes on the way ☺️

mosra:
Ha, did I not close this yet? Interesting.

Issue #105, opened 2021-09-06
Unable to install magnum-plugins.deb on Linux Mint 20.2


thehans:
Linux Mint 20.2 (based on Ubuntu 20.04, Focal)

I tried following instructions here to install `magnum-plugins`:
https://doc.magnum.graphics/magnum/building-plugins.html#building-plugins-packages-deb

When running `dpkg-buildpackage` the build attempts a secure upload at the end which fails.  
```
...
dpkg-deb: building package 'magnum-plugins-dev' in '../magnum-plugins-dev_2020.06_amd64.deb'.  
dpkg-deb: building package 'magnum-plugins' in '../magnum-plugins_2020.06_amd64.deb'.          
dpkg-deb: building package 'magnum-plugins-dbgsym' in 'debian/.debhelper/scratch-space/build-magnum-plugins/magnum-plugins-dbgsym_2020.06_amd64.deb'.
        Renaming magnum-plugins-dbgsym_2020.06_amd64.deb to magnum-plugins-dbgsym_2020.06_amd64.ddeb                                                                                           
 dpkg-genbuildinfo                                                                             
 dpkg-genchanges  >../magnum-plugins_2020.06_amd64.changes                                     
dpkg-genchanges: info: including full source code in upload                                    
 dpkg-source --after-build .                                                                   
dpkg-buildpackage: info: full upload; Debian-native package (full source is included)          
 signfile magnum-plugins_2020.06.dsc                                                           
gpg: keyblock resource '/home/hans/.gnupg/pubring.kbx': Permission denied                      
gpg: skipped "Vladimír Vondruš <mosra@centrum.cz>": No secret key
gpg: dpkg-sign.ampZCdiJ/magnum-plugins_2020.06.dsc: clear-sign failed: No secret key                                                                                                           
                                                                                               
dpkg-buildpackage: error: failed to sign .dsc file 
```

I assume this is inconsequential since it still appears to have successfully built the .deb file, but I might as well mention it because the following step also fails:
```
$ sudo dpkg -i ../magnum-plugins*.deb                                                                                                  
(Reading database ... 524126 files and directories currently installed.)                                                                                                                       
Preparing to unpack .../magnum-plugins_2020.06_amd64.deb ...                                                                                                                                   
Unpacking magnum-plugins (2020.06) over (2020.06) ...                                          
Preparing to unpack .../magnum-plugins-dev_2020.06_amd64.deb ...                                                                                                                               
[...]

PR #108, opened 2021-10-03
on Buster plugins should depend on libopenexr23

playgroo:
$ uname -a
Linux 4.19.0-17-amd64 #1 SMP Debian 4.19.194-3 (2021-07-18) x86_64 GNU/Linux

$ sudo apt install libopenexr22 libopenexr24
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package libopenexr22 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or is only available from another source
However the following packages replace it:
  libopenexr23

Package libopenexr24 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or is only available from another source

E: Package 'libopenexr22' has no installation candidate
E: Package 'libopenexr24' has no installation candidate

### DIFF
--- a/package/debian/control
+++ b/package/debian/control
@@ -17,5 +17,5 @@ Description: Plugins for the Magnum C++11/C++14 graphics engine
 Package: magnum-plugins
 Section: libs
 Architecture: any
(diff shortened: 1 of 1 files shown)


mosra:
Thank you! :)

I did a similar fix for Ubuntu 20.04 recently in #105 which used `libopenexr24` for OpenEXR 2.3 for some reason, wasn't aware Debian names it `libopenexr23`.

playgroo:
Thank you for your efforts! I made this pr exactly based on your explanations in #105 :)

PR #111, opened 2021-10-17
RPM build

1b00:
To build RPM packages into the default location cd into package/rpm, read the docs and run build_rpm.sh (with or without parameter: pass either 'master' or any valid git tag|hash or no parameter for last tag)


mosra:
Merged together with #147 as bfab5712c6fef7c34f5aff43f592409526239ce2, thank you!

Issue #110, opened 2021-10-12
KTX2 + Basis Universal


pezcode:
I had time to kill so I spent an afternoon investigating how to hook up `KtxImporter` with Zstandard supercompression and the latest Basis Universal formats. Mainly writing this down so I don't forget again, but also happy to get some input.

### Zstandard

This was rather trivial to add to `KtxImporter`, but might be unneeded (more below).

- `zstd` has a script to generate [single file libraries](https://github.com/facebook/zstd/blob/dev/build/single_file_libs/README.md). The decoder is a ~600kB .c file without dependencies that can be bundled. The readme mentions a compiled WASM binary size of 26kB for the decoder, but I haven't verified that.
- decoding is dead simple, not more than 3 function calls, and it's handled transparently

### Updating basisu

Updating basis_universal to 1.15 for `BasisImporter` worked fairly quickly, too, with two caveats:
- converter tests don't pass because image delta is over the limit, I assume this is the result of some new quality setting somewhere
- keeping compatability with old versions (mainly the prehistoric 1.11 on vcpkg) is a little trickier because some functions and options changed names, but can be worked around with some SFINAE dances (there might be a version define to `#if` on, haven't checked)

### Forwarding Basis-compressed KTX2 to `BasisImporter`

This one gave me a bit of a headache. I naively assumed `BasisImporter` could handle the raw data inside KTX2 level data, but currently it only decodes .basis containers. There are ways to use the lowlevel transcoders, but for BasisLZ (Basis ETC1 + LZ supercompression) this is completely insane because you need to validate and decompress the global LZ data and somehow pass that through to `BasisImporter`.

The far easier solution here is to detect Basis data in `KTXImporter` early and then directly send the entire file data to `BasisImporter`. Basisu 1.15 can decode KTX2 and handles all the BasisLZ supercompression stuff for us. `BasisImporter` then only needs a few minor changes to use `ktx2_transcoder` next to the existing `basisu_transcoder` (which only supports .basis containers).

Another effect would be that basisu handles all Zstandard decoding, so that might not have to be handled in `KtxImporter`, unless it's commonly used outside of Basis-compressed KTX files. Needs some investigation.

mosra:
Done with #112 and #113 :tada:

PR #112, opened 2021-10-19
Upgrade Basis plugins to 1.15

pezcode:
This PR upgrades `BasisImporter` and `BasisImageConverter` to work with basis_universal 1.15 (more specifically, tag `v1_15_update2` because that contains a few important changes).

Not entirely finished with everything, but the majority is there, so feel free to give it a first review :bow:

## New features

Along with the upgraded dependency come a few new features:

- import of all image types supported by Basis, including 2D arrays/cubemaps/cubemap arrays/3D/video as `ImageData3D`
- KTX2 import and export, including UASTC and Zstd supercompression
  - import happens transparently, with the minor caveat that 3D images are not supported. `basisu` itself can only generated 2D arrays for KTX2, anyway.
  - in `BasisImageConverter` this is simply a new config option (defaults to `off`)
- support for `sRGB` formats:
  - in `BasisImporter` this is a flag read from the file header, affecting the output format
  - in `BasisImageConverter` it's detected from the image format and sets the correct parameters for the encoder. Can be overridden by config options, if desired
- support for user-supplied mipmaps in `BasisImageConverter`
    - mipmap generation by basis is still possible, but is disabled if the user supplies their own mipmaps

## Todo

Or rather: questions to @mosra

- [x] ~~allow export of 3D images in `BasisImageConverter`? Is this a priority?~~ nope, but probably revisit once image flags are in, for cube and 2D array texture export
- [x] ~~the importer code got rather messy, should I write a small abstraction over the two transcoders to make the code legible again?~~ eh, that'll just sweep the code into another corner
- [x] ~~should we flip z in 3D images? Files have a y-flip flag so we could use that to deduce y-flip = z-flip, but I'm not sure how foolproof that is since the `basisu` tool doesn't do any z-flipping at all, regardless of `y_flip` being set~~ nope, "solved" now that we import 3D images as 2D array layers
- [x] ~~BC7RGB seems deprecated since UAST, should this go away? Investigate what this means for the output data~~ they both mean the same thing, removing the RGB variant
- [x] ~~print a useful error about PVRTC1 requiring power-of-two, instead of just "transcoding failed"? However, then we won't have an easy way to test the transcoding failing 🤡~~ PVRTC is pretty ancient, not worth the extra effort
- [x] more validation on `.basis` files. The KTX2 transcoder does all that, but the basis transcoder is missing a few checks we need, e.g. all array layers having the same size.
- [x] update the config options, there are a few new ones ~~, some of them seem obsolete and we should add some comments (just take whatever `basisu` itself says). **Also**: the docs say the config options follow the `basisu` tool, and the `.conf` file says they follow the C++ API. The latter is true, but we should decide on one. I think `basisu` options would be more intuitive but require some extra wiring + tests...~~
- [x] test `BasisImageConverter` sRGB handling

---

Related issue: https://github.com/mosra/magnum-plugins/issues/110. Can be closed if this PR and the upcoming PR for `KtxImporter` Basis-forwarding are merged.


mosra:
Merged as 809fc1829db8e41d52fd7d299a3bc511b7ba9d83...670133c2d35d649dbd78ed0dec538578a47f0ffe, with some of my commits mixed in (external zstd dependency, fix to that C language enablement craziness, ...). 

The commits are regrouped and squashed into small enough logical chunks so it looks quite different from the patch set in this PR, but the total bulk of changes is the same ;) Thank you for the excellent work again!

PR #113, opened 2021-10-29
Ktximporter: forward Basis Universal compressed images to BasisImporter

pezcode:
Follow-up to https://github.com/mosra/magnum-plugins/pull/112. Only the two most recent commits are relevant, since this is on top of https://github.com/mosra/magnum-plugins/pull/112 to make sure the tests pass on CI.

If KtxImporter finds a file compressed with Basis Universal (either through BasisLZ supercompression, or a Basis UASTC format in the DFD), it transparently proxies to an internal instance of `BasisImporter`. Flags and config options are propagated, I shamelessly copied that from `AnyImageImporter` 🤠

### Possible Todos

- [ ] config option to turn this behaviour off, in case users want full control over which plugin is used
- [ ] when `openData()` is called with `DataFlag::Owned`, there's currently no way to forward this to another importer, incurring a needless copy. A fix would depend on https://github.com/mosra/magnum/pull/240

Related issue: https://github.com/mosra/magnum-plugins/issues/110.


mosra:
Merged as 6fbb0b8ee3b48fb774a7d964d92114596704237b...cefedba5a07e5ca60774ed513a4e8db15dfef017. Sorry for the long delays with everything, to avoid delaying even further with another round of reviews, I pushed this directly with a few things done differently:

- The plugin is not passing through all options anymore as I imagine KtxImporter could eventually get its own options and this behavior may lead to a mess. On the other hand I don't expect BasisImporter to get that many new options over time, so there's now a `[basis]` section containing just the `format` option.
- There's extra logic about the `format` delegation in that it tries to preserve the format globally set for BasisImporter, and prints a warning in case the two plugins have conflicting values, with KtxImporter having a priority.
- I split out the "revert back to plain KTX" test into a dedicated test case, to not have the main instanced test do too much.
- And tested the absence of a plugin in a dedicated case as well, since that's quite easy to simulate there.

Thank you for completing this huge chunk of work!

PR #118, opened 2022-01-04
BasisImageConverter: support for 2D array images

pezcode:
This adds missing functionality to `BasisImageConverter` to allow 2D array images.

The rationale for treating `ImageView3D` as a 2D array image is the same reason all 3D images in `BasisImporter` are imported as 2D array images. The tl;dr is that basis_universal assumes mip levels are 2D and KTX2 export doesn't support 3D images either.

There are minor fixes for `KtxImageConverter` in the commit list too, noticed those while cross-referencing code and docs.

### DIFF
--- a/src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
+++ b/src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
@@ -80,9 +80,12 @@ Containers::Array<char> convertLevelsToData(Containers::ArrayView<const BasicIma
     /** @todo Handle different image types (cube/array/volume) once this can be
         queried from images */
     Math::Vector<dimensions, Int> mipMask{1};
-    if(dimensions < 3) {
+    if(dimensions == 1) {
         /* Basis doesn't support 1D images, we treat them as 2D images with
            height 1 */
+        Warning{} << "Trade::BasisImageConverter::convertToData(): exporting 1D image as a 2D image with height 1";
+        params.m_tex_type = basist::basis_texture_type::cBASISTexType2D;
+    } else if(dimensions == 2) {
         params.m_tex_type = basist::basis_texture_type::cBASISTexType2D;
     } else {
         /* Encoding 3D images as KTX2 always produces 2D array images and mip
--- a/src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
+++ b/src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
@@ -80,14 +80,9 @@ Containers::Array<char> convertLevelsToData(Containers::ArrayView<const BasicIma
     /** @todo Handle different image types (cube/array/volume) once this can be
         queried from images */
     Math::Vector<dimensions, Int> mipMask{1};
-    if(dimensions == 1) {
-        /* Basis doesn't support 1D images, we treat them as 2D images with
-           height 1 */
-        Warning{} << "Trade::BasisImageConverter::convertToData(): exporting 1D image as a 2D image with height 1";
+    if(dimensions == 2) {
         params.m_tex_type = basist::basis_texture_type::cBASISTexType2D;
-    } else if(dimensions == 2) {
-        params.m_tex_type = basist::basis_texture_type::cBASISTexType2D;
-    } else {
+    } else if(dimensions == 3) {
         /* Encoding 3D images as KTX2 always produces 2D array images and mip
            levels in .basis files are inherently 2D images, so we always export
            2D array images. This affects the expected mip sizes and prevents
@@ -95,7 +90,8 @@ Containers::Array<char> convertLevelsToData(Containers::ArrayView<const BasicIma
         Warning{} << "Trade::BasisImageConverter::convertToData(): exporting 3D image as a 2D array image";
         params.m_tex_type = basist::basis_texture_type::cBASISTexType2DArray;
         mipMask[dimensions - 1] = 0;
-    }
+    } else
+        CORRADE_INTERNAL_ASSERT_UNREACHABLE(); /* LCOV_EXCL_LINE */
 
     const auto baseSize = imageLevels.front().size();
     const UnsignedInt numMipmaps = Math::min<UnsignedInt>(imageLevels.size(), Math::log2((baseSize*mipMask).max()) + 1);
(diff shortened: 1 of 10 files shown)


src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
```diff
@@ -125,13 +126,31 @@ The @ref PixelFormat::R8Unorm, @relativeref{PixelFormat,R8Srgb},
...
+1D images will be saved as 2D images with height 1.
+
+Cube map images can be written but there is currently no way to mark them
+properly in the metadata. Exported files will be 2D array images with faces
+exposed as layers.
+
+3D images will be saved as 2D array images with one layer per depth slice. This
+is necessary because mip levels are always 2-dimensional in Basis compressed
+images. Note that this matches @ref BasisImporter behavior which imports all 3D
+images as 2D array images anyway because of the same Basis limitation.
```
mosra, line 140: I'd simplify this a bit:

```suggestion
The exporter can save 2D and 2D array images. Even though the KTX container
format supports 1D and 3D images, Basis Universal doesn't.

Cube map images can be written but there is currently no way to mark them
properly in the metadata. Exported files will be 2D array images with faces
exposed as layers.
```

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
```diff
@@ -125,13 +126,31 @@ The @ref PixelFormat::R8Unorm, @relativeref{PixelFormat,R8Srgb},
...
+
 @subsection Trade-BasisImageConverter-behavior-multilevel Multilevel images
 
 Images can be saved with multiple levels by using the list variants of
 @ref convertToFile() / @ref convertToData(). Largest level is expected to be
-first, with each following level having width and height divided by two,
+first, with each following level having width, height and depth divided by two,
 rounded down. Incomplete mip chains are supported.
```
mosra, line 147: Wait, depth also? :eyes: 

```suggestion
first, with each following level having width and height divided by two,
rounded down. Because only 2D array images are supported, depth has to have the
same size in all levels. Incomplete mip chains are supported.
```

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
```diff
@@ -125,13 +126,31 @@ The @ref PixelFormat::R8Unorm, @relativeref{PixelFormat,R8Srgb},
...
 
+Due to the way @ref Trade-BasisImageConverter-behavior-types "non-trivial image types"
+are handled, the level sizes are always expected to match the resulting image
+type. Specifically that means 1D array images and 3D images with multiple
+levels can currently not be exported and produce a level size mismatch error.
```
mosra, line 152: I'd drop this paragraph altogether, it's already clear from above.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
```diff
@@ -218,8 +237,17 @@ class MAGNUM_BASISIMAGECONVERTER_EXPORT BasisImageConverter: public AbstractImag
 
     private:
         MAGNUM_BASISIMAGECONVERTER_LOCAL ImageConverterFeatures doFeatures() const override;
+
+        MAGNUM_BASISIMAGECONVERTER_LOCAL Containers::Array<char> doConvertToData(Containers::ArrayView<const ImageView1D> imageLevels) override;
         MAGNUM_BASISIMAGECONVERTER_LOCAL Containers::Array<char> doConvertToData(Containers::ArrayView<const ImageView2D> imageLevels) override;
+        MAGNUM_BASISIMAGECONVERTER_LOCAL Containers::Array<char> doConvertToData(Containers::ArrayView<const ImageView3D> imageLevels) override;
+
+        template<UnsignedInt dimensions>
+        bool convertLevelsToFile(const Containers::ArrayView<const BasicImageView<dimensions>> imageLevels, const Containers::StringView filename);
```
mosra, line 246: Minor note, just for consistency: I always put `template<>` on the same line as the signature as I consider it a part of the signature.

src/MagnumPlugins/KtxImageConverter/Test/KtxImageConverterTest.cpp
```diff
@@ -197,7 +217,7 @@ const struct {
     {"ETC2", "1d-compressed-etc2.ktx2", CompressedPixelFormat::Etc2RGB8Srgb, {7}}
 };
 
-const struct {
+constexpr struct {
```
mosra, line 220: FYI I don't really bother with marking static data as `constexpr` in tests anymore, because I have a hunch that it might have a negative effect on compile times, due to the compiler having to execute all that C++ code through a slow interpreter. (And it causes nasty issues or crashes in older compilers.) For tests I don't care much about execution startup time, iteration time is a bigger priority.

Didn't really have time to measure the impact yet, but for new code -- unless the `constexpr` is needed to actually test a `constexpr` feature -- I just slap on a `const` and that's it. If there's already a `constexpr` and it doesn't cause issues I leave it as-is.

mosra (commented):
Thank you!

src/MagnumPlugins/KtxImageConverter/Test/KtxImageConverterTest.cpp
pezcode: Noted for next time. The Ktx{Importer, ImageConverter} and CgltfImporter tests are all plastered with `constexpr`, I figure a concerted update for all of those in the future makes more sense than reverting the change just here.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
pezcode: I fixed it, was an issue across multiple plugins (that I all touched :eyes:) but it didn't come up in previous reviews. Noted for the future.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -376,7 +404,38 @@ Containers::Array<char> BasisImageConverter::doConvertToData(Containers::ArrayVi
...
+        _format = {}; /* Overridable by openFile() */
+}
+
+ImageConverterFeatures BasisImageConverter::doFeatures() const {
+    return ImageConverterFeature::ConvertLevels1DToData | ImageConverterFeature::ConvertLevels2DToData | ImageConverterFeature::ConvertLevels3DToData;
+}
+
+Containers::Array<char> BasisImageConverter::doConvertToData(Containers::ArrayView<const ImageView1D> imageLevels) {
+    return convertLevelsToData(imageLevels, configuration(), flags(), _format);
+}
```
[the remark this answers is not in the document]
pezcode: It wasn't too much work luckily, but yeah, would've made sense to check back with you. My bad 🙊 I didn't realize you can cast image views to higher dimensions, that also simplifies my `StridedArrayView` struggles into a single cast.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
```diff
@@ -218,8 +229,14 @@ class MAGNUM_BASISIMAGECONVERTER_EXPORT BasisImageConverter: public AbstractImag
 
     private:
         MAGNUM_BASISIMAGECONVERTER_LOCAL ImageConverterFeatures doFeatures() const override;
+
         MAGNUM_BASISIMAGECONVERTER_LOCAL Containers::Array<char> doConvertToData(Containers::ArrayView<const ImageView2D> imageLevels) override;
+        MAGNUM_BASISIMAGECONVERTER_LOCAL Containers::Array<char> doConvertToData(Containers::ArrayView<const ImageView3D> imageLevels) override;
+
+        template<UnsignedInt dimensions> bool convertLevelsToFile(const Containers::ArrayView<const BasicImageView<dimensions>> imageLevels, const Containers::StringView filename);
```
mosra, line 236: Last thing: 

```suggestion
        MAGNUM_BASISIMAGECONVERTER_LOCAL template<UnsignedInt dimensions> bool convertLevelsToFile(const Containers::ArrayView<const BasicImageView<dimensions>> imageLevels, const Containers::StringView filename);
```

mosra (commented):
Thank you! This looks great now :+1: 

TODO for me post-merge: make AnyImageConverter aware of Basis in 3D image export.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
pezcode: Done, but GCC doesn't seem to like it before the `template<>`, only works after it or after the return value type.

mosra: Ah, yes, after the template, sorry. Attribute placement rules are crazy.

mosra:
Thank you! Merged as f9e10f4924bda41e9ec757441bc1df527202e290...2828427c0aa7fcf13518d26e7f19891e0b5c3a0a, I shuffled the commits around a bit to erase the temporary 1D image support from the history.

PR #119, opened 2022-01-31
Basis plugins: support basis_universal 1.50

pezcode:
Just a few minor changes to make `BasisImporter` and `BasisImageConverter` work with the recent ~~[1.16 release](https://github.com/BinomialLLC/basis_universal/releases/tag/1.16)~~ [1.50 release](https://github.com/BinomialLLC/basis_universal/releases/tag/v1_50_0_2).

Nothing major changed, just a few small API differences. Since 1.15 basis has version defines we can use to `#ifdef` around this stuff, makes things pretty simple 👍 

OpenCL works surprisingly well with the automatic fallback to CPU, with nice speedups at very minimal error increase (had to adapt the test error margins for it). What I don't like is the hard dependency, making portability an issue. I would've preferred it being loaded dynamically with `dlopen()` or similar.

The global selector palette code options were removed completely since apparently they've been a no-op since at least version 1.15 (the minimum version we support). See https://github.com/BinomialLLC/basis_universal/wiki/Release-Notes#12822-v116-release-notes.

~~I'm leaving this as a draft because I assume there'll still be a few fixes and merged PRs in basis_universal in the near future that could affect some of the workarounds we currently need. But maybe somebody else has a use for this (and could try and report back! 😊).~~ It's been over two years heh

### Todos:
- [x] Figure out how to make OpenCL dependency optional
- [x] Use all supported versions across different CIs
- [x] GltfSceneConverterTest compares against a .glb with an embedded Basis-encoded .ktx2, byte-by-byte. Needs some fuzzy comparison to make up for different output with different Basis versions/settings.


mosra:
Thank you!

> being loaded dynamically with dlopen()

Crazy thinking -- since the plugin itself is dlopen()'d, it should be technically possible to build a `BasisOpenCLImageConverter` variant of the plugin that has OpenCL enabled, have `BasisImageConverter` the CPU-only variant (with both sharing the same source, just passing a different define and resulting in a differently named module). And then *somehow* make the OpenCL variant the preferred one if it manages to work on the target machine, or maybe not, and leave it up to the user to select it explicitly the same way as multithreading has to be enabled explicitly.

> The global selector palette code options were removed completely

I suppose I could merge this bit already and left the rest of the PR for later -- can you separate it into a dedicated commit?

Interesting bit from the changelog (for 1.15, though):

> KTX2 1D textures are not supported yet

I suppose if/when that gets done, we could revive the code from #118 for that? As in, keep the PR branch in your repo for later, don't delete it :)

The problem described in #115 still persists, no change to `#include "../zstd/zstd.h". I should probably open an issue for that, then.

pezcode:
> can you separate it into a dedicated commit?

Will do!

> keep the PR branch in your repo for later, don't delete it :)

I don't ever delete branches for exactly this reason 😄 

> The problem described in #115 still persists, no change to `#include "../zstd/zstd.h"`. I should probably open an issue for that, then. Alongside with an ability to cut out the image loading/saving code we don't use.

They already seem to be on the roadmap:
zstd: https://github.com/BinomialLLC/basis_universal/pull/228
images: https://github.com/BinomialLLC/basis_universal/issues/269

mosra:
The first commit is now in master (as 90e3b3d14604b867caa8360b7291aa9b679d8e56), thanks.

Thanks for looking up the issues, apparently I stopped following the repo quite a while back and wasn't aware of the new stuff happening in there.

pezcode:
Soooo, Basis just got updated to 1.50 and the changeset to get it compiling again is minimal, so I'll push that into this MR as well. Neither of the two issues above (zstd + useless image loading code) got fixed, unfortunately.

The only open question here is how to handle the OpenCL dependency. Do you still want to pursue the idea of compiling two versions? If yes, how would that be exposed in CMake? Two separate targets? Two separate `MAGNUM_WITH_` options?

(I also started working on adding support for the new UASTC HDR, but that'll be a different MR. Got it all compiling but still needs proper tests 🐱)

modules/FindBasisUniversal.cmake
```diff
@@ -188,7 +188,11 @@ foreach(_component ${BasisUniversal_FIND_COMPONENTS})
                     lodepng.cpp
                     # Added in 1.16
                     basisu_opencl.cpp
-                    pvpngreader.cpp)
+                    pvpngreader.cpp
+                    # Added in 1.50
+                    basisu_astc_hdr_enc.cpp
+                    3rdparty/android_astc_decomp.cpp
+                    3rdparty/tinyexr.cpp)
```
mosra, line 195: Haha oh god, tinyexr. At this point I should really start shipping my own dummy stubs for all these...

mosra (commented):
Sorry, I didn't even know this was still open :sweat_smile: 

I fixed the zstd thing in 7f2941027d096608e3d84c1b746637be1be341d7 back in September 2023, so that's just OpenCL left. And I think what you did here (i.e., compiling with it supported if available, and not using it if not) is the way to go. One can then always call cmake with `-DCMAKE_DISABLE_FIND_PACKAGE_OpenCL=ON`. Can you add that bit to the docs?

I'd like to ensure that at least one of the CI jobs runs w/o OpenCL, with original error margins. I think that'd be the AppVeyor jobs, or if not then Emscripten at least (so, some extra ifdef in the test...).

The image importer stubs could be done in a similar way as I "fixed" zstd. Feel free to try that, or I can do this myself. (I spent the last 3 days fighting doxygen, so this library is quite manageable in comparison.)

pezcode:
I reverted the generic comparison threshold bumps, and added a separate test case for converting with OpenCL enabled. Since this requires no real extra setup in the plugin code, I'd say it's OK to only test one file here?

Also added the stubs, was easier than expected. I didn't test with a pre-compiled Basis, but I assume the duplicate symbols are simply ignored during linking. Or does that only hold true for static libs? Then it might need some extra indirection, rather than just adding it to the plugin CMake target source set.

All CI builds now use the latest version, I figured it's best to test that. Or do you want mixed versions?

MSVC 2017 now compiles with `CMAKE_DISABLE_FIND_PACKAGE_OpenCL` although I'm not even sure if OpenCL is available there to begin with 🐵

src/MagnumPlugins/BasisImageConverter/CMakeLists.txt
```diff
@@ -64,6 +64,10 @@ target_include_directories(BasisImageConverter PUBLIC
     # Turns #include "../zstd/zstd.h" into an actual external zstd.h include.
     # See the README in that directory for details.
     ${PROJECT_SOURCE_DIR}/src/external/basis-zstd-include-uncrapifier/put-this-on-include-path)
+# Adds empty definitions for unused image loading library functions.
+# See the README in that directory for details.
+target_sources(BasisImageConverter PRIVATE
+    ${PROJECT_SOURCE_DIR}/src/external/basis-image-loading-stubs/stubs.cpp)
```
mosra, line 70: This is great, thanks.

I'd just put this together with the above zstd workaround, let's say a dir named `basis-uncrapifier` containing both, since both are used together anyway.

modules/FindBasisUniversal.cmake
```diff
@@ -174,24 +174,19 @@ foreach(_component ${BasisUniversal_FIND_COMPONENTS})
...
                     basisu_global_selector_palette_helpers.cpp
-                    lodepng.cpp
                     # Added in 1.16
                     basisu_opencl.cpp
-                    pvpngreader.cpp
                     # Added in 1.50
                     basisu_astc_hdr_enc.cpp
-                    3rdparty/android_astc_decomp.cpp
-                    3rdparty/tinyexr.cpp)
+                    3rdparty/android_astc_decomp.cpp)
```
mosra, line 189: I was checking if this bit could be removed / stubbed as well, but I'm not sure. To me it doesn't make sense why an _encoder_ would need to decode ASTC blocks if the input is always RGBA, apart from some debug-only opt-in validation.

doc/building-plugins.dox
```diff
@@ -227,7 +227,7 @@ enable them, do the following:
 
 -   For @ref Trade::BasisImporter "BasisImporter" or
     @ref Trade::BasisImageConverter "BasisImageConverter", [download the
-    `v1_15_update2` tag of the Basis Universal repo](https://github.com/BinomialLLC/basis_universal/archive/v1_15_update2.tar.gz),
+    `v1_50_0_2` tag of the Basis Universal repo](https://github.com/BinomialLLC/basis_universal/archive/v1_50_0_2.tar.gz),
```
mosra, line 230: Since it should work with 1.15 and 1.16 as well, I'd mention those tags too, but give a link only to the latest and greatest.

Same in the plugin docs, mention that it's tested with all three.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.conf
```diff
@@ -77,6 +77,10 @@ rdo_uastc_favor_simpler_modes_in_rdo_mode=true
 ktx2_uastc_supercompression=true
 ktx2_zstd_supercompression_level=6
 
+# OpenCL acceleration. Falls back to CPU encoding if OpenCL isn't supported or
+# fails during encoding.
+use_opencl=false
```
mosra, line 82: This default makes sense given that multithreading is off by default as well, but I'm thinking I'll eventually make it multithreaded + with OpenCL by default, since it makes sense to have things fast without having to set plugin-specific opt-in values.

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
```diff
@@ -1425,6 +1429,58 @@ void BasisImageConverterTest::swizzle() {
...
+        Warning redirectWarning{&out};
+        compressedData = converter->convertToData(*original);
+    }
+
+    /* If built without OpenCL, converting falls back to CPU and still succeeds */
+    #ifndef OpenCL_FOUND
+    CORRADE_COMPARE(out.str(), "Trade::BasisImageConverter::convertToData(): OpenCL not supported, falling back to CPU encoding\n");
+    #else
+    CORRADE_COMPARE(out.str(), "");
+    #endif
```
mosra, line 1458: For easier verification of which CI jobs run with OpenCL and which not, I'd do

```suggestion
    #ifndef OpenCL_FOUND
    CORRADE_WARN("OpenCL is not available.");
    CORRADE_COMPARE(out.str(), "Trade::BasisImageConverter::convertToData(): OpenCL not supported, falling back to CPU encoding\n");
    #else
    CORRADE_INFO("OpenCL is available.");
    CORRADE_COMPARE(out.str(), "");
    #endif
```

mosra (commented):
> I didn't test with a pre-compiled Basis

I don't think _anybody_ uses a pre-compiled Basis anymore. There used to be a vcpkg package that I used as a dependency, but due to general upstream unwilligness to do _anything_ about the buildsystem it got abandoned. So it's just the sources directly compiled into the plugin now (OTOH, for example ImGui still has a pre-compiled vcpkg package, and that one I _do_ use.)

> Or do you want mixed versions?

Yes, I'd like at least one build to stay on the previous version, certain projects (such as habitat-sim) are on the older version and I want them to be able to update Magnum but keep old Basis, without breakages if I accidentally do something that doesn't work with 1.15 anymore. Basically what I commented on the other PR:

> I think e.g. the MSVC 2015/2017 could be at 1.15, 2019 and MinGW at 1.16, 2022 and clang-cl at 1.50 and then linux + mac at 1.16 / 1.50, because those are so fragmented that we want to verify that OpenCL works on all.

> `CMAKE_DISABLE_FIND_PACKAGE_OpenCL` although I'm not even sure if OpenCL is available there to begin with 🐵

Yeah, better do that on some job where you _can_ install OpenCL... There's no best option where to do that, but I think the ARM64 build could be abused for this (install the same opencl packages as on other linux jobs, but then use this option). OpenCL on ARM will then (or _might_) be verified on the Apple jobs at least, unless they axed the support completely now.

For the CI failures you'll need to add `-D__is_trivially_copyable=__has_trivial_copy` to compiler flags for all the Basis files on GCC 4.8. That's just Binomial having insufficient testing yet again.

modules/FindBasisUniversal.cmake
```diff
@@ -214,8 +254,6 @@ foreach(_component ${BasisUniversal_FIND_COMPONENTS})
                 # itself.
                 set_property(TARGET BasisUniversal::Encoder APPEND PROPERTY
                     INTERFACE_LINK_LIBRARIES BasisUniversal::Transcoder)
-                set_property(TARGET BasisUniversal::Encoder APPEND PROPERTY
-                    INTERFACE_COMPILE_DEFINITIONS "BASISU_NO_ITERATOR_DEBUG_LEVEL")
```
mosra, line 218: This needs to be still present for older versions, right? One more reason to keep MSVC 2015 / 2017 on 1.15 and 1.16.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.h
```diff
@@ -122,6 +122,16 @@ add_subdirectory(magnum-plugins EXCLUDE_FROM_ALL)
 add_dependencies(your-app MagnumPlugins::BasisImageConverter)
 @endcode
 
+When compiling Basis from source with `BASIS_UNIVERSAL_DIR`, Basis will be
+built with OpenCL support if the [OpenCL package](https://cmake.org/cmake/help/v3.1/module/FindOpenCL.html)
+is found. To build Basis without a hard dependency on OpenCL, do the following:
```
mosra, line 127: I'd say this instead, as otherwise it might look like OpenCL is possible to disable only if you're having a CMake subproject.

```
pass `-DCMAKE_DISABLE_FIND_PACKAGE_OpenCL=ON` to CMake, or 
@cmake set() @ce it before the @cmake add_subdirectory() @ce call if 
using Magnum Plugins as a subproject.
```

modules/FindBasisUniversal.cmake
pezcode: Yeah, from a quick glance it's used with the `validate` option(s), so can't be removed unfortunately.

mosra: It's not _massive_ at least :) 

Actually I wasn't aware there was such a self-contained ASTC decoding thingy, I may integrate that as another plugin next to bcdec and friends. In comparison, https://github.com/richgel999/astc_dec doesn't seem to have a HDR variant yet.

modules/FindBasisUniversal.cmake
pezcode: It's still there, just moved to a `BasisUniversalEncoder_DEFINITIONS` variable that is appended to if OpenCL is found

pezcode: Ah you mean whether they still use that at all, I'll check

pezcode: Apparently, it's not needed for 1.16 anymore, but 1.15 still needs it. I'll add a comment so it'll removed if support for 1.15 is ever purged.

pezcode:
On the finish line 🐱 

`GltfSceneConverterTest` seems to be failing because it's comparing the .glb output of KTX+Basis byte-by-byte, which is not very stable across versions (or with different options) I assume.

And then just missing the different basis versions across CIs.

mosra:
Thank you! Yeah, not really sure what to do with that... what the test wants to verify is that the correct plugin got called, and correct MIME type etc was used.

I can look at that tomorrow, maybe an option would be to export a `*.gltf` instead and do just some fuzzy check on the referenced image.

mosra:
> tomorrow

which is actually today :see_no_evil: Done in 8e2cf9a7d0f9c7119031bd187aa51f54bdac9c90, if you rebase the tests should pass now.

package/ci/circleci.yml
```diff
@@ -349,6 +353,7 @@ jobs:
     - install-cmake:
         version: "3.5.2"
     - install-basis
```
mosra, line 355: YAML is a mess, right?

```suggestion
    - install-basis:
```

Same in the other cases below, that'll make it parse correctly.

package/ci/circleci.yml
```diff
@@ -224,14 +224,17 @@ commands:
           ninja install
 
   install-basis:
+    parameters:
+      tag:
+        type: string
+        default: "v1_50_0_2"
```
mosra, line 230: I'd make the latest the default instead, and override it for older versions. So when there's a new version, one doesn't need to update several places.

package/ci/unix-desktop.sh
```diff
@@ -54,6 +54,7 @@ cmake .. \
     -DCMAKE_INSTALL_PREFIX=$HOME/deps \
     -DCMAKE_INSTALL_RPATH=$HOME/deps/lib \
     -DCMAKE_BUILD_TYPE=$CONFIGURATION \
+    -DCMAKE_DISABLE_FIND_PACKAGE_OpenCL=${DISABLE_OPENCL:-OFF} \
```
mosra, line 57: Wow, fancy bash! :scream: I never have the courage to do that myself :smile:

mosra (commented):
Two remaining minor things, and I think that's it :1st_place_medal:

Thanks for the _massive_ amount of work done on these plugins.

package/ci/appveyor-desktop.bat
```diff
@@ -65,13 +65,16 @@ cd .. && cd ..
...
+rem OpenCL is disabled on MSVC 2017 to get some test coverage for
+rem BasisImageConverter without OpenCL support compiled in.
 mkdir build && cd build || exit /b
 cmake .. ^
     -DCMAKE_BUILD_TYPE=Debug ^
     -DCMAKE_INSTALL_PREFIX=%APPVEYOR_BUILD_FOLDER%/deps ^
     -DCMAKE_PREFIX_PATH=%APPVEYOR_BUILD_FOLDER%/openal;%APPVEYOR_BUILD_FOLDER%/libwebp;%APPVEYOR_BUILD_FOLDER%/devil;C:/Tools/vcpkg/installed/x64-windows ^
     -DMAGNUM_WITH_ASSIMPIMPORTER=%EXCEPT_MSVC2015% ^
     -DMAGNUM_WITH_ASTCIMPORTER=ON ^
+    -DCMAKE_DISABLE_FIND_PACKAGE_OpenCL=%EXCEPT_MSVC2015% ^
```
mosra, line 77: This disables it for all builds except MSVC 2015 tho, not for 2017 only.

Also 2017 gets the 1.15 version now, which doesn't have OpenCL yet, so try to disable it on 2019 instead?

modules/FindBasisUniversal.cmake
```diff
@@ -161,30 +163,71 @@ foreach(_component ${BasisUniversal_FIND_COMPONENTS})
...
+                # So close, yet so far.
+                # Needs to be transitive because it's used in a header file.
+                if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "5.0")
+                    list(APPEND BasisUniversalEncoder_DEFINITIONS
+                        "__is_trivially_copyable=__has_trivial_copy")
+                endif()
+
+                # Try to find an external OpenCL library and enable support for
+                # it in basis if found.
+                find_package(OpenCL QUIET)
```
mosra, line 225: I'm thinking this should probably not be `QUIET`, same as Zstd is not `QUIET` either.

package/ci/appveyor.yml
```diff
@@ -121,10 +121,21 @@ install:
...
 
-# Basis Universal
-- set BASIS_VERSION=1_15_update2
-- IF NOT EXIST %APPVEYOR_BUILD_FOLDER%\v%BASIS_VERSION%.zip appveyor DownloadFile https://github.com/BinomialLLC/basis_universal/archive/v%BASIS_VERSION%.zip
-- 7z x v%BASIS_VERSION%.zip && ren basis_universal-%BASIS_VERSION% basis_universal
+# Basis Universal, all the recent supported versions
+- IF "%TARGET%" == "desktop" IF "%APPVEYOR_BUILD_WORKER_IMAGE%" == "Visual Studio 2022" IF "%COMPILER%" == "msvc" set BASIS_VERSION=v1_50_0_2
+- IF "%TARGET%" == "desktop" IF "%APPVEYOR_BUILD_WORKER_IMAGE%" == "Visual Studio 2019" IF "%COMPILER%" == "msvc" set BASIS_VERSION=1.16.4
+- IF "%TARGET%" == "desktop" IF "%APPVEYOR_BUILD_WORKER_IMAGE%" == "Visual Studio 2017" IF "%COMPILER%" == "msvc" set BASIS_VERSION=v1_15_update2
+- IF "%TARGET%" == "desktop" IF "%APPVEYOR_BUILD_WORKER_IMAGE%" == "Visual Studio 2015" IF "%COMPILER%" == "msvc" set BASIS_VERSION=v1_15_update2
```
mosra, line 128: MSVC 2015 doesn't have Basis built at all, IIRC because they didn't even support that compiler in the first place.

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
```diff
@@ -1425,6 +1429,61 @@ void BasisImageConverterTest::swizzle() {
...
+
+    std::ostringstream out;
+    Containers::Optional<Containers::Array<char>> compressedData;
+    {
+        Warning redirectWarning{&out};
+        compressedData = converter->convertToData(*original);
+    }
+
+    /* If built without OpenCL, converting falls back to CPU and still succeeds */
+    #ifndef OpenCL_FOUND
```
mosra, line 1454: I guess this needs another branch with `#if BASISU_VERSION < 115`, with `CORRADE_INFO("OpenCL not used in version 1.15 yet.");`.

That's what is making the MSVC 2017 build to fail.

mosra (commented):
Was looking at the MSVC build failures, found some more bits.

The clang build failing is usual flakiness that happens from time to time. No need to worry with that.

package/ci/appveyor.yml
pezcode: I removed the exhaustive version selection (and made the latest version the default + a few specific overrides), but as before it still downloads Basis even for MSVC 2015. Making the whole unzipping portion + the string manipulation optional seemed like too much of a struggle.

package/ci/circleci.yml
pezcode: That's already latest, 1.50 😊

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
pezcode: I went the "easy" route and just added `BasisUniversalEncoder_INCLUDE_DIR` to the test so it has access to the version define in one of the Basis headers. Of course this broke on GCC < 5 because that needs the `__is_trivially_copyable=__has_trivial_copy` fix on the actual Encoder target. I fixed that by copying the INTERFACE_COMPILE_DEFINITIONS from the Encoder, but I'm wondering if this is already too ugly.

Is just linking to the Encoder target from the test the better way? I'm just a bit afraid of weird side effects this might have on coverage etc.

pezcode:
Note that Linux jobs now all install PoCL, a CPU OpenCL implementation. But that drags in a bunch of stuff and may prolong the CI package install step. Could you just double-check if that's OK or if we need another solution?

package/ci/circleci.yml
mosra: Oh, my apologies. I saw it ending with 2 and being rather long, which made me think it's that old `v1_15_update2` tag. Their versioning is a mess.

package/ci/appveyor.yml
mosra: Alright, that's fine then. Given it's `cmd.exe` it's better to not try to do too much with it :D

pezcode:
Not sure why I thought this would be easy 😹 

Basis, nice as it is, tries to fall back to a CPU OpenCL device, but then does this:

```c++
if (ret == CL_DEVICE_NOT_FOUND)
{
    // ASSERTS IN HERE
    ocl_error_printf("ocl::init: Couldn't get any GPU device ID's, trying CL_DEVICE_TYPE_CPU\n");

    ret = clGetDeviceIDs(platforms[0], CL_DEVICE_TYPE_CPU, 1, &m_device_id, &num_devices);
}
```

```c++
// THIS CAN'T BE TURNED OFF!!!!!!
// WHAT
#define BASISU_OPENCL_ASSERT_ON_ANY_ERRORS (1)

static void ocl_error_printf(const char* pFmt, ...)
{
    va_list args;
    va_start(args, pFmt);
    error_vprintf(pFmt, args);
    va_end(args);

#if BASISU_OPENCL_ASSERT_ON_ANY_ERRORS
    assert(0);
#endif
}
```

Just causing an assert + crash in debug builds with no way to turn it off 🌵

mosra:
... yeah, was just writing a comment on that. FFS.

I guess pass `-DNDEBUG` to this file alone? Because we're otherwise crashing with no fucking message whatsoever every time something slightly goes wrong.

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
mosra: Linking `Encoder` target could do quite a mess on static builds, suddenly everything being twice in there etc. 

Copying the compile definitions is fine, it's quite clean actually :)

mosra:
> Could you just double-check if that's OK

As far as I can see on Linux it's okay. APT isn't the fastest package manager, but it spends most of the time during some initial warmup step independently of how much stuff is installed. Looking at the CI times, it fluctuates anywhere between 40 seconds and three minutes, I guess depending on external factors like mirror download speed.

It'll be _much_ worse with Homebrew tho, any extra package there is extra painful due to how is that damn thing reliant on Git and Ruby ... can't you use the regular (deprecated) Apple OpenCL there? The ARM Macs should have even the GPU accessible from the CI VM, or so I heard at least.

pezcode:
I reverted the homebrew changes, I'm just not sure if a blank system has the headers and libraries required. Might still need the `opencl-headers` formula at least.

pezcode:
OK, it finds OpenCL on Mac, but fails at runtime with:

> 7: ERROR: ocl::init: Unable to get any device ID's
7: ERROR: opencl_init: Failed initializing OpenCL

Not sure what's missing... I can't find any relevant issues on the Basis repo.

The ASan failures seem to be coming from PoCL 🙄 Is there a way to selectively disable checks for specific libraries?

mosra:
Ugh, fuck all this.

1. Make the two sanitizer builds with `CMAKE_DISABLE_FIND_PACKAGE_OpenCL`, and enable it on the ARM build instead. If this would mean something leaks and gets undetected, it's not our problem, it's either Basis (which we have no way of fixing) or something inside OpenCL (which we _also_ have no way of fixing, or it's likely false positives due to `libLLVM-6.so` not being ASan-aware). There isn't any  OpenCL-specific code path in the plugin itself so it's enough to test w/o OpenCL enabled.

2. For Mac I'd keep building against the Apple implementation, because that's the likely default scenario for anyone there. I was googling a bit, and found various hints that `OpenCL.framework` is still there, but isn't really working anymore. I'd do something like

    ```cmake
    if(CORRADE_TARGET_APPLE AND OpenCL_LIBRARY MATCHES "System/Library/Frameworks/OpenCL\\.framework")
        set(_BASISIMAGECONVERTER_EXPECT_OPENCL_FRAMEWORK_FAILURE ON)
    endif()
    ```

    in CMake, propagate that variable to the test `configure.h`, and then do this in the test. Assuming it'll still fall back to the CPU implementation and produce a valid output, it's fine like this. 

    ```cpp
    {
        #ifdef _BASISIMAGECONVERTER_EXPECT_OPENCL_FRAMEWORK_FAILURE
        CORRADE_EXPECT_FAIL("Apple OpenCL implementation is used, which likely doesn't work anymore.");
        #endif
        CORRADE_COMPARE(out.str(), "");
    }
    ```

    And if people become annoyed at the warnings and the Apple OpenCL implementation is indeed unsalvageable, we could subsequently move this check directly to `FindBasisUniversal.cmake` and not even enable OpenCL on Apple at all if it's the system framework being found.

pezcode:
OK so current state:
- `linux` tests 1.16 + **disabled** OpenCL with `CMAKE_DISABLE_FIND_PACKAGE_OpenCL` (mainly for testing this interaction)
- `linux-arm64` tests 1.16 + CPU OpenCL
- `linux-static` tests 1.15
- `linux-nondeprecated` tests 1.50 + CPU OpenCL
- `linux-sanitizers`/`linux-threadsanitizer` test 1.50 + **disabled** OpenCL
- `macos` tests 1.50 + **disabled** OpenCL (testing)
- `macos-static` tests 1.16 + built-in OpenCL
- ~~MSVC 2015 tests nothing, no Basis there~~
- MSVC 2017 tests 1.15
- MSVC 2019 tests 1.16
- MSVC 2022 tests 1.50
- MSVC rt tests 1.16
- mingw tests 1.50
- clang-cl tests 1.50
- none of the Windows builds test OpenCL

pezcode:
I opened an issue for the OpenCL assert during CPU fallback: https://github.com/BinomialLLC/basis_universal/issues/378

mosra:
Not sure what is codecov reporting and why it thinks that OpenCL warning line isn't covered, but it looks good to me! Going to merge this today.

mosra:
Squashed the commit series a bit and merged as 8e2cf9a7d0f9c7119031bd187aa51f54bdac9c90...be3ce68036c36cc815dee3b27afaf12fae063bdd. Thank you!

Issue #36, opened 2017-12-15
CMake Missing: FREETYPE_LIBRARY


kklouzal:
The cmake script does not currently give the user an input to configure their freetype.lib directory location.
https://i.imgur.com/p0VYrAV.png
It's quite possible I'm doing something wrong here.

kklouzal:
I manually added the FREETYPE_LIBRARY entry in cmake-gui and pointed it to the library directory then had to set bot FREETYPE_INCLUDE_DIR_freetype2 and FREETYPE_INCLUDE_DIR_ft2build to the same directory and it moved to the next step. cmake now wants the directories for all freetype dependancies and that doesn't seem right since it's already compiled.

I'm thinking the freemake cmake script is a little wonky..
https://i.imgur.com/vFItW4O.png

mosra:
Hi, can you try removing `modules/FindFreetype.cmake` and trying again?

The local copy of the FindFreetype module attempts to detect when the library is built as static and then tries to find all the dependencies as well. But by just briefly looking at it I doubt it ever worked correctly :/ Hopefully the upstream version of FindFreetype that's bundled with CMake behaves better. But even there I'm not sure if it will be able to find everything correctly. What's the value of `FREETYPE_LIBRARY` you had to manually add?

mosra:
By the way, since you are apparently using `vcpkg` -- do you know about this? https://github.com/microsoft/vcpkg/tree/master/ports/magnum Someone contributed vcpkg packages of Corrade, Magnum and Magnum Plugins a while ago, but they are left pinned to some particular Git revision and not updated since.

kklouzal:
Yes before going down the cmake rabbit-hole here I searched vcpkg for Magnum. For a project this large with so many configuration options I feel it's wise to build it on my own and like you said the vcpkg version uses an older version of Magnum 👎 

Should I delete the `FindFreetype.cmake` file from within cmake? or from within `magnum-plugins/modules`?

mosra:
The local one from magnum-plugins, because it has the broken finding of zlib etc. Sorry for not being clear ;)

kklouzal:
I deleted the `magnum-plugins/FindFreetype.cmake` and the same error appears.

```
CMake Error at C:/Program Files/CMake/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Freetype (missing: FREETYPE_LIBRARY FREETYPE_INCLUDE_DIRS)
[...]

PR #121, opened 2022-04-11
Adding WebPImporter folder

melikebatihan:
Hey @mosra, @Squareys,

here's the work in progress for importing *.webp images with libwebp.
I'll let you know once it's ready for the first review.

Best,
Melike

Progress so far:
- All file names and declarations/instances have been changed.
- A small amount of libpng code has been replaced with (equivalent) libwebp code in WebPImporter.cpp 

**TODOs**
 - [ ] Find module for libwebp
 - [ ] Remove remaining references to libPNG/PngImporter
 - [ ] The importing code
 - [ ] Convert test files to webp
 - [ ] Adapt test


melikebatihan:
Hi @mosra, @Squareys,

in this new commit, I have made changes in accordance with my TODO list from the last commit.

**Finding libwebp Module**

I have created a FindWebP.cmake file by adapting it from other projects. It turns out the previous compiling errors were not caused by not linking libraries (they were already linked) but by libwebp header files belonging to separate libraries which had to be detected by FindWebP.cmake file individually. I am not sure if this file needs improvement or any correction.  

I have also made changes in existing CMakeLists.txt files to integrate WebPImporter in the project.
     

**Remove Remaining References to PngImporter**

I had missed some data previously, which have been replaced with equivalent WebPImporter references.  Also some naming mistakes (e.g. "Webp" instead of "WebP" have been corrected.

**The Importing Code**

I have used advanced decoding to extract uncompressed data. As far as I have understood from the information I got about webp, the color depth (bit depth) is always 8 bits per channel. Hence, I have omitted the code in PngImporter checking other bits and configuring accordingly.  

As for colour spaces, in webp it seems to be user specific, not inherent to the WebP data (bit-stream), hence not returned by WebPGetInfo (decode) API. That's why I have merely checked if the image had alpha and used colour spaces MODE_RGB or MODE_RGBA accordingly while decompressing. 

There is another feature in WebP bitstream; it can be found out from WebPGetFeatures() function, whether it is a lossy or lossless bitstream. I don't know how to approach this information, if I should add a different code for decompression of lossy images. I couldn' t find anything on documents about it except for a comment, saying that  in a lossy bitstream, pixels are stored in format YUV or YUVA depending on alpha being present. And since I couldn't see YUV or YUVA format in PixelFormat.h, I have just commented out the decoding function using YUV format.

I haven't done anything in terms of testing yet. I thought I would start that after getting a feedback on this last commit, since I am not sure if I have everything covered in WebPImporter.

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,168 @@
...
+        return Containers::NullOpt;
+    }
+
+    /* Reading the file information */
+    //WebPBitstreamFeatures info;
+    //CORRADE_ASSERT(WebPGetFeatures(webp_data.bytes, webp_data.size, &info) == VP8_STATUS_OK, "The features couldn't be retrieved.", Containers::NullOpt);
+
+    WebPDecoderConfig config;
+    CORRADE_ASSERT(WebPInitDecoderConfig(&config), "Decoder version mismatch", Containers::NullOpt);
+    CORRADE_ASSERT(WebPGetFeatures(webp_data.bytes, webp_data.size, &config.input) == VP8_STATUS_OK, "The features couldn't be retrieved.", Containers::NullOpt);
```
mosra, line 101: There's an important distinction between assertions, internal assertions and runtime errors. 

Assertions are meant to check either programmer errors (say, passing a negative value to `sqrt()`) in which case the `CORRADE_ASSERT()` macros with a descriptive message is used, or assumptions about program state that are not expected to fail, which is then checked with the `CORRADE_INTERNAL_ASSERT()` macro. As the internal assertion is not expected to fire for a library user, the macro has no message argument (in contrast it prints a file/line info which is useful for the library author). And finally, there's `*_ASSERT_OUTPUT()` macros, which don't get compiled away in builds that have assertions disabled, which is useful if you're wrapping an actual function call and not just a comparison.

Then, runtime errors are meant to check external conditions that are out of programmer control -- files not found, files being corrupted, etc. Compared to assertions, these don't abort the program, jumping into a debugger, but instead fail gracefully, printing an error message and returning a failure.

So, the above should be instead like this -- the first one isn't expected to fail as it doesn't depend on the file data (if I see correctly), and the case of a decoder version mismatch is so rare and hard to test that I don't think we should bother at this point :) Because the call needs to be done even with assertions disabled, it needs to be an `*_OUTPUT()` macro. The second one, as far as I can see, may fail if the file data is corrupted, so that should be a runtime error instead:

```cpp
    CORRADE_INTERNAL_ASSERT_OUTPUT(WebPInitDecoderConfig(&config));
    
    if(WebPGetFeatures(webp_data.bytes, webp_data.size, &config.input) != VP8_STATUS_OK) {
        Error{} << "The features couldn't be retrieved.";
        return {};
    }
```

The third assertion macro below needs the same treatment -- also a runtime error.

And a tip -- you can use `return {}` as a shorthand for error return. It'll be `false`, `NullOpt` or `nullptr` depending on what the return type is, all signaling a failure.

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /* decompression of the image */
+    WebPDecBuffer* const output = &config.output;
+    output->colorspace = colour_depth;
+    output->u.RGBA.size = size.x() * size.y() * channels;
+    Containers::Array<char> outData{NoInit, output->u.RGBA.size};
+    auto* rgba_buf = reinterpret_cast<uint8_t*>(outData.data());
+    output->u.RGBA.rgba = rgba_buf;
+    output->u.RGBA.stride = size.x() * channels;
```
mosra, line 124: `ImageData` expects row stride to be divisible by 4 by default (carried over from OpenGL, it's because such images have faster memory access). One option would be to pass a `PixelStorage{}.setAlignment(1)` to the `ImageData` constructor if stride isn't divisible by 4, but since WebP has an option to configure the stride, I think a better option would be to just make the rows four-byte aligned always:

```cpp
/* Round up to a multiple of 4 bytes */
const std::size_t stride = 4*((size.x()*channels + 3)/4); 
output->u.RGBA.size = stride*size.y();
output->u.RGBA.stride = stride;
...
```

When you get to testing, be sure to add a dedicated test with an image that has the stride not divisible by 4 (so, say, a 7x3 RGB image).

mosra (commented):
Hi, welcome and thank you for the work so far!

I commented just on the error handling and one easy-to-forget corner case, the rest looks great for an initial draft.

About lossy and YUV -- I hope there's a way to tell the decoder to always output an 8-bit RGB(A) (maybe by setting some values in `config.input` before passing it to `WebPGetFeatures()`?), or hopefully maybe it's even done like that by default? Maybe just try with a lossy file first and if the current code works, that's it, no further work needed :) In any case, when you get to writing tests, there should definitely be a test for both a lossy and lossless file.

About YUV in `PixelFormat` -- such formats are supported in Vulkan at least, but handling of those is extremely complex due to different resolution of chroma and luma planes and not what regular users would expect anyway, so unconditionally expanding them to RGB makes sense. I might have a look at this again in the future when I get to video support (where those mainly make sense), but it definitely isn't something you should worry about right now.

melikebatihan:
Thank you for the enlightening review @mosra. 

I hope I've made all necessary changes in error handling code. 

> About lossy and YUV -- I hope there's a way to tell the decoder to always output an 8-bit RGB(A) (maybe by setting some values in config.input before passing it to WebPGetFeatures()?), or hopefully maybe it's even done like that by default? Maybe just try with a lossy file first and if the current code works, that's it, no further work needed :) In any case, when you get to writing tests, there should definitely be a test for both a lossy and lossless file.

Telling the decoder which format to output is not just possible but seemingly the only way to decode. Apparently the colour depth mode is not something to be obtained from the image, rather the user decides in which mode to decode by assigning that information as a feature of config.output which is the decoder output buffer. And since webp bitstreams have always 8-bit colour depth per channel, it outputs always in 8-bit format. Thus, I made no changes in the code for decompressing: it simply decodes in 8-bit RGBA  when the bitstream has alpha, and in 8-bit RGB format when it doesn't. I have tried the code with both lossy and lossless files, it works without any problem. 

Lastly, I have added code to release memory used by WebPData structure, which I missed before. Put that code also inside the error handling blocks. Though I am not sure, if i should have done that.

> ImageData expects row stride to be divisible by 4 by default (carried over from OpenGL, it's because such images have faster memory access).

"row stride" here means stride in bytes, right? not in pixels? The webp decoder expects a stride in bytes in its configuration, so I wanted to make sure that I understand correctly.

mosra:
> it outputs always in 8-bit format [...] it works without any problem.

Okay, awesome. 

> "row stride" here means stride in bytes, right?

Yes, bytes.

> changes in error handling code.

Looks like you forgot to push those changes tho, I see only a new test file being added :)

melikebatihan:
Hi @mosra,

> Looks like you forgot to push those changes tho, I see only a new test file being added :)

Sorry, I have committed and pushed the wrong file. I corrected it now.

> Okay, awesome. (I checked the API to see what all decoding options it provides and discovered there's even a way to scale the image during import, interesting feature :D)

Should I add any other feature to the webp importer or improve it any other way?  Or shall I directly start testing?

mosra:
> Should I add any other feature

Nono, sorry, it's fine like this. Just collecting ideas for potential future additions.

melikebatihan:
> Nono, sorry, it's fine like this. Just collecting ideas for potential future additions.

Okay, thank you :) Then I can start on testing.

melikebatihan:
I have created small webp images (3x3, 7x3) in lossy and lossless format, with/without alpha value and added the first new tests:
 
    `WebPImporterTest::rgb()` 
    `WebPImporterTest::rgba()`   

The first one is complete, I thought of completing the latter after asking for a review. I have imitated the relative png importer tests. 

I have kept tests like `empty(); invalid(); openmemory(); opentwice(); importtwice()`, already present in other importers by default.  

Small changes (corrections) have also been made in `WebPImporter.cpp` in terms of memory release: 

    `WebPDataClear(&webp_data);`    

because I got an error about it (`free(): double free detected in tcache 2`) while testing. 

I am not sure how to be creative in testing, so I am open to ideas about adding new tests :)

melikebatihan:
I skipped tagging you @mosra @Squareys in my last commit, doing it now since I don't know if you still get notified on commits of a pull request without tagging or re-requesting review (which I've just found out :))

mosra:
I get notified for every push and comment after you tag me, plus I'm watching the repository so I literally see your every step :D

Just didn't have time to go through the code yet. Hopefully later today, or tomorrow.

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
```diff
@@ -0,0 +1,297 @@
...
+#include <Corrade/Utility/FormatStl.h> /** @todo remove once Debug is stream-free */
+#include <Corrade/Utility/Path.h>
+#include <Magnum/PixelFormat.h>
+#include <Magnum/Math/Color.h>
+#include <Magnum/Trade/AbstractImporter.h>
+#include <Magnum/Trade/ImageData.h>
+#include <webp/demux.h>
+#include <webp/mux.h>
+#include <webp/decode.h>
+#include <webp/encode.h>
```
Squareys, line 43: You shouldn't need any of these headers in the tests, since you're not using any webp functions directly here. They are all hidden inside WebPImporter.

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,152 @@
...
+    if(WebPGetFeatures(webp_data.bytes, webp_data.size, &bitstream) != VP8_STATUS_OK) {
+        Error{} << "The features couldn't be retrieved.";
+        WebPDataClear(&webp_data);
+        return {};
+    }
+
+    /* Image size */
+    const Vector2i size(bitstream.width, bitstream.height);
+
+    /* CHANNEL NUMBER AND PIXEL FORMAT (always 8-bit per channel in webp images) */
```
Squareys, line 110: [Corrade's Coding Style](https://doc.magnum.graphics/corrade/corrade-coding-style.html#corrade-coding-style-comments) (which also applies to Magnum) suggests some rules for comments to maintain readability.
Here for example, avoid the all-uppercase letters. Other comments should start with uppercase first letter, like in the beginning of a sentence.

src/MagnumPlugins/WebPImporter/WebPImporter.h
```diff
@@ -0,0 +1,147 @@
...
+namespace Magnum { namespace Trade {
+
+/**
+@brief WebP importer plugin
+
+
+@m_class{m-block m-success}
+
+@thirdparty This plugin makes use of the
+    [libPNG](http://www.libpng.org/pub/png/libpng.html) library, released under
```
Squareys, line 62: Be sure to eventually update this documentation, it's still referring to libPNG and png.

Squareys (changes requested):
Hey @melikebatihan!

Thanks for working on this, you're off to a great start! I added a few suggestions, overall, it's looks to be going in the right direction! You may want to remove some of the obsolete files, which may make the diff a bit easier to read for us for the reviews.

melikebatihan:
Hello @Squareys and @mosra,

Thank you for the last review. Here are the changes and additions in this last commit:

**CMakeLists**   

I restored the initial CMakeLists.txt as per the comment above. I had made those changes because the option

`cmake .. -DWITH_... -DBUILD_TESTS=ON`

seemed to be not working when I tried to build and run the project/executables on CLion. But now I see it works fine also after restoring the CMakeLists file :) The build error I got back then was obviously caused by another factor.

> You don't need line 27, line 28 redundantly does everything line 27 does :)

Thank you, apparently it will take a long time for me to learn CMake :) 
Here I actually thought of removing line 28 and demux library altogether because demux library is only used for decoding animated webp files but I am not sure if you would like me to add animated webp decoding later on.  

**FindWebP.cmake**

> Is this file from the official WebP repository? Or where was it from originally?

I have adapted FindWebP.cmake file from three different projects I've found. Initially I used the file on this link:

[https://fossies.org/linux/tiff/cmake/FindWebP.cmake](url)

Later on, I made changes on it by removing some parts and adding new blocks of code from the files on these two links:

[https://github.com/WebKit/webkit/blob/main/Source/cmake/FindWebP.cmake](url)
[https://github.com/libgd/libgd/blob/master/cmake/modules/FindWEBP.cmake](url)

Lastly, for this commit, I changed it again by removing the code about finding and comparing the version via pkg-config. 

**README.md**

In this file I included the information on how I generated input files as requested. I find its content a bit too simple, but the methods I used to created those files were simple too :) It might need improvement together with the input files maybe.

**Changes in WebPImporter & Addition of New Tests**

I included an error case in the importer source code for animated webp files and added a test case regarding it. Also tests including corrupt webp files have been added together with RGBA tests. 

**Pngimporter References & Other Changes**

As for old pngimporter files and references to pngimporter, I had already removed them but failed to commit their removal, which caused confusion I guess. I included those changes in the last commit, also updated the documentation referring to pngimporter, removed the redundant webp header files in WebPImporterTest.cpp and corrected the comments in WebPImporter.cpp according to the Corrade Coding Style and the relevant comment above.

CMakeLists.txt
```diff
@@ -87,9 +87,10 @@ option(WITH_STBTRUETYPEFONT "Build StbTrueTypeFont plugin" OFF)
 option(WITH_STBVORBISAUDIOIMPORTER "Build StbVorbisAudioImporter plugin" OFF)
 option(WITH_STLIMPORTER "Build StlImporter plugin" OFF)
 option(WITH_TINYGLTFIMPORTER "Build TinyGltfImporter plugin" OFF)
+option(WITH_WEBPIMPORTER "Build WebPImporter plugin" OFF)
 
 include(CMakeDependentOption)
-option(BUILD_TESTS "Build unit tests" OFF)
+option(BUILD_TESTS "Build unit tests" ON)
```
Squareys, line 93: There still seems to be a few changes to the main CMakeLists.txt here. Adding `WITH_WEBPIMPORTER` is great, but the others should be changed back 👍

src/MagnumPlugins/WebPImporter/Test/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+    INPUT ${CMAKE_CURRENT_BINARY_DIR}/configure.h.in)
+
+corrade_add_test(WebPImporterTest WebPImporterTest.cpp
+    LIBRARIES Magnum::Trade
+    FILES
+        empty.webp
+        nature.png
+        tampered_penguin.webp
+        animated.webp
+        RGB_lossless.webp
```
Squareys, line 51: From the naming convention in other tests, these should be `rgb-lossless.webp`, `tampered-penguin.webp` for example, also called [`kebap-case`](https://betterprogramming.pub/string-case-styles-camel-pascal-snake-and-kebab-case-981407998841) :)

These conventions make it easier to decide on how to name files and remember/guess how they were named.

Here's a unix command for linux/macos/msys (git bash on windows) to rename all the webp files:
```sh
for f in *.webp; do mv $f "$(echo $f | tr [A-Z] [a-z] | tr _ -)"; done
```

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
```diff
@@ -0,0 +1,415 @@
...
+#include <Corrade/Containers/StridedArrayView.h>
+#include <Corrade/Containers/String.h>
+#include <Corrade/TestSuite/Tester.h>
+#include <Corrade/TestSuite/Compare/Container.h>
+#include <Corrade/Utility/Algorithms.h>
+#include <Corrade/Utility/DebugStl.h> /** @todo remove once Debug is stream-free */
+#include <Corrade/Utility/FormatStl.h> /** @todo remove once Debug is stream-free */
+#include <Corrade/Utility/Path.h>
+#include <Magnum/PixelFormat.h>
+#include <Magnum/Math/Color.h>
```
Squareys, line 38: Color(3/4) isn't used in this file, I believe, so you can remove this commit. You could also check if the other includes are required or if some can be removed :)

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,155 @@
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019,
+                2020, 2021, 2022 Vladimír Vondruš <mosra@centrum.cz>
```
Squareys, line 5: You could also add yourself here (and in other files you changed)

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,155 @@
...
+        _in = Containers::Array<char>{NoInit, data.size()};
+        Utility::copy(data, _in);
+    }
+}
+
+UnsignedInt WebPImporter::doImage2DCount() const { return 1; }
+
+Containers::Optional<ImageData2D> WebPImporter::doImage2D(UnsignedInt, UnsignedInt) {
+
+    /* Create the structure that contains the WebP image data. */
```
Squareys, line 82: Being very picky: usually there's no final `.` in comments in magnum

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,155 @@
...
+        return {};
+    }
+
+    /* Image size */
+    const Vector2i size(bitstream.width, bitstream.height);
+
+    /* Channel number and pixel format (always 8-bit per channel in webp images) */
+    /* lossy or lossless bitstream without alpha transparency. */
+    int channels = 3;
+    PixelFormat pixel_format = PixelFormat::RGB8Unorm;
```
Squareys, line 117: The [Corrade Coding Style](https://doc.magnum.graphics/corrade/corrade-coding-style.html) suggests naming these variables `pixelFormat` and `colorDepth`. Also check elsewhere, `webpData` and some more examples should also be renamed.

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    /* decompression of the image */
+    if(WebPDecode(webp_data.bytes, webp_data.size, &config) != VP8_STATUS_OK) {
+        Error{} << "Image couldn't be decompressed. Might be a corrupt file.";
+        return {};
+    }
+
+    return Trade::ImageData2D{pixel_format, size, std::move(outData)};
+}
+}}  // end namespace Trade and namespace Magnum
```
Squareys, line 152: This comment doesn't add much here, but would require updating if the namespace is ever changed (i.e. adds a tiny bit of maintenance overhead), so I would remove it.

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,155 @@
...
+    output->u.RGBA.stride = stride;
+    output->colorspace = colour_depth;
+
+    /* create external memory pointed by output buffer */
+    Containers::Array<char> outData{NoInit, output->u.RGBA.size};
+    auto* rgba_buf = reinterpret_cast<uint8_t*>(outData.data());
+    output->u.RGBA.rgba = rgba_buf;
+    output->is_external_memory = 1;
+
+    /* decompression of the image */
```
Squareys, line 144: See [Corrade Coding Style - Comments](https://doc.magnum.graphics/corrade/corrade-coding-style.html#corrade-coding-style-comments), comments should start with a uppercase character, like for the beginning of a sentence in normal writing.

Squareys (changes requested):
Hey @melikebatihan!

It's starting to look pretty great! I added a few more comments, after which, let's get a review from @mosra.

Regarding the animation support, no need to do that for this initial PR 👍 

The Continuous Integration (CI) is currently failing on most platforms, maybe have a look at the logs and check if you can fix some issues there. You will probably need to add to the CI scripts in `package/ci` such that it downloads and builds the libwebp source code.
After that, some might be failing due to differences in compiler (GCC vs clang vs MSVC).

Best, Jonathan

src/MagnumPlugins/WebPImporter/Test/CMakeLists.txt
melikebatihan: I've changed the variable and file names, as well as commenting style after thoroughly reading Corrade Coding Style a couple of times but all those rules are a bit overwhelming and will take time to get used to :) So, I might have made some mistakes again.

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
melikebatihan: I have removed unused includes except for those in the test file, considering the possibility that new ideas about test cases come up.

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
melikebatihan: I have added myself in the files where I did substantial changes and files directly related to WebPImporter. But I ignored the files which already existed, where I only did very small changes (e.g. renaming the importer, like "webp" instead of "png") and CMakeLists / configuration files. I wanted to ask before making additions to them too.

CMakeLists.txt
melikebatihan: But now I see I get an error for trying to change CMakeLists.txt (together with other config files) because I don't have "write access".

src/MagnumPlugins/WebPImporter/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+#   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+#   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+
+find_package(Magnum REQUIRED Trade)
+find_package(WebP REQUIRED)
+find_package(WebP REQUIRED demux)
```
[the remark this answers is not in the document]
melikebatihan: I updated this section after you informed that decoding of animated webp files wouldn't be supported in this importer, by removing line 28 instead of 27, since demux library is exclusively used for decoding animated webp files and would be redundant here.

melikebatihan:
Hello, thank you for the last review @Squareys.

Apart from the modifications I made according to your comments above, I have focused on configuration in this last commit -in order to have a better understanding, figure out why integration tests fail - and made changes in certain configuration files.

**FindMagnumPlugins.cmake** 

Here I added “WebPImporter” as a member of _MAGNUMPLUGINS_PLUGIN_COMPONENTS on line 158, so that it can be used as a static plugin or as a dependency of another plugin. 


**CircleCI Files**

- You had suggested extension of the build scripts for other platforms in `package/ci/` by adding at least Windows, MacOS and Linux. Integration tests are totally new to me :) As far as I could learn so far, the logic is setting up virtual machines and running the projects on them through `config.yml` and platform specific files in `package/ci`. I started making changes on Linux platform by adding `libwebp-dev` package to the installation of base linux on lines 318 and 349 in .circleci/config.yml file as shown below:

```c++    
steps:
    - install-base-linux:
    extra: libfaad-dev libfreetype6-dev libjpeg-dev libopenal-dev libpng12-dev libdevil-dev libharfbuzz-dev libassimp-dev wget libwebp-dev
```     

- Since the project is built on Linux platform via `package/ci/unix-desktop.sh` for CI testing, I've added `-DWITH_WEBPIMPORTER=ON`  to line 92 in this file. I also commented out multiple CMake options to skip building several plugins, intending to catch information about the cause of test failure, because the latest CI test failure on Linux platform was due to `FreeTypeFont` not being found by CMake which also led to error while building `HarfBuzzFont`:

`CMake Error at src/MagnumPlugins/HarfBuzzFont/Test/CMakeLists.txt:59 (add_dependencies):
  The dependency target "FreeTypeFont" of target "HarfBuzzFontTest" does not
  exist.`

- The root of the problem seems to be that CMake can't find freetype library in the system. This problem had also occurred for libwebp, which I solved by adding `FindWebP.cmake` file as per your suggestion. I thought of adding `FindFreetype.cmake` file, however, when I came across in the document `changelog-plugins-old.dox` that although it existed before
, it was removed because of a couple of issues:

changelog-plugins-old.dox, line 223: 
`Removed our copy of FindFreetype.cmake which attempted to link to all dependencies of FreeType if it discovered that it's 
 built as static. But it got out of sync with upstream and the detection was broken on MSVC. See` 
`[mosra/magnum-plugins#13](https://github.com/mosra/magnum-plugins/pull/13),`
`[mosra/magnum-plugins#36](https://github.com/mosra/magnum-plugins/issues/36).`


**FindWebP.cmake**

- The section related to demux library has been removed, because it is specific to decompression of animated webp files and I updated `WebPImporter/CMakeLists.txt` accordingly as I mentioned above (removing `demux` in `find_package(WebP REQUIRED demux`).

- I tried to simplify the file in general as much as possible, removed redundant lines and did research about licencing and copyright topic. There are many different opinions about this subject as far as I could gather, some argue that one should always include copyright notice, no matter how much the original code gets modified or even if a small portion of it gets used, while others say, you don't need to manage copyright notice if the original code is a custom program, resembling many others and not having any substantial creativity. I updated the copyright notice according to the the current state of the file, although I came across several very similar code which got attributed to different persons, which might be supporting the idea that this is a custom code which doesn't need copyright notice. Moreover, I also observed other `Find *.cmake` files in `magnum-plugins`, saw that they are almost exactly the same as `FindWebP.cmake` and they have been referred as part of Magnum with no other copyright attribution to any third parties. Hence, I think the copyright notice this file currently has, can (or should?) be removed.

package/ci/circleci.yml
```diff
@@ -315,7 +315,7 @@ jobs:
       CONFIGURATION: Debug
     steps:
     - install-base-linux:
-        extra: libfaad-dev libfreetype6-dev libjpeg-dev libopenal-dev libpng12-dev libdevil-dev libharfbuzz-dev libassimp-dev wget
+        extra: libfaad-dev libfreetype6-dev libjpeg-dev libopenal-dev libpng12-dev libdevil-dev libharfbuzz-dev libassimp-dev wget libwebp-dev
```
Squareys, line 318: @mosra Is it possible that additions here require approval/whitelisting from you? I believe @melikebatihan is getting configuration errors on CircleCI, forwarding:

image

src/MagnumPlugins/WebPImporter/Test/CMakeLists.txt
Squareys: No worries, these things do indeed take time :)

src/MagnumPlugins/WebPImporter/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+endif()
+target_include_directories(WebPImporter
+    PUBLIC
+        ${PROJECT_SOURCE_DIR}/src
+        ${PROJECT_BINARY_DIR}/src
+    PRIVATE
+        ${WebP_INCLUDE_DIRS})
+target_link_libraries(WebPImporter PUBLIC
+    Magnum::Trade
+    ${WebP_LIBRARIES})
```
Squareys, line 54: You can use `WebP::libwebp` instead, which would link the library and include the header automatically. You can remove line 51 (and the "PRIVATE" there) then.

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+void WebPImporterTest::importTwice() {
+    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("WebPImporter");
+    CORRADE_VERIFY(importer->openFile(Utility::Path::join(WEBPIMPORTER_TEST_DIR, "penguin-lossless.webp")));
+
+    /* Verify that everything is working the same way on second use */
+    {
+        Containers::Optional<Trade::ImageData2D> image = importer->image2D(0);
+        CORRADE_VERIFY(image);
+        CORRADE_COMPARE(image->size(), (Vector2i{386, 395}));
```
Squareys, line 401: You may want to chose a smaller version of this file, reasoning being that with the many test files in the repository, it would take a long time to clone the repository.
Maybe something like 12x11 pixels could work, you can just rescale this one. (Maybe wait for input from @mosra, he might have a clearer idea of what this image should be)

Especially also nature.png

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,149 @@
...
+    if(bitstream.has_alpha) {
+        channels = 4;
+        pixelFormat = PixelFormat::RGBA8Unorm;
+        colourDepth = MODE_RGBA;
+    }
+
+    /* Structure and configuration for decoding */
+    WebPDecBuffer* const outputBuffer = &config.output;
+    const std::size_t stride = 4*((size.x()*channels + 3)/4);
+    outputBuffer->u.RGBA.size = stride * size.y();
```
Squareys, line 128:
```suggestion
    outputBuffer->u.RGBA.size = stride*size.y();
```

Minor code-style issue

Squareys (changes requested):
Added a few more minor comments. Apart from that, let's see what @mosra finds! :)

mosra (commented):
Our endless comments must be overwhelming, sorry about that :sweat_smile: I tried to focus mainly on what could be useful for your future adventures in general, so here it is.

I still need to check what's up with the CI (it worked for everyone before, so hopefully just a temporary hiccup), and I still want to get back with a suggestion about fuzzy image compare so you don't need to painstakingly write down the pixel values for lossy files.

Thank you for your patience!

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
```diff
@@ -0,0 +1,411 @@
...
+    Containers::Optional<Trade::ImageData2D> image = importer->image2D(0);
+    CORRADE_VERIFY(image);
+    CORRADE_COMPARE(image->size(), Vector2i(3, 3));
+    CORRADE_COMPARE(image->format(), PixelFormat::RGB8Unorm);
+
+    CORRADE_COMPARE_AS(image->pixels<UnsignedByte>().asContiguous(), Containers::arrayView<UnsignedByte>({
+        1, 2, 3,
+        4, 5, 6,
+        7, 8, 9
+    }), TestSuite::Compare::Container);
```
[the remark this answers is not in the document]
mosra: Ah well ... in case of PngImporter it's because it uses a greyscale file (so each pixel is 1 byte), and then the image has no row padding, so there the test can use `asContiguous()`. Here you have 3-byte RGB pixels and row padding, so you can't.

It would be probably easiest to just copy the comparison of (just lossless) `data()` from the `rgb()` test, it can't get much shorter than that anyway.

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
mosra: Yes, testing with one of the tiny images used already in the other tests above should be fine, no need to add another file just for the `openTwice()` and `importTwice()` tests.

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
```diff
@@ -0,0 +1,411 @@
...
+                '\x52', '\x52', '\xbe',
+                '\x52', '\x52', '\xbe',
+                '\x52', '\x52', '\xbe', 0, 0, 0,
+            }), TestSuite::Compare::Container);
+            break;
+
+        /* Same image with 90% image quality (lossy bitstream) */
+        case 1:
+            CORRADE_COMPARE_AS(image->data(), Containers::arrayView<char>({
+                '\x4d', '\x55', '\x24',
```
mosra, line 228: Testing for bit-exact equality of lossy-compressed images is quite a headache .. as you probably already found out :) And it's not unlikely that the output would change slightly when a new WebP version comes out.

So let's try a fuzzy comparison using [DebugTools::CompareImage](https://doc.magnum.graphics/magnum/classMagnum_1_1DebugTools_1_1CompareImage.html) instead, where you can say how much can the pixels differ from the lossless ground truth. Those thresholds are best put into the instance data to not have to branch on `testCaseInstanceId()` again:

```cpp
constexpr struct {
    const char* name;
    const char* filename;
    Float maxThreshold, meanThreshold;
} RgbData[]{
    {"lossless", "rgb-lossless.webp", 0.0f, 0.0f},
    {"lossy with 90% image quality", "rgb-lossy-90.webp", 10.0f, 2.0f},
};
```

The lossless image should be *exactly* the pixel values we expect, which is why the zeros. For the lossy I picked `10.0f` and `2.0f` just at random, when you run the test it'll tell you what's the actual threshold and a (slightly larger) value is then what should go here instead.

I also don't think we need to test that many variants, so just a losless and the 90% quality could suffice -- again, our goal is to just make sure the library is used correctly by the plugin, it's the WebP authors' job to ensure the library gives reasonable output for various quality settings.

The test would then look like this, I hope it's not much of a handholding this way :) Since it uses the DebugTools library now, you have to find it in CMake and link the test to it (look into `BasisImporter/Test/CMakeLists.txt`, there it's also being used). 

```cpp
#include <Magnum/DebugTools/CompareImage.h>

…

void WebPImporterTest::rgb() {
    auto&& data = RgbData[testCaseInstanceId()];
    setTestCaseDescription(data.name);

    Containers::Pointer<AbstractImporter> importer = _manager.instantiate("WebPImporter");
    CORRADE_VERIFY(importer->openFile(Utility::Path::join(WEBPIMPORTER_TEST_DIR, data.filename)));

    Containers::Optional<Trade::ImageData2D> image = importer->image2D(0);
    CORRADE_VERIFY(image);
    /* Format and size checked by CompareImage, the image should also have 
       four-byte aligned rows */
    CORRADE_COMPARE(image->storage().alignment(), 4);
    
    const char expected[]{
        '\x1e', '\x6e', '\x1e',
        '\x1e', '\x6e', '\x1e',
        '\x1e', '\x6e', '\x1e', 0, 0, 0,

        '\xef', '\x91', '\x91',
        '\xef', '\x91', '\x91',
        '\xef', '\x91', '\x91', 0, 0, 0,

        '\x52', '\x52', '\xbe',
        '\x52', '\x52', '\xbe',
        '\x52', '\x52', '\xbe', 0, 0, 0,
    };
    
    CORRADE_COMPARE_WITH(*image, 
        (ImageView2D{PixelFormat::RGB8Unorm, {3, 3}, expected}),
        (DebugTools::CompareImage{data.maxThreshold, data.meanThreshold}));
}
```

And similar for the RGBA test.

package/ci/android-x86.sh
```diff
@@ -140,6 +140,7 @@ cmake .. \
     -DWITH_STBVORBISAUDIOIMPORTER=OFF \
     -DWITH_STLIMPORTER=ON \
     -DWITH_TINYGLTFIMPORTER=ON \
+    -DWITH_WEBPIMPORTER=ON \
```
mosra, line 143: Getting libwebp on Android is a task that could easily take weeks, so don't even bother for now :) It can always be done later. Keep the option though but set it to `OFF`, to make it clear that it was explicitly disabled instead of forgotten:

```suggestion
    -DWITH_WEBPIMPORTER=OFF \
```

Same goes for `emscripten.sh` and `ios-simulator.sh`, those platforms don't have any easily available package either -- just explicitly disable it there, and if you added a package download for these three platforms in `circleci.yml`, remove it again. 

Building on Windows can also wait for later.

mosra (commented):
Nice, the CI runs through now! What did you have to do to fix it?

melikebatihan:
Hello @mosra, thank you very much for the useful and educative reviews. I haven't written yet any comment on my last commit about changes in program files as per your last review. I thought of writing after having some tests run successfully and I have just seen your review on tests, thank you :) I will reply to them individually.

package/ci/android-x86.sh
melikebatihan: I am happy that tests run successfully at least on Linux and macos. They were the easiest to configure with a bit of googling and looking how the configuration was done with other plugins as @Squareys suggested. The problem in the beginning was my fork on magnum-plugins not being updated. I updated my local repository by fetching latest changes and rebasing. Then I simply added libwebp/webp to the config.yml for installation on Linux and macos, and enabled the build of the WebPImporter.

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
melikebatihan: This definitely looks much cleaner and better :) 

> when you run the test it'll tell you what's the actual threshold and a (slightly larger) value is then what should go here instead.

In my first commit after determining the threshold, I got CI errors because the values were not large enough for other some Linux versions. Hence, I increased them a bit more, now they are all 0.2 larger than actual threshold.

src/MagnumPlugins/WebPImporter/Test/README.md
```diff
@@ -0,0 +1,31 @@
...
+WebP images without alpha channel (`RGB_*.webp`) have been created on Linux OS using GIMP (GNU Image
+Manipulation Program). With GIMP one can create lossless or lossy images with a range of 0-100% image quality,
+with or without transparency (alpha channel), although WebP compression algorithm might not be fruitful
+when used on very small data because some of the lossy images (eg.`RGB_lossy_0.webp`-`748 bytes`)
+in the test folder are bigger in file size than their less compressed (eg.`RGB_lossy_45.webp`-`72 bytes`)
+or even lossless equivalents (eg.`RGB_lossless.webp`-`54 bytes`).
+Here is a supporting explanation on this matter:
+https://developers.google.com/speed/webp/faq#can_a_webp_image_grow_larger_than_its_source_image
+
+WebP images with alpha channel (`RGBA_*.webp`) have been generated by compressing png images created with GIMP
```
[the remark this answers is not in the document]
melikebatihan: I have updated this document with new file names and also by deleting some information about the images I removed (like `tampered-penguin.webp` and `empty.webp` as per your suggestion about the tests.

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+constexpr struct {
+    const char* name;
+    Containers::ArrayView<const char> data; /** @todo string view once done */
+    const char* filename;
+    const char* error;
+} InvalidData[] {
+    {"invalid data", Containers::arrayView("invalid"), nullptr, "Trade::WebPImporter::image2D(): not a webp image\n"},
+    {"wrong extension", nullptr, "nature.png", "Trade::WebPImporter::image2D(): not a webp image\n"},
+    {"corrupt file", nullptr, "tampered-penguin.webp", "Image couldn't be decompressed. Might be a corrupt file.\n"},
```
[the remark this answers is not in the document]
melikebatihan: I edited this test and added the cases you suggested, also removed the tampered image/data.

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+    /* Create external memory pointed by outputBuffer buffer */
+    Containers::Array<char> outData{NoInit, outputBuffer->u.RGBA.size};
+    auto* rgbaBuffer = reinterpret_cast<uint8_t*>(outData.data());
+    outputBuffer->u.RGBA.rgba = rgbaBuffer;
+    outputBuffer->is_external_memory = 1;
+
+    /* Decompression of the image */
+    if(WebPDecode(image.bytes, image.size, &config) != VP8_STATUS_OK) {
+        Error{} << "Image couldn't be decompressed. Might be a corrupt file.";
```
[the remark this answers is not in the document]
melikebatihan: I did this also for another function (`WebPGetFeatures()`), which was checking VP8 Status. But I am not sure if it looks good using such a long switch block as duplicate code. Should I use something like `std::map` or define a function (named e.g. `statusInfo()`) containing that switch block to get rid of the duplication?

modules/FindWebP.cmake
```diff
@@ -0,0 +1,53 @@
...
+mark_as_advanced(
+   WebP_INCLUDE_DIR
+   WebP_LIBRARY)
+
+if(WebP_LIBRARY AND NOT TARGET WebP::libwebp)
+    add_library(WebP::libwebp UNKNOWN IMPORTED)
+    set_target_properties(WebP::libwebp PROPERTIES
+        IMPORTED_LOCATION "${WebP_LIBRARY}"
+        INTERFACE_INCLUDE_DIRECTORIES "${WebP_INCLUDE_DIR}")
+endif()
```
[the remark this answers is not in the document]
melikebatihan: Thank you for the information :)

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
melikebatihan: I have removed unnecessary images and scaled nature.png down to a very small image.

package/ci/android-x86.sh
melikebatihan: In my latest commit I removed all changes I made on android, ios, emscripten and appveyor configuration and set the build option to `OFF`.

melikebatihan:
I hope I made the changes correctly and didn't miss anything, especially with the coding style :)  

> The sudden CI error seems to be that you're apparently building the forked repository via your own CircleCI profile, instead of the build being done via my profile as a PR. Not sure what led to that, but I see two options:
> 
> * Either do as CircleCI tells you, go to the [Organization Security setting](https://app.circleci.com/settings/organization/github/melikebatihan/security) via the cog on the left and "Allow uncertified orbs", however cryptic that may seem.
> * Or there should be a possibility to stop building the project via your profile, if you go to the [project setting](https://app.circleci.com/pipelines/github/melikebatihan/magnum-plugins) via the cog on the right and press the big red "Stop building" button. Hopefully :D A good side-effect of this would be that it won't use up your free credits, which you may want to keep for your own projects instead.

Thank you for the advice @mosra, I did both options just in case :))

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
mosra: > the values were not large enough for other some Linux versions

That's quite interesting, and telling something about this format. When I said "it's not unlikely that the output would change slightly when a new WebP version comes out" above, I didn't expect that to happen so soon and so dramatically that it breaks even fuzzy comparison :D Other lossy codecs such as JPEG put a lot of effort into ensuring that the decoded output of the same file is bit-exact across versions, and if it changes, it's viewed as a critical bug. A bit sad to see that WebP doesn't live up to that standard. 

It's like a Butterfly Effect, even a single-bit change can result in large differences down the line if the image is further recompressed or if it's used as a source for texture baking operations. And then a minor game patch that was meant just to improve image loading times is suddenly several gigabytes large.

src/MagnumPlugins/WebPImporter/WebPImporter.cpp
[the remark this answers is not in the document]
mosra: A helper function would be a good idea, yes. Compared to a `std::map` or just about anything else, a `switch` and returning a `const char*` is the fastest way to do this. The actual branching in the switch is a code you don't ever see or step through in a debugger, so the compiler is free to do any crazy optimization it can think of. Often it ends up being compiled into some sort of a table lookup.

What's a bit hard to get right at first is the unreachable state, to avoid a compiler error saying that a `return` is missing somewhere. This should do it, if you keep the `VP8_STATUS_OK` in the switch (so all known cases are handled), but put the unreachable macro outside:

```cpp
namespace {

const char* vp8StatusCodeString(VP8StatusCode status) {
    switch(status) {
        case VP8_STATUS_OUT_OF_MEMORY: return "out of memory";
        ...
        case VP8_STATUS_OK: ;
    } 
    
    CORRADE_INTERNAL_ASSERT_UNREACHABLE();
}

}
```

Put the function just above `doImage2D()` so it's close to where it's used, the `namespace {` makes it a hidden symbol to avoid conflicting with similarly named functions that could be elsewhere.

package/ci/android-x86.sh
```diff
@@ -139,7 +139,7 @@ cmake .. \
     -DWITH_STBTRUETYPEFONT=ON \
     -DWITH_STBVORBISAUDIOIMPORTER=OFF \
     -DWITH_STLIMPORTER=ON \
-    -DWITH_TINYGLTFIMPORTER=ON \
+    -DWITH_TINYGLTFIMPORTER=OFF \
```
mosra, line 142: :eyes: 

You probably meant this instead, haha:

```suggestion
    -DWITH_TINYGLTFIMPORTER=ON \
    -DWITH_WEBPIMPORTER=OFF \
```

package/ci/appveyor-desktop.bat
```diff
@@ -142,6 +142,7 @@ cmake .. ^
     -DWITH_STBVORBISAUDIOIMPORTER=ON ^
     -DWITH_STLIMPORTER=ON ^
     -DWITH_TINYGLTFIMPORTER=ON ^
+    -DWITH_WEBPIMPORTER=ON ^
```
mosra, line 145:
```suggestion
    -DWITH_WEBPIMPORTER=OFF ^
```

I saw you temporarily added a download on Windows via vcpkg. Maybe that would have worked, but often it ends up being a real nightmare, working on some compilers and some not, and breaking every second Monday. I'll try that later, maybe. For now, to get the CI green, let's just not bother.

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+    CORRADE_COMPARE(image->storage().alignment(), 4);
+
+    /* The image has four-byte aligned rows, clear the padding to deterministic
+       values */
+    CORRADE_COMPARE(image->data().size(), 36);
+    image->mutableData()[9] = image->mutableData()[10] =
+        image->mutableData()[11] = image->mutableData()[21] =
+            image->mutableData()[22] = image->mutableData()[23] =
+                image->mutableData()[33] = image->mutableData()[34] =
+                        image->mutableData()[35] = 0;
```
mosra, line 194: This block shouldn't be needed anymore with `CompareImage`, it ignores the padding. 

(It's still needed in case of `openMemory()` tho, keep it there.)

mosra (commented):
Okay, amazing, I think we're done here :) Just three minor bits plus if you could make the error printing into a helper function as we discussed above, and then I'm happy to merge.

The last thing that still needs to be done is finding the WebP dependency in `FindMagnumPlugins`, but that's hard to get right in a PR because nothing uses the Find module here. So I'll do that post-merge, once I can test locally.

Thank you!

mosra:
Thank you! And sorry for the conflicts, had to update the CI yesterday because they deprecated old Ubuntu images...

melikebatihan:
> Thank you! And sorry for the conflicts, had to update the CI yesterday because they deprecated old Ubuntu images...

It is good practice for me to get the hang of Git :) I hope I didn't miss anything in my last commit.

src/MagnumPlugins/WebPImporter/Test/WebPImporterTest.cpp
melikebatihan: Yes, I thought it was interesting but had no idea about its consequences. Could it be prevented by forcing the installation of a specific WebP version and check via CMake configuration if the correct version is installed?

mosra: Possibly, but staying at a fixed version has other downsides such as unpatched security issues. More like just something to keep in mind, that unexplainable minor differences in rendered/baked output might be due to a WebP upgrade. A radical suggestion might be to not store your source assets in WebP and treat it only as an export format on which nothing else depends.

Vaguely related -- I came across a similar issue when upgrading a 3rd party glTF parser once, somehow it started parsing transformations slightly differently and [led to unexpected test failures](https://github.com/facebookresearch/habitat-sim/pull/496#issuecomment-587461427). Wasted quite some time trying to find the root cause.

mosra:
Even if you did, you suffered more than enough already :D My turn now, I'll do a final check locally and merge this, together with other busywork like packages, detection in `AnyImageImporter` or glTF support.

I hope you had at least some fun with this :)

melikebatihan:
> Even if you did, you suffered more than enough already :D My turn now, I'll do a final check locally and merge this, together with other busywork like packages, detection in `AnyImageImporter` or glTF support.
> 
> I hope you had at least some fun with this :)

No problem at all :D 
I had a lot of fun, thank you for helping along the way :) it became more and more interesting with what I've learned although I wish I could've allocated more time to it and learned -hence, finished the task- faster.

PR #123, opened 2022-06-13
CMake: Set plugins runtime output directory even when the rest are not set

hugoam:
Hey :)
Sibling plugins PR with this one in main Magnum repo: https://github.com/mosra/magnum/pull/570

> Hello !
> Our project only sets CMAKE_RUNTIME_OUTPUT_DIRECTORY but not the other two CMake variables, because those tend to mess with the way Visual Studio produces and consumes .lib so we don't want to open that can of worms.
>
> This wasn't an issue before as we only consumed Magnum in a prebuilt way, but now I'm including magnum as part of our CMake tree for development convenience, and I'd still like the .dll to end up in the correct plugin subfolder without having to do some ugly stuff.
>
> So this change proposes to set the RUNTIME_OUTPUT_DIRECTORY of the plugins when only CMAKE_RUNTIME_OUTPUT_DIRECTORY is set even if the other two variables are not set. In order to facilitate the change, I also factored the directory selection logic of the plugins to a single place since it was relatively complex code that was copied in more than 20 locations.
>
> Let me know what you think and if you see issues with this approach, thanks :)


mosra:
Merged as 9f37ac5c93c30823f2b0730a7810c06b30e505e6, thank you!

PR #124, opened 2022-06-14
GltfImporter: Only copy actually used vertex data ranges

hugoam:
Hello !
We are using `GltfImporter` plugin on a model such as this one: https://sketchfab.com/3d-models/issum-the-town-on-capital-isle-e433923a64d549fabb2d30635d643ab6 and we discovered the behavior of the importer on this specific file is unfortunately quite unoptimal, so this PR attempts to fix it

One characteristic of this model is that it has a single huge 12MB buffer, and all the ~600 meshes in the scene are just disjoint (very far apart) views on parts of this huge array. So for example the vertex positions of one mesh will be a small range in the beginning of this 12MB buffer and the UVs will be a small range at the end.

With the current `GtfImporter` logic, when calling the `importer->mesh(id)` for all those 600 meshes, this results in 600 `MeshData` objects where each has a copied vertexData array of about 12MB, for a total of a couple of GB, which is quite far from optimal memory usage. Looking at the code, the reason seem to be that we use a single range that encompasses all the attributes accessor ranges in the same buffer (so when those are disjoint and very far apart, we end up with huge unused buffer parts in the middle)

So this PR tries to address this by creating a number of ranges when the attribute accessors are disjoint (but we just keep one if they are interleaved/overlapping or contiguous)

Let me know your thoughts and if you think this approach for a solution is suitable 😅

### DIFF
--- a/src/MagnumPlugins/GltfImporter/GltfImporter.cpp
+++ b/src/MagnumPlugins/GltfImporter/GltfImporter.cpp
@@ -2954,8 +2954,11 @@ Containers::Optional<MeshData> GltfImporter::doMesh(const UnsignedInt id, Unsign
     UnsignedInt vertexCount = 0;
     std::size_t attributeId = 0;
     Containers::Pair<Containers::StringView, Int> lastNumberedAttribute;
-    Math::Range1D<std::size_t> bufferRange;
+    Containers::ArrayView<const char> inputVertexData;
     Containers::Array<MeshAttributeData> attributeData{uniqueAttributeCount};
+    Containers::Array<UnsignedInt> attributeRange{uniqueAttributeCount};
+    Containers::Array<Math::Range1D<std::size_t>> bufferRanges{uniqueAttributeCount};
+    UnsignedInt bufferRange = 0;
     /** @todo use suffix() once it takes suffix size and not prefix size */
     for(const Containers::Pair<Containers::StringView, UnsignedInt>& attribute: attributeOrder.exceptPrefix(attributeOrder.size() - uniqueAttributeCount)) {
         /* Duplicate attribute, skip */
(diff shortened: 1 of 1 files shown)


hugoam:
Hey, no problem, this is not urgent on our side, I was just investigating a performance topic that unveiled this issue and then I tackled it while I was at it, but I don't think we are in a big hurry.

The idea indeed floated through my mind that the exporter might give us sometimes fully owned mesh data but also sometimes views or references to an already imported chunk of mesh data, but I thought this complicates the importer API quite a bit, but I'm sure you already have ideas about how to do this.

I gave a try at a quick practical solution as I assumed something more well designed would probably take many months to design. It might be nice to expose something like the code path I wrote through an optional `compactVertexData` importer setting or something of that nature if that's something you're willing to consider. And yeah, it would be nice to separate the logic to something that is easily testable as standalone, indeed.

We'll wait for your input then :)

mosra:
Sorry that this took almost three years to fix. I took your idea, because that seems to be the best way to approach this after all, my other "simpler" attempts failed to cover all corner cases. Just further expanded it by sorting the ranges first so it does the same thing regardless of the order in which the attributes are defined. This also allowed me to remove the single buffer limitation. It's commited as 46516d152b7a02336228b2a642e55d57e4366d64, hopefully tested sufficiently enough that it won't introduce nasty regressions.

Regarding the mesh views I mentioned above, that will likely happen before this whole operation, in particular by merging accessors together if they look like being slices of the same interleaved attribute. And then this operation still needs to happen anyway. For (opt-in) memory mapping it'll be likely back to the original single buffer limitation and making just a min/max range over all accessors, but until that becomes a thing, such code doesn't need to be there at all.

PR #126, opened 2022-07-13
modules: Fix FindWebP.cmake for windows

Squareys:
Hi @mosra !

This is a very simple one, the precompiled binaries on windows are named `libwebp.lib` and similar, and on Windows, `NAMES webp` is therefore not enough, since windows has empty `CMAKE_STATIC_LIBRARY_PREFIX`.

Best,
Jonathan

### DIFF
--- a/modules/FindWebP.cmake
+++ b/modules/FindWebP.cmake
@@ -39,7 +39,7 @@
 #
 
 # Library
-find_library(WebP_LIBRARY NAMES webp)
+find_library(WebP_LIBRARY NAMES webp libwebp)
 
 # Include dir
 find_path(WebP_INCLUDE_DIR


mosra:
Ah lol once again Google's "buildsystem" that doesn't care about platform differences and uses a `lib` prefix everywhere. Thanks!

Annoying request: since you say there's a prebuilt library, could you make this be built on AppVeyor as well? (I won't get mad if you refuse.)

mosra:
Merged as 36c8af44d2b6f0d31a7f1f89c8222bc07a920aa8, and in e28d05b6a6e1ecdddad7926cdfe2a6548aed14bf I added the Windows CI build to have this covered there as well. Wasn't too bad in the end :P

Issue #127, opened 2022-08-04
AssimpImporter: Embedded textures fail to load with FBX


Squareys:
Unfortunately I cannot provide the file in question here, but some investigation resulted in finding the following comment:
https://github.com/assimp/assimp/issues/1830#issuecomment-372945539

Which seems to indicate that testing `*` prefix for the filename is no longer sufficient to detect embedded images.
Instead we would try to load every image with `scene->GetEmbeddedTexture(texture_file.C_Str())` and if it fails, load it as file.

We might fix this soon, since that is a fairly common case for scene files out in the wild.

mosra:
If you submit a PR, is it possible to craft a minimal FBX file to reproduce this? Similarly how @pezcode made the test file for material layers using some Python library.

mosra:
Fixed with #128.

PR #128, opened 2022-08-06
AssimpImporter: Fix loading of embedded textures from FBX files

Squareys:
Hi @mosra,

As promised, here's the fix for #127 .

The test file was created with Blender 3.1.0 as follows:
 - Delete default scene
 - Add plane
 - Set the diffuse color to image texture with this image: embedded
 - File > Export > FBX, chose path mode "copy" and press the button next to the dropdown, which enables embedding images
 
The file now has the absolute path of the PNG embedded (C:/Users/Squareys/Desktop/embedded.png), similar to how I've seen it in "real world files", but the image data is embedded, so the loader doesn't actually access this file ever. This was ensured by renaming the image file and opening it with the Windows 3D viewer, the texture was still correctly displayed.

The test was added first, to ensure this was actually the culprit, it fails on the `image` expression, i.e. image was not found, exactly what I was observing in #127. 

Since the `GetEmbeddedImage()` was added with 5.0.0, I kept the legacy path behind the `ASSIMP_IS_VERSION_5_OR_GREATER`, but I did not test the code (~~really hoping the CI does this, would be a hassle to set up~~ the CI is testing this, luckily).

I double checked the drive-by std::string => StringView/String changes, let's hope that won't blow up in my face like on the EmscriptenApplication 😅 

Looking forward to the review!

Best,
Jonathan


Squareys:
linux-arm64 failed with `StbResizeImageConverterTest` sefaulting 🤔

mosra:
Thanks! Especially for the extra care to put the test in an earlier commit.

Merged with one minor change -- I just made the `#ifdef` a bit less surprising in 1db20034a4ad277f102a934c679a692417aa2fb1.

Squareys:
Amazing, thank you! 🎉 🙏

PR #130, opened 2022-09-10
AssimpImporter: clean up STL types + update tests to 5.2.5

pezcode:
5.2.5 just got released and finally fixes the issues with glTF and multiple sets of vertex joint weights. I just wanted to update the tests but then got caught in a cleanup session :eyes:

Some of these changes may not be desired so I tried to keep them as granular as possible, so they can be reverted/dropped individually.

Mostly, this removes the majority of `std` types that have equivalents in `Corrade::Containers`, except for `unordered_map` and a few cases of `std::string` required by the assimp or importer API.

Tests got adapted to:
- 5.2.3 not outright failing to import glTFs with multiple skinning attribute sets anymore
- XFAIL 5.2.4 temporarily breaking skinning attributes for all file types
- 5.2.5 correctly importing *all(!)* skinning attributes, for all file types


src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1198,12 +1261,16 @@ Containers::Optional<MeshData> AssimpImporter::doMesh(const UnsignedInt id, Unsi
 }
 
 MeshAttribute AssimpImporter::doMeshAttributeForName(const Containers::StringView name) {
-    return _f ? _f->meshAttributesForName[name] : MeshAttribute{};
+    if(_f) {
```
mosra, line 1264: Since the mesh attribute names are not stored inside `_f` anymore, this branch wouldn't need to be here.

And I think it would be enough to have just `JointsAttribute` and `WeightAttribute` defined, and do the rest here with a plain old `if` cascade instead of iterating over arrays:

```cpp
    if(name == "JOINTS"_s) return JointsAttribute;
    if(name == "WEIGHTS"_s) return WeightsAttribute;
    return {};  
```

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -1198,12 +1261,16 @@ Containers::Optional<MeshData> AssimpImporter::doMesh(const UnsignedInt id, Unsi
...
+            if(attribute.first() == name) return attribute.second();
+    }
+    return MeshAttribute{};
 }
 
 Containers::String AssimpImporter::doMeshAttributeName(UnsignedShort name) {
-    return _f && name < _f->meshAttributeNames.size() ?
-        _f->meshAttributeNames[name] : "";
+    return _f && name < Containers::arraySize(MeshAttributeNames) ?
+        MeshAttributeNames[name] : "";
```
mosra, line 1273: Same here, no check for `_f` and just a plain old `if` cascade:

```cpp
    if(meshAttributeCustom(name) == JointsAttribute) return "JOINTS"_s;
    ...
```

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -96,44 +117,40 @@ using namespace Containers::Literals;
...
        imageIndex points to the (deduplicated) images array */
-    std::vector<std::tuple<const aiMaterial*, UnsignedInt, UnsignedInt>> textures;
+    Containers::Array<Containers::Triple<const aiMaterial*, UnsignedInt, UnsignedInt>> textures;
     /* (materialPointer, propertyIndexInsideMaterial) tuple defining the first
        (unique) location of an image */
-    std::vector<std::pair<const aiMaterial*, UnsignedInt>> images;
+    Containers::Array<Containers::Pair<const aiMaterial*, UnsignedInt>> images;
 
-    std::unordered_map<std::string, UnsignedInt> materialIndicesForName;
+    std::unordered_map<Containers::String, UnsignedInt> materialIndicesForName;
```
mosra, line 135: Because this one takes a `String` instead of a `StringView` and STL maps are stupid, `materialForName()` should call `find(Containers::String::nullTerminatedView(name))` instead of `find(name)` to (hopefully) avoid an unnecessary string allocation on every lookup.

I really need to make my own hashmaps already :sweat_smile:

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -96,44 +117,40 @@ using namespace Containers::Literals;
...
-    std::vector<std::size_t> meshesWithBones;
+    Containers::Array<std::size_t> meshesWithBones;
     /* For each mesh: the index of its skin (before any merging), or -1 */
-    std::vector<Int> meshSkins;
+    Containers::Array<Int> meshSkins;
     bool mergeSkins = false;
     /* For each mesh, map from mesh-relative bones to merged bone list */
-    std::vector<std::vector<UnsignedInt>> boneMap;
-    std::vector<const aiBone*> mergedBones;
+    Containers::Array<Containers::Array<UnsignedInt>> boneMap;
```
mosra, line 152: Because the `boneMap` is filled sequentially (if I see correctly), I was wondering if we could avoid the nested allocations by having just one array and an offset array to index into that? If it's not too much extra testing effort.

mosra (commented):
Ohh, this is a nice surprise cleanup. Thank you!

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -408,41 +456,51 @@ void AssimpImporter::doOpenData(Containers::Array<char>&& data, DataFlags) {
...
-            _f->materialIndicesForName[name] = i;
-        }
+        if(mat->Get(AI_MATKEY_NAME, matName) == AI_SUCCESS)
+            _f->materialIndicesForName[matName] = i;
 
         /* Store first possible texture index for this material, next textures
            use successive indices. */
         _f->textureIndices[mat] = textureIndex;
-        for(std::size_t j = 0; j != mat->mNumProperties; ++j) {
+        for(UnsignedInt j = 0; j != mat->mNumProperties; ++j) {
```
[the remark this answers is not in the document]
pezcode: Reason I changed it is because I'd have to cast to `UnsignedInt` twice for the `arrayAppend(InPlaceInit)` onto `Pair`/`Triple` arrays to make it compile at all. I'll leave it as is, but let me know if you'd prefer the casts.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: Good idea 👍  I'll check if the current tests cover this well enough, otherwise I'll extend them a bit

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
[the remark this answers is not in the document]
mosra: Ah, right, then `UnsignedInt` is the way to go. Didn't see that myself, sorry. The code style advice in this case would then be "pick a type that's the least annoying to use" :)

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
```diff
@@ -466,9 +521,10 @@ void AssimpImporter::doOpenData(Containers::Array<char>&& data, DataFlags) {
                we didn't miss anything in the root node. */
             CORRADE_INTERNAL_ASSERT(!_importer || !root->mNumMeshes);
 
-            _f->nodes.reserve(root->mNumChildren);
-            _f->nodes.insert(_f->nodes.end(), root->mChildren, root->mChildren + root->mNumChildren);
-            _f->nodeIndices.reserve(root->mNumChildren);
+            const size_t numDescendants = countNodeDescendants(root);
+            arrayReserve(_f->nodes, numDescendants);
+            arrayAppend(_f->nodes, Containers::arrayView(const_cast<const aiNode* const*>(root->mChildren), root->mNumChildren));
```
mosra, line 526: My idea here was that the extra `const` could hopefully avoid the need for `const_cast`? Or, if not working directly, at least it could work with `InPlaceInit`?

```suggestion
            arrayAppend(_f->nodes, InPlaceInit, root->mChildren, root->mNumChildren);
```

pezcode: It's using the [`ArrayView<const T>`](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Containers.html#a42ea2d796126a0411960911e18b35447) overload to append all children at once. The compiler doesn't know what to do when the view is non-const:
> none of the 2 overloads could convert all the argument types

And the two overloads it seems to consider are the ones taking a reference or an rvalue. I don't really know enough about overload resolution to understand why `arrayAppend(Array<T>&, ArrayView<const T>)` is not a candidate. But it's a general pain point with the function:
```c++
ArrayView<int> source = ...;
Array<int> destination;
// The cast is required here
Containers::arrayAppend(destination, Containers::arrayView<const int>(source));
```

pezcode: Ah, I guess the template type deduction simply fails there. What exactly does the `std::common_type<T>` do in the single-item overloads?

mosra: Mhm, yes, let me try to add `std::common_type` there as well, that could hopefully fix it. Otherwise it says "deduced conflicting type for T" and fails.

mosra: Fixed in mosra/corrade@7db68ab54666d2f83c9cc5d2775a0753677c6a09, and another annoyance in mosra/corrade@322be0685c948cc2b4a8509863925e333719fdc1. This was a massive PITA for me as well, and the fix allowed me to clean up some horrible stuff like in 373df05ffb443539ef3ff0517800cd7a863160cc. So I hope it'll work in your case as well :)

mosra:
Oh, and the stupid ARM test crash is fixed in c0337ef500efaa057c196ef22a64d03ce4e9254f. I was neglecting it because I expected some alignment issues deep inside `stb_image_resize`, but turns out it wasn't that at all. If you rebase, the build will pass and a Codecov report gets triggered properly.

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: Thanks for the fix, works great now 😌

src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
pezcode: I extended the tests a bit and I feel like I've hit a bug either with assimp or the bone merging logic in our code. Not too much time to investigate before I leave on vacation for ~2 weeks, but I'll pick this back up afterwards.

mosra: In that case, can you just push/rebase the remaining changes except this one? This particular optimization can wait two weeks, but I now have 5.2.5 on my system as well (and I expect Homebrew will soon as well) and it'd be nice to have this merged to have tests passing again. 

Thank you!

pezcode: I pushed everything except this change. https://github.com/pezcode/magnum-plugins/tree/assimp-skin-merge-better-test has the change + extended tests that fail, if you care to look at it.

mosra: Awesome, thanks :) Looked at the commit, but didn't see anything suspicious. Strange library, strange bugs.

Anyway, enjoy your vacation! :palm_tree:

pezcode:
Some Linux CIs are failing because of old GCC not liking my array initialization syntax. I feel like I've fixed this before, but before I waste CI minutes on this, do you know what's wrong there?

```c++
constexpr Containers::StringView objectNames[]{
    "Mesh_1"_s, "Mesh_2"_s, "Plane"_s
};
constexpr Containers::StringView jointNames[][2]{
    {"Node_1"_s, "Node_2"_s},
    {"Node_3"_s, "Node_4"_s}
};
```

> AssimpImporterTest.cpp:1382:5: error: array must be initialized with a brace-enclosed initializer

mosra:
Try with just const and not constexpr? That's what usually makes it go crazy.

mosra:
I squashed in some of the fixup commits and merged the commit series as 477cace0913d2f8d8526d7898e37f3314305b6b9...549486d40bd0611020d8d52ce6b143e74b594b59. Now the only remaining test failure for me locally is glslang which I have to look at, Assimp is sparkling clean :sparkles: 

Thank you!

PR #131, opened 2022-10-19
GltfImporter: fix msvc build

sthalik:


mosra:
Strange, my MSVC CIs passed even without this include, so I wonder what am I missing in my test setup...

- Anything specific with your setup?
- MSVC version?
- Compiling with the default (C++14) or a higher standard?
- What's the error message it gives otherwise?

sthalik:
Oops. It's not MSVC but native clang 15.0.2 (C++20; mingw-w64 CRT; libc++; no MSVC STL; no GNU).

The error is:

```console
c:/msys64/clang64/include/c++/v1/__algorithm/iterator_operations.h:100:92: error: no type named 'reference' in 'std::iterator_traits<Corrade::Containers::StridedIterator<1, Corrade::Containers::Pair<Corrade::Containers::BasicStringView<const char>, unsigned int>>>'
    static_assert(is_same<__deref_t<_Iter>, typename iterator_traits<__uncvref_t<_Iter> >::reference>::value,
                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
c:/msys64/clang64/include/c++/v1/__algorithm/iterator_operations.h:115:5: note: in instantiation of function template specialization 'std::_IterOps<std::_ClassicAlgPolicy>::__validate_iter_reference<Corrade::Containers::StridedIterator<1, Corrade::Containers::Pair<Corrade::Containers::BasicStringView<const char>, unsigned int>> &>' requested here
    __validate_iter_reference<_Iter>();
    ^
c:/msys64/clang64/include/c++/v1/__algorithm/unique.h:38:44: note: in instantiation of function template specialization 'std::_IterOps<std::_ClassicAlgPolicy>::__iter_move<Corrade::Containers::StridedIterator<1, Corrade::Containers::Pair<Corrade::Containers::BasicStringView<const char>, unsigned int>> &>' requested here
        *++__first = _IterOps<_AlgPolicy>::__iter_move(__i);

```

mosra:
Ah, okay. There's already one such conditional include a bit below at https://github.com/mosra/magnum-plugins/blob/6da185c7069a224b9d10a8863c4a9148d4d5df42/src/MagnumPlugins/GltfImporter/GltfImporter.cpp#L72-L76

If you change it to `#if defined(CORRADE_MSVC2015_COMPATIBILITY) || (defined(CORRADE_TARGET_MINGW) && defined(CORRADE_TARGET_LIBCXX))` instead of adding the same include unconditionally, does that work for you as well?

I still need to investigate locally if it's something that affects libc++ elsewhere, because it feels strange it would have this problem only on one platform.

sthalik:
I think it's going to happen on all platforms clang 15 + libc++ is invoked on. At least there's nothing mingw-specific.

src/MagnumPlugins/GltfImporter/GltfImporter.cpp
```diff
@@ -69,8 +69,8 @@
 #include "MagnumPlugins/GltfImporter/decode.h"
 #include "MagnumPlugins/GltfImporter/Gltf.h"
 
-#ifdef CORRADE_MSVC2015_COMPATIBILITY
-/* Otherwise std::unique() fails to compile on MSVC 2015. Other compilers are
+#if defined CORRADE_MSVC2015_COMPATIBILITY || defined CORRADE_TARGET_LIBCXX
```
mosra, line 72: I tried to reproduce with libc++ 14 (latest I have on ArchLinux), but there the compilation passed, so I think it's only happening since version 15 ([this commit in particular](https://github.com/llvm/llvm-project/commit/c9905b8cb0139f410ce63081989a328559e11374), which adds the `static_assert()`). 

So the final change would be the following -- can you confirm I didn't mess up the condition? :)

```suggestion
/* Otherwise std::unique() fails to compile on MSVC 2015 and libc++ 15 (commit
   https://github.com/llvm/llvm-project/commit/c9905b8cb0139f410ce63081989a328559e11374) */
#if defined(CORRADE_MSVC2015_COMPATIBILITY) || (defined(CORRADE_TARGET_LIBCXX) && _LIBCPP_VERSION >= 15) 
```

mosra:
Merged as c8bd027d8eab33c343a66369fef7a55862c74708, thanks!

PR #137, opened 2023-03-13
Add PBR properties

melikebatihan:
Hi @mosra, @Squareys,
This is my first commit on the task of integrating PBR materials. 

I have trouble understanding/discerning the boundaries of textures and materials, especially in the code. I added PBR properties in texture related properties but I am not sure if my approach is relevant at all for the solution of the task. Besides, I am quite confused about the definitions in assimp/material.h and not sure if I used these definitions correctly as property keys in AssimpImporter. An clarification would be great.

Another question I want to ask is the usage of the function _str(). I wonder why it is used in only some of the equality comparisons of property keys, e.g:

`key == _str(AI_MATKEY_SHININESS)`      vs.     
`key == _AI_MATKEY_TEXTURE_BASE`

### DIFF
--- a/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
+++ b/src/MagnumPlugins/AssimpImporter/AssimpImporter.cpp
@@ -1628,6 +1628,15 @@ Containers::Optional<MaterialData> AssimpImporter::doMaterial(const UnsignedInt 
                             case aiTextureType_NORMALS:
                                 attribute = MaterialAttribute::NormalTextureMatrix;
                                 break;
+                            case aiTextureType_METALNESS:
+                                attribute = MaterialAttribute::MetalnessTextureMatrix;
+                                break;
+                            case aiTextureType_BASE_COLOR:
+                                attribute = MaterialAttribute::BaseColorTextureMatrix;
+                                break;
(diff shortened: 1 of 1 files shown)


mosra:
Hi, 

thank you, and sorry for totally neglecting your other PR, #129 :sweat_smile: 

What is the intended file format this would be used for? If it's FBX or OBJ then I'd say a less painful path to success would be for me to *finally* merge #136 (sorry on that front as well), because the `UfbxImporter` has proper and non-buggy support for PBR already.

If it's not FBX/OBJ, there's #91 where I attempted to do something similar, before @pezcode went ahead and integrated most of it for custom material import in #116. Would some code or at least notes from there be salvageable? I vaguely remember the PBR support got reworked in Assimp 5.1 (?), and this PR was made before, so maybe what's there is mostly obsolete now.

mosra:
Closing due to the reasons stated above -- for glTF, FBX and OBJ there's now `GltfImporter` / `UfbxImporter` with a much more robust PBR support than what Assimp can offer. And for Assimp itself, there's the barrage of issues listed in #91, so it's unfortunately not just about recognizing new texture types but also adding a pile of format-specific workarounds to be able to know whether a diffuse texture is indeed diffuse or it's PBR base color, etc.

PR #141, opened 2023-07-25
GltfImporter: Support morph target attributes

mrxz:
This PR adds parsing of the morph targets of a mesh primitive to the `GltfImporter`. The morph target attributes are added to the `attributeOrder` collection so that they follow the same code path as other attributes. The stable sort is adjusted to first sort on morph target id followed by attribute name, ensuring only duplicates _within_ a morph target/base attributes are filtered and not _across_ morph targets.


src/MagnumPlugins/GltfImporter/Test/GltfImporterTest.cpp
```diff
@@ -4492,6 +4495,53 @@ void GltfImporterTest::meshUnorderedAttributes() {
...
+    CORRADE_COMPARE(importer->meshCount(), 1);
+
+    Containers::Optional<Trade::MeshData> mesh = importer->mesh(0);
+    CORRADE_VERIFY(mesh);
+    CORRADE_COMPARE(mesh->attributeCount(), 6);
+
+    /* Base mesh (position, normal and color) */
+    CORRADE_COMPARE(mesh->attributeCount(-1), 2);
+
+    CORRADE_VERIFY(mesh->hasAttribute(MeshAttribute::Position));
```
mosra, line 4511: It should be enough to just compare the `attributeCount()`, the `hasAttribute()` is internally just `return attributeCount(name) > 0;`.

src/MagnumPlugins/GltfImporter/Test/GltfImporterTest.cpp
```diff
@@ -4492,6 +4495,53 @@ void GltfImporterTest::meshUnorderedAttributes() {
...
+    CORRADE_VERIFY(mesh->hasAttribute(MeshAttribute::Normal));
+    CORRADE_COMPARE(mesh->attributeCount(MeshAttribute::Normal), 1);
+    CORRADE_COMPARE(mesh->attributeFormat(MeshAttribute::Normal), VertexFormat::Vector3);
+
+    /* First morph target (position and normal) */
+    CORRADE_COMPARE(mesh->attributeCount(0), 2);
+
+    CORRADE_VERIFY(mesh->hasAttribute(MeshAttribute::Position, 0));
+    CORRADE_COMPARE(mesh->attributeCount(MeshAttribute::Position, 0), 1);
+    CORRADE_COMPARE(mesh->attributeFormat(MeshAttribute::Position, 0, 0), VertexFormat::Vector3);
```
mosra, line 4524: To ensure the accessor IDs aren't accidentally mixed up, can you check also all attribute content? I.e.,

```cpp
    CORRADE_COMPARE_AS(mesh->attribute<Vector3>(MeshAttribute::Position, 0), Containers::arrayView<Vector4>({
        {0.1f, 0.2f, 0.3f},
        {0.4f, 0.5f, 0.6f},
        {0.7f, 0.8f, 0.9f}
    }), TestSuite::Compare::Container);
```

and similarly for others, including the non-morph attributes.

Ideally, to differentiate further, they would also have different contents and types, let's say `Vector3sNormalized` + `Vector3bNormalized` for the extra normals and `Vector3us` for extra positions. See the `mesh-primitives-types.gltf` and `mesh-primitives-types.bin.in` files for an example.

src/MagnumPlugins/GltfImporter/GltfImporter.cpp
```diff
@@ -3286,15 +3286,40 @@ Containers::Optional<MeshData> GltfImporter::doMesh(const UnsignedInt id, Unsign
         return {};
     }
 
+    /* Morph target attributes */
+    const Utility::JsonToken* gltfTargets = gltfPrimitive.find("targets"_s);
+    if(gltfTargets) {
+        if(!_d->gltf->parseArray(*gltfTargets)) {
+            Error{} << "Trade::GltfImporter::mesh(): invalid targets property in mesh" << _d->gltfMeshPrimitiveMap[id].first();
+            return {};
+        }
```
mosra, line 3295: Can you test these error cases as well? Those would go into the `mesh-invalid.gltf` file, with corresponding test cases added to the `MeshInvalidData` array in the test. 

It may be a bit annoying as changing the file would shift the reported line numbers, but hopefully not too tragic.

src/MagnumPlugins/GltfImporter/GltfImporter.cpp
```diff
@@ -3286,15 +3286,40 @@ Containers::Optional<MeshData> GltfImporter::doMesh(const UnsignedInt id, Unsign
...
+
+            for(Utility::JsonObjectItem gltfMorphAttribute: gltfTarget.value().asObject()) {
+                if(!_d->gltf->parseUnsignedInt(gltfMorphAttribute.value())) {
+                    Error{} << "Trade::GltfImporter::mesh(): invalid morph target attribute" << gltfMorphAttribute.key() << "in mesh" << _d->gltfMeshPrimitiveMap[id].first();
+                    return {};
+                }
+
+                arrayAppend(attributeOrder, InPlaceInit, gltfMorphAttribute.key(), gltfMorphAttribute.value().asUnsignedInt(), gltfTarget.index());
+            }
+        }
```
mosra, line 3311: The `Trade::MeshAttributeData` constructor asserts that the morph target [isn't a weights, joints or object ID attribute](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Trade.html#a8a0ec4599a941c8f155bd76efc765808), so the importer has to check that and fail the import in that case.

Best place to do that would probably be in the if cascade below that translates the string names to `MeshAttribute`.

src/MagnumPlugins/GltfImporter/GltfImporter.cpp
mosra: Can you also add a case that verifies behavior with a morph target attribute having a different component count? It'd be a variant of the `"different vertex count for each accessor"` case, and ideally the error message that's printed would also mention the morph target ID in that case, for example as `mismatched vertex count for attribute TEXCOORD_0 in morph target 1, expected 3 but got 4`.

src/MagnumPlugins/GltfImporter/Test/mesh-morph-target-attributes.gltf
```diff
@@ -0,0 +1,104 @@
+{
```
mosra, line 1: To fix the CI failures, the `mesh-morph-target-attributes.gltf` file needs to be listed in among the others in the CMakeLists.

Also, I don't think anything uses the `*.glb`, remove that one.

mosra (commented):
Thank you!

src/MagnumPlugins/GltfImporter/GltfImporter.cpp
```diff
@@ -3286,15 +3286,40 @@ Containers::Optional<MeshData> GltfImporter::doMesh(const UnsignedInt id, Unsign
...
-        [](const Containers::Pair<Containers::StringView, UnsignedInt>& a, const Containers::Pair<Containers::StringView, UnsignedInt>& b) {
-            return a.first() < b.first();
+        [](const Containers::Triple<Containers::StringView, UnsignedInt, Int>& a, const Containers::Triple<Containers::StringView, UnsignedInt, Int>& b) {
+            return a.third() != b.third() ? a.third() < b.third() : a.first() < b.first();
         },
-        [](const Containers::Pair<Containers::StringView, UnsignedInt>& a, const Containers::Pair<Containers::StringView, UnsignedInt>& b) {
-            return a.first() == b.first();
+        [](const Containers::Triple<Containers::StringView, UnsignedInt, Int>& a, const Containers::Triple<Containers::StringView, UnsignedInt, Int>& b) {
+            return a.third() == b.third() && a.first() == b.first();
         });
```
mosra, line 3323: One more thing :sweat_smile: 

> The stable sort is adjusted to first sort on morph target id followed by attribute name, ensuring only duplicates *within* a morph target/base attributes are filtered and not *across* morph targets.

Good that you said that -- can you add an explicit test for this behavior? Should be enough to expand the existing `meshDuplicateAttributes()` case.

src/MagnumPlugins/GltfImporter/GltfImporter.cpp
mrxz: I've updated it to behave similarly to the unsupported vertex format case. So it can still get imported as a custom attribute. Not entirely happy about the resulting error message, as it indicates the format being unsupported for said attribute, which isn't incorrect per se, but misleading in the sense that it could imply there exists a valid format (instead of the entire attribute not being supported in morph targets).

src/MagnumPlugins/GltfImporter/GltfImporter.cpp
mrxz: Done, updating the line numbers wasn't too bad.

mrxz:
Worth mentioning that it's quite common for morph target attributes to use sparse accessors. These aren't supported by the `GltfImporter`, but this has the peculiar side effect that .gltf models with morph targets and sparse accessor do load without this change (albeit without morph targets), but with this change will skip those meshes. Not sure if such a (pseudo-)regression is acceptable or not.

Now it would be nice if sparse accessor could be supported. But seeing the mention of memory-mapped file support, I take it that sparse accessor support doesn't really fit that picture?

src/MagnumPlugins/GltfImporter/Test/mesh-unsupported-vertex-formats.gltf
```diff
@@ -36,6 +36,22 @@
...
+    },
+    {
+      "primitives": [
+        {
+          "attributes": {
+            "POSITION": 1
+          },
+          "targets": [
+            {
+              "WEIGHTS": 4,
```
mosra, line 48: It took me a considerable amount of time to realize why the CI nondeprecated build (i.e., a build with `MAGNUM_BUILD_DEPRECATED` disabled) throws :sweat_smile: It's because the default build, with deprecated features enabled, has custom named `WEIGHTS` and `JOINTS` attributes for backwards compatibility, which were used before there was builtin skinning support. And if the deprecated features are turned off, the `WEIGHTS` and `JOINTS` are not in `_d->meshAttributesForName`, causing an exception to be thrown in `at()`.

You'd fix the consequence by changing this to the actual correct attribute name:

```suggestion
              "WEIGHTS_0": 4,
```

But the root cause is that custom attribute names in morph targets are not collected. I.e., the following (although an invalid glTF, because the morph target should reference only attributes that are actually in the mesh) would blow up too:

```json
    {
      "primitives": [
        {
          "attributes": {
            "POSITION": 1
          },
          "targets": [
            {
              "_FANCY_PROPERTY3": 4
            }
          ]
```

To fix this, the ["Go through all meshes, collect custom attributes" block](https://github.com/mosra/magnum-plugins/blob/90a0da68f17fbe7758271d64ed02c4a822a23755/src/MagnumPlugins/GltfImporter/GltfImporter.cpp#L1241-L1327) in `doOpenData()` needs to go through `"targets"` subobjects as well, not just `"attributes"`, and do basically the same. 

I think the change should be manageable by extracting the inner block that loops over the attributes into a lambda and calling it for both `"targets"` subobjects and `"attributes"`. This in turn means you move most of (all?) the new `gltf->parseWhatever()` calls and error checking directly into `doOpenData()` and in `doMesh()` you'll access the already-parsed tokens directly with `->asWhatever()`, relying on them being already checked to be correct. 

Three out of the four new failure tests thus move from `MeshInvalidData[]` into `MeshInvalidWholeFileData[]` as they now cause an error during file opening already, and from `mesh-invalid.gltf` into three new files (which means one more painful line renumbering).

For actually verifying that the custom attribute names get imported (so, that in the above snippet the `_FANCY_PROPERTY3` gets recognized and available in `importer->meshAttributeName()` and such) you'd expand the `mesh-custom-attributes.gltf` file and the `meshCustomAttributes()` case. That should be everything needed I think.

mosra:
> Now it would be nice if sparse accessor could be supported. 

Yup, I expected this to become the "next step" :)

> But seeing the mention of memory-mapped file support, I take it that sparse accessor support doesn't really fit that picture?

For starters I'd just "unsparse" the buffer first -- I assume that's how it would usually be uploaded to the GPU as well, right? That of course would mean the file contents can't be zero-copy imported from a memory-mapped anymore, but better than not having any support for sparse buffers at all. Eventually there could be dedicated builtin support in `MeshData` for such buffers (I imagine some sort of a per-attribute index buffer or something?) but so far I don't have any clear idea on how that would be done, or if it would be a good UX at all.

But before implementing support for sparse buffers I'd like to deal with the problem in https://github.com/mosra/magnum-plugins/pull/124, as those two have quite a lot of overlap. I have a semi-complete idea of how this would be done in a way that preserves both memory mappability and doesn't cause insane memory use, just need to find some time to actually implement it.

src/MagnumPlugins/GltfImporter/GltfImporter.cpp
mosra: I ended up making this an import error, as I don't expect these to be hit much, if at all, in practice, as they're disallowed by glTF anyway. Originally the unsupported vertex formats were also an import error, I just made those non-fatal after discovering there actually _are_ such assets in the wild, in particular non-normalized 16-bit vertex colors. So if something similar happens for morph targets, I can reconsider, but until then I'm keeping it simple and just failing.

The case where an unsupported format is in an otherwise supported morph target attribute (such as those non-normalized colors) is still handled as custom attribute, that stays.

mosra:
Merged, thank you! 

In addition to what I mentioned re disallowed attributes in the comment above, I expanded some tests in 4cfe3b118fc10f7bf5a73971e0a9d59db6db95d0 to test also that single uncovered line. The rest went in with no changes, thanks for top-class work :)

Is there anything else I should expected related to morph targets, like hooking up the animation support? I'll definitely let you know once I have https://github.com/mosra/magnum-plugins/pull/124 tackled so it's possible to implement sparse buffers as well.

PR #143, opened 2023-10-11
StbResizeImageConverter: update to stb_image_resize v2.04

pezcode:
Pretty straight-forward, and comes with new things!
- half-float support
- point filter

Disclaimer: Only tested on MSVC, might be worthing checking on a few other platforms with and without SIMD. It's enabled automatically, but libraries tend to get automated SIMD detection wrong. Also haven't done any benchmarking to verify their "oooo so fast :coconut:" claims.


src/MagnumPlugins/StbResizeImageConverter/StbResizeImageConverter.cpp
```diff
@@ -34,9 +35,9 @@
 #include <Magnum/Math/ConfigurationValue.h>
 #include <Magnum/Trade/ImageData.h>
 
+#define STBIR_ASSERT CORRADE_INTERNAL_ASSERT
```
mosra, line 38: To avoid accidental perf regressions in release builds, I'd make this a `CORRADE_INTERNAL_DEBUG_ASSERT`, for consistency with standard `assert()`.

src/MagnumPlugins/StbResizeImageConverter/Test/StbResizeImageConverterTest.cpp
```diff
@@ -337,6 +340,43 @@ void StbResizeImageConverterTest::rg16() {
...
+        {Math::packHalf(Math::unpack<Float, UnsignedByte>(0x33)),
+            Math::packHalf(Math::unpack<Float, UnsignedByte>(0x99))},
+        {Math::packHalf(Math::unpack<Float, UnsignedByte>(0xcc)),
+            Math::packHalf(Math::unpack<Float, UnsignedByte>(0xcc))}
+    };
+    const UnsignedShort expected[][2]{
+        /* 0xba4d (0.729412, 0.301961), 0x99c3 (0.6, 0.764706) was in the 8-bit
+           case */
+        {Math::packHalf(0.7305f), Math::packHalf(0.3005f)},
+        {Math::packHalf(0.5981f), Math::packHalf(0.7637f)}
```
mosra, line 364: Could be just

```suggestion
        {0.7305_h, 0.3005_h},
        {0.5981_h, 0.7637_h}
```

(`#include <Magnum/Math/Half.h>`). For the above you could do casts with `Half{...}` instead of `Math::packHalf(...)` to make it slightly shorter, unfortuantely I don't know anything better than that.

src/MagnumPlugins/StbResizeImageConverter/StbResizeImageConverter.cpp
```diff
@@ -138,9 +154,11 @@ Containers::Optional<ImageData3D> convertInternal(const ImageView3D& image, Util
         filter = STBIR_FILTER_CATMULLROM;
     else if(filterString == "mitchell"_s)
         filter = STBIR_FILTER_MITCHELL;
+    else if(filterString == "point"_s)
+        filter = STBIR_FILTER_POINT_SAMPLE;
     /* LCOV_EXCL_STOP */
     else {
-        Error{} << "Trade::StbResizeImageConverter::convert(): expected filter to be empty or one of box, triangle, cubicpline, catmullrom or mitchell, got" << filterString;
+        Error{} << "Trade::StbResizeImageConverter::convert(): expected filter to be empty or one of box, triangle, cubicspline, catmullrom, mitchell or point, got" << filterString;
```
mosra, line 161: Can you also add the `point` filter to the docs in the `StbResizeImageConverter.conf`? Thanks!

mosra (commented):
No way, what a surprise PR -- didn't expect this to come up in my notifications at 2 AM, thank you! :star_struck: 

The ASan test failure has me worried a bit, other than that it looks alright. For the SIMD stuff I may eventually play with building it several times with different flags and then dispatching with [Corrade::Cpu](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Cpu.html), the default build flags are usually not enough to pick up AVX2 or F16C.

Also, the new library is _big_, heh.

pezcode:
Opened an issue regarding the ASan error: https://github.com/nothings/stb/issues/1526

mosra:
Not sure if that fix was delayed by https://securitylab.github.com/advisories/GHSL-2023-145_GHSL-2023-151_stb_image_h/ in any way, worst case if nothing happens for several more days I'll just apply the patch myself here to have this merged.

pezcode:
I updated to v2.04 which fixes the ASan errors.

For some reason CircleCI timed out, not sure what to do about that :eyes:

mosra:
Oh, is it out already? I was subscribed to the issue and to the PR and of course no mention was put in any of these. Strange release management.

> CircleCI timed out

That's just Ninja running with 38 threads because it queried that wrong from the system, and then randomly crashing or OOMing or getting stuck due to not enough RAM being available. Everything is fine :fire: (I restarted the build.)

pezcode:
Is there anything else that needs to be done from my side? This would be nice to have, but it doesn't have high priority.

mosra:
Nothing needed from your side, sorry -- this fell through the cracks. Merged as 8ca90cc19f12049e12fcee07c41362cc1462a07f...e84a3fc3e0afabf36016203325534248b4bb0e41.

pezcode:
Amazing, thanks ☺️

PR #145, opened 2024-09-11
Basis plugins: support UASTC HDR

pezcode:
Adds support for HDR format encoding and transcoding to `BasisImageConverter` and `BasisImporter`. UASTC HDR support got added to Basis in version 1.50.

The limitations for HDR currently are:
- can't transcode LDR -> HDR or HDR -> LDR
- HDR alpha is always 1

Depends on https://github.com/mosra/magnum-plugins/pull/119.

### Todos:

- [x] tests
- [x] tests
- [x] did I mention tests
- [x] ~~update example code for supported format detection code~~ requires `_ldr` and `_hdr` pseudo extensions for profiles of `WEBGL_compressed_texture_astc`


src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -257,18 +341,35 @@ template<UnsignedInt dimensions> Containers::Optional<Containers::Array<char>> c
...
+    if(isHdr) {
+        #if BASISU_LIB_VERSION >= 150
+        params.m_source_images_hdr.resize(numImages);
+        if(numMipmaps > 1) {
+            params.m_source_mipmap_images_hdr.resize(numImages);
+            for(auto& slice: params.m_source_mipmap_images_hdr)
+                slice.resize(numMipmaps - 1);
+        }
+        #else
+        CORRADE_INTERNAL_ASSERT_UNREACHABLE();
```
mosra, line 362: The problem with the unreachable assert is that it will fail only at runtime if some refactoring goes wrong and `isHdr` becomes set. I think a better way would be to wrap also `if(isHdr)`, the `bool isHdr` and everything related in `#if BASISU_LIB_VERSION >= 150`. Then a lot more of the correctness is checked at compile time. Same with the unreachable assert below.

Related -- to avoid accidental breakages in existing projects that bundle older Basis, at least one CI job should stay at the old version. Can you update the configs in `package/ci` to make some stay at older and some at newer? I think e.g. the MSVC 2015/2017 could be at 1.15, 2019 and MinGW at 1.16, 2022 and clang-cl at 1.50 and then linux + mac at 1.16 / 1.50, because those are so fragmented that we want to verify that OpenCL works on all.

(I can do that myself too, but there aren't any prebuilt packages so it should be doable from your end as well).

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -46,30 +46,90 @@
...
+        if(!customSwizzle)
+            for(std::size_t y = 0; y != src.size()[0]; ++y)
+                for(std::size_t x = 0; x != src.size()[1]; ++x)
+                    dst[y][x] = Math::gather<'r', 'r', 'r'>(src[y][x]);
+        else
+            for(std::size_t y = 0; y != src.size()[0]; ++y)
+                for(std::size_t x = 0; x != src.size()[1]; ++x)
+                    dst[y][x] = Math::Vector3<T>::pad(src[y][x]); /* Alpha implicitly opaque */
+
+    } else CORRADE_INTERNAL_ASSERT_UNREACHABLE(); /* LCOV_EXCL_LINE */
```
mosra, line 105: Ah, this reminds me you requested the batch swizzle. On my list :)

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -243,6 +307,26 @@ template<UnsignedInt dimensions> Containers::Optional<Containers::Array<char>> c
...
+    if(params.m_use_opencl && !basisu::opencl_is_available()) {
+        Warning{} << "Trade::BasisImageConverter::convertToData(): OpenCL not supported, falling back to CPU encoding";
+        params.m_use_opencl = false;
+    }
+    #endif
+
+    /* HDR */
+    #if BASISU_LIB_VERSION >= 150
+    params.m_hdr = isHdr;
+    // TODO Does this enable UASTC automatically?
```
mosra, line 323: I hope? Heh.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -46,30 +46,90 @@
...
             isSrgb = true;
             break;
+        #if BASISU_LIB_VERSION >= 150
+        case PixelFormat::R32F:
+        case PixelFormat::RG32F:
+        case PixelFormat::RGB32F:
+        case PixelFormat::RGBA32F:
+            isHdr = true;
+            break;
+        #endif
```
mosra, line 132: It doesn't support half floats as input? Sad :(

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
```diff
@@ -46,30 +46,90 @@
...
     const PixelFormat pixelFormat = imageLevels.front().format();
-    bool isSrgb;
+    bool isSrgb = false;
+    bool isHdr = false;
     switch(pixelFormat) {
         case PixelFormat::RGBA8Unorm:
         case PixelFormat::RGB8Unorm:
         case PixelFormat::RG8Unorm:
         case PixelFormat::R8Unorm:
-            isSrgb = false;
```
mosra, line 65: I'd rather keep this here, to not have some of those accidentally left at default in some codepaths. Latest GCC and/or Clang warns (or they do in Release builds at least) if the variable isn't initialized in all codepaths, which is a good hint that something went wrong. With the default initialization an accident would be silent.

src/MagnumPlugins/BasisImporter/BasisImporter.h
```diff
@@ -388,6 +394,30 @@ class MAGNUM_BASISIMPORTER_EXPORT BasisImporter: public AbstractImporter {
...
+             * Uncompressed half float RGB HDR. Loaded as
+             * @ref PixelFormat::RGB16F.
+             */
+            RGB16F = 24,
+
+            /**
+             * Uncompressed half float RGBA HDR. The alpha channel is always
+             * set to opaque. Loaded as @ref PixelFormat::RGBA16F.
+             */
+            RGBA16F = 25,
```
mosra, line 420: I'd like to have at least the RGB16F / RGBA16F target tested (as usual, so a file generated by BasisImageConverter from the EXR and compared, and then a resulting KTX2/Basis imported with BasisImporter and compared). The OpenExrImporter test files contain a HDR file, it's _really tiny_ so the block compression will likely do a total mess out of it, but better than nothing. OpenExrImporter has a [forceChannelType=FLOAT](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1OpenExrImporter.html#Trade-OpenExrImporter-configuration) option to import half-floats as floats that you can use here. There's also a cubemap EXR file which is slightly bigger and could maybe get less distorted.

Or alternatively if the Basis repo contains some small enough sample HDR file with known output, use it. But I doubt that.

Besides that, BasisImporter test has `FormatData` which you could expand with the HDR formats. It doesn't verify the actual contents but could be enough to verify it doesn't enter some infinite allocation loop or whatever in these.

mosra (commented):
First of all, thanks a lot! 

Doing a _really tired_ review, so apologies if something doesn't make sense or the ramblings are too long. ... or if I'm pointing out something you mentioned in the description, such as tests. Or tests.

src/MagnumPlugins/BasisImageConverter/BasisImageConverter.cpp
pezcode: Do we want to add that to the plugin as a convenience? We're already doing a bunch of handholding that the command line utility does (but not the API), like implicit swizzling, so I think it'd make sense.

mosra: As we discussed elsewhere, I'd rather provide a dedicated utility for the most common swizzles and format conversions instead, usable with any plugins. Which will then be exposed in the `magnum-imageconverter` utility, or available to be called from C++ / Python directly.

pezcode:
Oops, I committed the basisu binary 🙈 Have to undo that

src/MagnumPlugins/BasisImporter/BasisImporter.h
pezcode: I tried to cover all affected code paths, without duplicating every single test:

Importer:
- test error when transcoding LDR `<->` HDR
- test unconfigured plugin defaulting to `RGBA16F` for HDR .basis files
- transcode .basis to `RGB16F` + compare against source RGB .exr
- transcode .basis to `RGBA16F` + compare against source RGBA .exr
- transcode .basis to BC6 and ASTC4x4RGBAF

Converter
- test mip generation with HDR input
- test patching away .basis sRGB flag for HDR input - HDR can't be sRGB
- convert a 2D HDR mip chain + compare against source .exr files
- last test also tests Basis behavior of dropping the alpha channel and making it 1 always

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -952,6 +1071,42 @@ void BasisImporterTest::rgbaUncompressedUastc() {
...
+
+    if(_manager.loadState("AnyImageImporter") == PluginManager::LoadState::NotFound)
+        CORRADE_SKIP("AnyImageImporter plugin not found, cannot test contents");
+    if(_manager.loadState("OpenExrImporter") == PluginManager::LoadState::NotFound)
+        CORRADE_SKIP("OpenExrImporter plugin not found, cannot test contents");
+
+    CORRADE_COMPARE_WITH(*image,
+        Utility::Path::join(BASISIMPORTER_TEST_DIR, "rgba-63x27.exr"),
+        /* There are moderately significant compression artifacts */
+        (DebugTools::CompareImageToFile{_manager, 0.535f, 0.12f}));
```
pezcode, line 1107: Todo - this is incorrect, because it compares the source file with varying alpha to the transcoded file with alpha always set to 1. The error thresholds are *way* too high.

mosra: You can slice just the RGB part from the file. Something like `image->pixels<Color4>().slice(&Color4::rgb)`  instead of `*image` should work, and then for the other side you can configure the importer to skip the alpha channel by setting `a=` in the config, which will not even import the alpha channel.

I think, at least.

pezcode: I tried the slice magic but it doesn't seem to work on a const view. Only with `mutablePixels()`. Is this an overload issue?

mosra: Huh. What does it say?

pezcode: It says that it can't find a suitable overload. But I figured it out, `Color4<T>::rgb() const` (same for `Vector4`) return a value, not a reference. Will just do it manually then ☺️

pezcode: I actually ended up using `mutablePixels()` with a comment explaining why, hope that's no crime in the eyes of the const-ness committee

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -657,7 +691,30 @@ template<UnsignedInt dimensions> Containers::Optional<ImageData<dimensions>> Bas
...
-        Trade::ImageData<dimensions> out{pixelFormat(*targetFormat, _state->isSrgb), Math::Vector<dimensions, Int>::pad(Vector3i{size}), Utility::move(dest), ImageFlag<dimensions>(UnsignedShort(_state->imageFlags))};
+        const Magnum::PixelFormat outFormat = pixelFormat(*targetFormat, _state->isSrgb);
+        const Math::Vector<dimensions, Int> outSize = Math::Vector<dimensions, Int>::pad(Vector3i{size});
+        /* All (compressed or uncompressed) formats have a block/pixel size
+           that's a multiple of 4, matching the PixelStorage default alignment
+           of 4. *Except* for RGB16F. Basis transcoding functions only take row
+           stride in pixels, not bytes, so we need to manually copy to a
+           correctly aligned image. */
+        const UnsignedInt rowWidthInBytes = formatSize*size.x();
+        if(rowWidthInBytes % 4 != 0) {
```
pezcode, line 702: Basis only accepts row stride in multiples of the pixel size, *not* bytes, which breaks for RGB16F. For some reason I figured an intermediate copy is better than returning an image with `PixelStorage` with alignment of 1. But I just realized some importers (Dds, Ktx) do this, so that's probably the better option?

Another alternative would be to simply not allow RGB16F transcoding. Some GPU APIs don't support this at all, but it's also weird to not have it since Basis always transcodes HDR images to alpha = 1.

mosra: I'd personally return 1-byte-aligned image, yeah. In that case the users have the option of either aligning themselves if it's actually a problem (which might not be at all, for example when saving the output as another image format) or importing as RGBA16F instead. With the current behavior they'd get an extra alloc + copy that cannot be avoided, which might be a problem with giant images on memory-constrained systems such as the web.

> Another alternative would be to simply not allow RGB16F transcoding

The output isn't always just fed to GPUs, it can be an input to some test, or it can be somewhere in the processing pipeline of an `magnum-imageconverter` invocation, and in both cases the extra flexibility is always useful.

src/MagnumPlugins/BasisImporter/BasisImporter.cpp
```diff
@@ -717,7 +776,10 @@ template<UnsignedInt dimensions> Containers::Optional<ImageData<dimensions>> Bas
                     flipped = false;
                     break;
                 /* We'd be in the isUncompressed branch above for this */
-                case TargetFormat::RGBA8: CORRADE_INTERNAL_ASSERT_UNREACHABLE(); /* LCOV_EXCL_LINE */
+                case TargetFormat::RGBA8:
+                case TargetFormat::RGB16F:
+                case TargetFormat::RGBA16F:
+                    CORRADE_INTERNAL_ASSERT_UNREACHABLE(); /* LCOV_EXCL_LINE */
```
mosra, line 782: Given that codecov complains here, it has to be this now

```suggestion
                /* LCOV_EXCL_START */
                case TargetFormat::RGBA8:
                case TargetFormat::RGB16F:
                case TargetFormat::RGBA16F:
                    CORRADE_INTERNAL_ASSERT_UNREACHABLE(); 
                /* LCOV_EXCL_STOP */
```

mosra (commented):
I'll try to do a more thorough review later today, but so far seems like I won't have much to complain about. :ok_hand:

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
mosra: > But I figured it out, `Color4<T>::rgb() const` (same for `Vector4`) return a value, not a reference.

Ahh, damn, that's because of the `constexpr` -- it seems that having the accessor constexpr was more valuable than having it sliceable. Sigh.

Yeah, `mutablePixels()` is fine. Or you could `arrayCast<const Color3h>()`, that'd achieve the same as the slice.

src/MagnumPlugins/BasisImageConverter/Test/BasisImageConverterTest.cpp
```diff
@@ -878,22 +918,107 @@ void BasisImageConverterTest::convert2DMipmaps() {
...
+        /* Drop alpha channel to only test RGB. Not using CompareImageToFile to
+           avoid overwriting the 4-channel expected .exr with a 3-channel
+           actual image when --save-diagnostic is specified. */
+        imageImporter->configuration().setValue("a", "");
+        
+        CORRADE_VERIFY(imageImporter->openFile(Utility::Path::join(BASISIMPORTER_TEST_DIR, levels[i].file)));
+        const auto expected = imageImporter->image2D(0);
+        CORRADE_VERIFY(expected);
+
+        imageImporter->configuration().setValue("a", "A");
```
mosra, line 1016: Instead of resetting the config values back, which makes it a bit uncertain what the importer is actually doing at given point, I'd create two instances, one for importing RGBA32F and one for RGB16F.

The config, if set on a plugin instance and not on the manager, is local to the instance, so they won't clash.

mosra (commented):
One last bit, other than that looks good to me (plus the 1-byte alignment from above, and `arrayCast()` instead of `mutablePixels()`).

I'm still rushing deadlines, so I won't be able to add that fake ldr/hdr extension for WebGL until later next week. So either the PR will wait, or I'll merge and the loading code will get (partially) updated later.

pezcode:
> So either the PR will wait, or I'll merge and the loading code will get (partially) updated later.

I think it's okay to do that later, otherwise this PR might get stalled out needlessly. Sometimes things come up etc. and this is just a convenient doc detail.

I reverted back to `arrayCast()` for the slicing, but my code for getting a view on the alpha channel looks terrible. Feel free to codegolf that a bit (or revert the commit) if it's worse than before 😸

mosra:
Damn, I thought it _is_ possible to slice a const view to single channels at least, sorry. I'll fix that, those getters should return a `const&`, like Pair etc does.

mosra:
Squashed a bit and merged as a18f9d87f5e31e53ca98b59788c8181f85adde06...e327796aac53a05f2c16b6394ea5e2da52954798. I reverted back to `.slice(&Color4h::a)`, as of mosra/magnum@d5b515886970d5204be84018e26a394ec2c6ef9b that works properly on const views as well. (Oh and it seems some trailing whitespace sneaked into your commits, I cleaned those up in the process but you might want to check your editor settings, or whether it could use the `.editorconfig` that's in the repository.)

In any case, thanks a lot for this!

PR #146, opened 2024-10-10
BasisImporter: add formatHdr config option

pezcode:
As discussed on Gitter. Allows opening Basis images without having to successfully guess whether it's an HDR image or not. Now you can set both and it will get transcoded successfully, without an error about mismatching LDR <-> HDR format.

I thought about splitting the HDR formats out of `TargetFormat` into their own `TargetHdrFormat` enum for easier validation and better documentation, but shied away from it because of backwards compatibility. Might be worth doing though, seeing how the HDR formats only got added very recently.

Still missing some updated docs, but code + tests should be ready for a review 😌

### Todo
- [x] Update docs to mention `formatHdr` and `setTargetFormatHdr()`
- [x] Update target format selection code

### DIFF
--- a/src/MagnumPlugins/BasisImporter/BasisImporter.h
+++ b/src/MagnumPlugins/BasisImporter/BasisImporter.h
@@ -61,9 +61,11 @@ namespace Magnum { namespace Trade {
 
 @m_keywords{BasisImporterEacR BasisImporterEacRG BasisImporterEtc1RGB}
 @m_keywords{BasisImporterEtc2RGBA BasisImporterBc1RGB BasisImporterBc3RGBA}
-@m_keywords{BasisImporterBc4R BasisImporterBc5RG BasisImporterBc7RGBA}
-@m_keywords{BasisImporterPvrtc1RGB4bpp BasisImporterPvrtc1RGBA4bpp}
-@m_keywords{BasisImporterAstc4x4RGBA BasisImporterRGBA8}
+@m_keywords{BasisImporterBc4R BasisImporterBc5RG BasisImporterBc6hRGB}
+@m_keywords{BasisImporterBc7RGBA BasisImporterPvrtc1RGB4bpp}
+@m_keywords{BasisImporterPvrtc1RGBA4bpp BasisImporterAstc4x4RGBA}
+@m_keywords{BasisImporterAstc4x4RGBAF BasisImporterRGBA8 BasisImporterRGB16F}
+@m_keywords{BasisImporterRGBA16F}
 
 Imports [Basis Universal](https://github.com/binomialLLC/basis_universal)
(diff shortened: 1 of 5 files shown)


src/MagnumPlugins/BasisImporter/BasisImporter.h
```diff
@@ -440,13 +440,23 @@ class MAGNUM_BASISIMPORTER_EXPORT BasisImporter: public AbstractImporter {
         /** @brief Target format */
         TargetFormat targetFormat() const;
 
+        /** @brief HDR target format */
+        TargetFormat targetFormatHdr() const;
```
mosra, line 444: I'm deprecating direct plugin instantiation because it's next to impossible to test, causes maintenance issues with duplicated config etc.. It's in bdc2bfd7731884aff757e36ec6d106abe7e8ac8c, I just didn't push to master to not cause you headaches because I assumed you're already working on this in some way.

So ... I'm wondering if it makes sense to expose these two APIs at this point, probably not? Given they're not really callable in the usual case, and basically just add impossible-to-cover lines.

OTOH in that commit I'm keeping the `TargetFormat` as is, because it still feels like the best way to document the various formats, link to them, etc., but other than that it's only used internally, not settable or queryable by the user in any way.

(I also hope you're not relying on this direct instantiation anywhere in your codebase. IIRC it used to be the case many many years ago, but shouldn't be anymore.)

mosra (commented):
Awesome, thanks a lot.

Now that the extra WebGL extensions are there, could you update the format setup snippet in the docs as well?

src/MagnumPlugins/BasisImporter/BasisImporter.h
pezcode: Feel free to push to master, I can rebase on top of that

mosra: Done.

src/MagnumPlugins/BasisImporter/Test/BasisImporterTest.cpp
```diff
@@ -776,7 +775,7 @@ void BasisImporterTest::invalidConfiguredFormat() {
     setTestCaseDescription(data.name);
 
     #if BASISD_LIB_VERSION < 150
-    if(data.requiresHdr)
+    if(data.formatHdr != ""_s)
```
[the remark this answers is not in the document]
pezcode: `formatHdr` is a `const char*`, felt weird making it a `StringView` when all other string literals in the test are `const char*` as well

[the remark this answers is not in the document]
mosra: Right, sorry, bad suggestion then.

mosra:
Thank you!

PR #149, opened 2025-06-26
Update stb_truetype.h from 1.21 to 1.26

asmaloney:


asmaloney:
There doesn't seem to be anything I can do about this AppVeyor issue. It just cancelled stuff on its own…

asmaloney:
> Is there any particular feature in the update you rely on?

No—I'd just updated it hoping to get better results and so decided to submit it since it seemed to have a bunch of fixes.

Kerning would be awesome—that's probably what's causing some of my issues!

Thanks Vladimír!

mosra:
> that's probably what's causing some of my issues

Feel free to list them here. 

Currently ([for the UI library in magnum-extras](https://github.com/mosra/magnum-extras/tree/master/src/Magnum/Ui)) main effort goes into HarfBuzzFont which implements kerning along with ligatures and [all other OpenType features](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Text.html#a79e8fd839a580539c272e145949d8f25), and StbTrueTypeFont doesn't get as much attention in comparison. So let me know what else would you need and I can look into that.

asmaloney:
I've decided to let Qt render my text for now, so please feel free to close this one if it's of no use.

Thank you!

mosra:
Finally got around to finish this. Merged as b2f0992e19ebdfad76fdb11bb96a0625236ca069, and kerning is implemented in 32e3270380ed9b261f0c03459d3c57ba485a461b.

Likely irrelevant now that you're using Qt for text rendering, but just for the record -- glyphs being clipped is likely either due to blending not enabled when drawing, or due to depth test enabled. The glyphs are rendered as individual quads and unless the font is fixed-width, the rectangles overlap. The only way to render them correctly is with depth test turned off (so a glyph doesn't cause a part of another glyph, rendered at the same depth, to be discarded) and blending enabled (so overlapping glyph quads correctly blend together).

asmaloney:
I was careful with the blending mode, but I likely missed turning off the depth test. I still have my experiment in a branch, so now I'll know what to check if/when I revive it.

Thanks Vladimír!

PR #152, opened 2026-05-29
add support for TTC collections in Font classes

arsnyder16:


mosra:
Hi, thank you!

The PR makes sense, thanks for the accompanying tests and docs as well. I'm thinking I would expose the font index in the AbstractFont interface already along with querying how many fonts there is. That makes it cleaner as it doesn't require plugin-specific options that would need to be explained for each plugin separately. I'll think about the best way to do this and will adapt your PR accordingly. 

~~Last time I looked into stb_truetype APIs for this there seemed to be no way to query how many fonts there are in a particular collection besides trial and error, I hope FreeType has that at least.~~ Actually there is an API now, got added 9 years ago :)

Besides that, do you perhaps have a script or a command line that was used to produce `Oxygen.collection.ttc` file, or was it by hand? If not, I can attempt to replicate that using FontForge scripting. (Just trying to minimize sizes of test files and have them reproducible.)

PR #153, opened 2026-05-31
Use AbstractFont font index for font collections

arsnyder16:
Updates FreeType, StbTrueType, and HarfBuzz font plugins to use the new AbstractFont font-index API for TTC font selection, with regression tests and a reproducible TTC fixture generator.

Requires https://github.com/mosra/magnum/pull/695


mosra:
This is awesome, thank you. I'll merge this some time this week, along with a bunch of other changes I have pending for the font APIs.

mosra:
Thank you! Merged as fac7685303a9d4e5e01564b614c4f7e72becee5a with changes to corresponds to how the API looks in mosra/magnum@8ae70a038abb0667c953553f4a57b46b693af4c6.
 
Thanks for the Python script as well, I abused it a bit to instead create a tiny subset file that's just 3 kB. Likely did some font crimes there, but both implementations seem to be fine with what I did.

